PR #174, opened 2026-01-05
fix: remove chunked writing that breaks ANSI escape sequences

ms2sato:
## Summary
- Remove `writeInChunks` function that was splitting terminal history by lines and writing in chunks
- This chunking broke ANSI escape sequences like `\x1b[1A` (cursor up) and `\x1b[2K` (clear line) that operate across chunk boundaries
- Result: terminal history restoration now correctly processes escape sequences, fixing duplicate line display issue

## Problem
When restoring large terminal history from server files, lines that should be overwritten (e.g., progress indicators like "Ideating...") were displayed multiple times instead of being properly updated in place.

## Root Cause
The `writeInChunks` function split data by `\n` and wrote each chunk separately. ANSI escape sequences that move the cursor up (`\x1b[1A`) or clear lines (`\x1b[2K`) only work within their chunk, not across chunk boundaries.

## Solution
Remove chunked writing and use direct `terminal.write(data)`. xterm.js internally handles buffering, so this is safe for any data size. Users will simply wait for the write to complete.

## Test plan
- [x] TypeScript type check passes
- [x] All tests pass
- [ ] Manual verification: reload a session with large history containing progress indicators

## Summary by CodeRabbit

* **Refactor**
  * Reworked terminal history writing to use a safer, full-history writer for large outputs, improving reliability and throughput.
  * Simplified history restore flow to ensure consistent replay after reloads.

* **Bug Fix**
  * Prevented fragmentation of ANSI control sequences during large writes, preserving colors and terminal behavior.

* **Tests**
  * Added comprehensive tests validating chunked-writing safety and full-history restore behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/packages/client/src/components/Terminal.tsx
+++ b/packages/client/src/components/Terminal.tsx
@@ -7,6 +7,7 @@ import '@xterm/xterm/css/xterm.css';
 import { useTerminalWebSocket, type WorkerError } from '../hooks/useTerminalWebSocket';
 import { clearVisibilityTracking, requestHistory } from '../lib/worker-websocket.js';
 import { isScrolledToBottom } from '../lib/terminal-utils.js';
+import { writeFullHistory } from '../lib/terminal-chunk-writer.js';
 import { saveTerminalState, loadTerminalState } from '../lib/terminal-state-cache.js';
 import type { AgentActivityState } from '@agent-console/shared';
 import { ChevronDownIcon } from './Icons';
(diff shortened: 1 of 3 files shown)


ms2sato:
Addressed CodeRabbit review in 6305520:

- Added `findSafeSplitPoint()`: Detects and avoids breaking ANSI CSI sequences at chunk boundaries
- Added `writeDataInChunks()`: Uses backpressure (waits for terminal.write callback) between chunks
- Default 100KB chunk size balances performance with xterm.js limits
- Small data (<100KB) bypasses chunking for efficiency

PR #632, opened 2026-04-16
fix: pending buffer not flushing to disk (closes #631)

### LINKED ISSUE #631: bug: pending buffer not flushing to disk — output file stale for weeks
## Summary

Worker output pending buffer accumulates indefinitely without being flushed to disk. In the observed case, 5.3MB of pending buffer data remained in memory while the output file hadn't been modified for 17 days.

## Evidence

**CTO session worker** (`f5d6194b.../48c33df0...`):

| Item | Value |
|------|-------|
| Output file size | 3,230,254 bytes (3.2MB) |
| Output file last modified | 2026-03-30 (17 days ago) |
| Pending buffer size | 5,341,227 bytes (5.3MB) |
| Total offset (file + pending) | 8,571,481 bytes |
| Server PID | 56267 (started ~15h ago) |

The server has been restarted multiple times recently, which clears the in-memory pending buffer. Yet 5.3MB accumulated since the latest restart without being flushed.

## Expected Behavior

`bufferOutput()` schedules a flush timer (100ms) or triggers immediate flush when buffer exceeds 64KB threshold. The pending buffer should be flushed to disk regularly and not grow beyond the threshold.

## Impact

This bug causes multiple downstream issues:

1. **Stale history served to clients**: `readLastNLines` combines file + pending buffer. When the file is stale, the last N lines may contain outdated content, causing the terminal to display old history on session switch.
2. **Memory leak**: Pending buffer grows without bound (5.3MB observed, potentially more).
3. **Truncation never triggers**: File size remains constant (below max), so `truncateFile` is never called even though total output exceeds the limit.
4. **Data loss on server crash**: Pending buffer is in-memory only — a crash loses all unflushed output.

## Relevant Code

- `packages/server/src/lib/worker-output-file.ts`
  - `bufferOutput()` (line 162): Adds data to buffer, schedules flush
  - `flushBuffer()` (line 197): Writes buffer to disk, resets buffer
  - Flush threshold: `WORKER_OUTPUT_FLUSH_THRESHOLD = 64KB`
  - Flush timer interval: `WORKER_OUTPUT_FLUSH_INTERVAL = 100ms`

## Debugging Technique: Accessing Pending Buffer via WebSocket

The pending buffer is in-memory only and not exposed via REST API. To inspect it, request history at the file size offset:

```bash
bun -e '
const ws = new WebSocket("ws://localhost:6340/ws/session/<SESSION_ID>/worker/<WORKER_ID>");
ws.onopen = () => {
  // Request from just past the file size to get only pending buffer content
  ws.send(JSON.stringify({ type: "request-history", fromOffset: <FILE_SIZE_BYTES> }));
};
ws.onmessage = (e) => {
  const msg = JSON.parse(e.data);
  if (msg.type === "history") {
    console.log("Pending buffer size (chars):", msg.data.length);
    console.log("Total offset:", msg.offset);
    console.log("Pending buffer (bytes):", msg.offset - <FILE_SIZE_BYTES>);
    ws.close();
    process.exit(0);
  }
};
setTimeout(() => { ws.close(); process.exit(1); }, 5000);
'
```

To check the full history response (simulating a fresh client with no cache):
```bash
bun -e '
const ws = new WebSocket("ws://localhost:6340/ws/session/<SESSION_ID>/worker/<WORKER_ID>");
ws.onopen = () => {
  ws.send(JSON.stringify({ type: "request-history", fromOffset: 0 }));
};
ws.onmessage = (e) => {
  const msg = JSON.parse(e.data);
  if (msg.type === "history") {
    console.log("offset:", msg.offset, "dataLen:", msg.data.length);
    // Strip ANSI and find last meaningful lines
    const stripped = msg.data.replace(/\x1b\[[0-9;]*[a-zA-Z]/g, "");
    const lines = stripped.split("\n").filter(l => l.trim());
    console.log("Last 5 lines:", lines.slice(-5));
    ws.close(); process.exit(0);
  }
};
setTimeout(() => { ws.close(); process.exit(1); }, 5000);
'
```

## Relationship to #627

This bug is a contributing factor to the terminal stale display issue (#627). When pending buffer is not flushed:
- The generation counter (PR #628) cannot detect staleness because truncation never fires
- The offset grows in-memory but the file never changes, creating a mismatch between file-based and memory-based state

### Code path: `bufferOutput` → `flushBuffer`

- `bufferOutput()` (line 162): Appends data to `pending.buffer`, triggers flush if buffer >= 64KB threshold, otherwise schedules 100ms timer
- `flushBuffer()` (line 197): Synchronously clears timer and copies buffer to local var (`pending.buffer = ''`), then performs async file I/O
- The synchronous buffer clearing means the buffer **should** never exceed threshold + one chunk
- Yet 5.3MB accumulated — this implies flushBuffer is either never called or always failing

### Key observations
- `flushBuffer` clears the buffer **before** the first `await` — so even if file I/O fails, the buffer is already empty
- This also means **data loss on write failure**: the data is removed from buffer but not written to disk
- There is no retry or buffer restoration on failure
- The `void this.flushBuffer()` fire-and-forget pattern means errors are only logged, not propagated

### Possible root causes (to investigate)
1. **Timer not firing** in Bun under certain conditions (long-lived process, heavy I/O)
2. **Silent failure in `flushBuffer`** that causes early return without clearing the buffer (need to verify)
3. **Race condition** between concurrent flush calls that causes data to be orphaned
4. **File I/O failure** that's consistently failing (permission, path resolution) but buffer IS cleared → data loss not accumulation

Note: Cause #4 would result in data loss (empty buffer after failed write), NOT accumulation. The 5.3MB accumulation suggests flushBuffer is never called, or something prevents the synchronous buffer-clearing from executing.

## Acceptance Criteria

- [ ] Root cause of buffer not flushing is identified and documented in PR description → investigation
- [ ] Buffer data is reliably flushed to disk: after flush interval (100ms), pending buffer is empty and file reflects the data → unit test
- [ ] Buffer cannot grow beyond threshold + single data chunk under sustained writes → unit test  
- [ ] On file write failure, buffered data is NOT lost (buffer is restored or retry is scheduled) → unit test
- [ ] `flushAll()` reliably writes all pending data to disk (existing behavior preserved) → unit test
- [ ] No regression in existing `worker-output-file.test.ts` tests

ms2sato:
## Summary

Fixes #631. The root bug: `flushBuffer()` clears `pending.buffer` **synchronously** before starting async file I/O, so any write failure silently discards the data (buffer empty, data never on disk, no retry).

The fix:
- **Restore on failure**: If the critical write step (`appendFile` / `writeFile` during legacy migration) throws, prepend `dataToWrite` back to `pending.buffer` (preserves order with any data accumulated during the failed flush) and schedule a retry via the existing timer mechanism.
- **Separate critical from non-critical I/O**: Post-write `stat` + `truncateFile` are non-critical — data is already persisted — so failures there are logged, not restored.
- **Preserve sync buffer-clear invariant**: `pending.buffer = ''` still happens synchronously before any `await`, so concurrent `bufferOutput` calls during in-flight I/O write to a fresh buffer. No mutex introduced — a mutex would *increase* accumulation under load when writes outpace flushes.
- **Worker-reset safety**: If the pending entry is removed (via `resetWorkerOutput` / `deleteWorkerOutput` / `deleteSessionOutputs`) during a failed flush, the restore path discards the data silently — the caller intentionally wanted it gone.

## Root cause analysis

The observed 5.3MB in-memory accumulation with the output file stale for 17 days cannot be fully explained by the code as written — the synchronous buffer-clear should bound growth to `threshold + one chunk`. The most plausible path to silent data loss is consistent write failures that clear the buffer without writing. This PR hardens that path specifically.

Candidates ruled out / not addressed here:
- Timer not firing — no evidence; setTimeout is well-behaved on Bun.
- Concurrent flush race — each flush captures its own `dataToWrite` via sync clear; no accumulation.
- `truncateFile` read-modify-write race losing concurrent appends — possible failure mode but causes *loss*, not accumulation.

## Tests added (`describe('write failure recovery')`)

1. `should restore buffer and retry on appendFile failure` — First timer-fired flush fails (blocking dir at file path → EISDIR), retry after unblock succeeds, file contains `'critical data'`.
2. `should preserve data when multiple writes occur during failed flush` — Verifies prepend order: buffer ends up `'data1data2'` after failure + concurrent write + retry.
3. `should not lose data when flushAll encounters write failure followed by success` — `flushAll` returns; blocking dir still present; unblock + wait 200ms; file contains `'important'`.
4. `should not accumulate buffer beyond threshold + chunk under sustained writes` — 50 × 50-byte writes with 5ms interleaves. All 2500 bytes land on disk in order.

### Note on mocking strategy

Diagnostic testing showed that `spyOn(memfs.promises, 'appendFile')` / monkey-patching does **not** intercept production code that imports `from 'fs/promises'`. The `mock.module('fs/promises', () => fs.promises)` registration copies own-properties into the ESM namespace binding, so later property mutations are invisible. Instead, the tests inject failure via memfs itself: pre-create a directory at the target file path so `fs.appendFile` throws `EISDIR`, then remove the directory to let the retry succeed. This produces genuinely failing first flushes and verifies the retry path end-to-end.

## Test plan

- [x] `bun test` in `packages/server` — 2233 pass / 0 fail (78 pass / 0 fail in `worker-output-file.test.ts` specifically).
- [x] `bunx tsc --noEmit` in `packages/server` — clean.
- [ ] Client typecheck currently fails on pre-existing TanStack Router `routeTree.gen.ts` type errors that reproduce on `origin/main`; unrelated to this PR.

### DIFF
--- a/packages/server/src/lib/__tests__/worker-output-file.test.ts
+++ b/packages/server/src/lib/__tests__/worker-output-file.test.ts
@@ -1014,4 +1014,132 @@ describe('WorkerOutputFileManager', () => {
       expect(filePath).toBe(`${TEST_CONFIG_DIR}/_quick/outputs/session-1/worker-1.log`);
     });
   });
+
+  describe('write failure recovery', () => {
+    /**
+     * Helper: pre-create a directory at the path where the worker output file
+     * will be written. This causes `fs.appendFile` (and `fs.writeFile`) to fail
+     * with EISDIR when the flush tries to write. Returns the file path so the
+     * test can `vol.rmdirSync` later to allow a retry to succeed.
+     *
+     * This approach is used instead of monkey-patching memfs because ESM
+     * namespace bindings (from `import * as fs from 'fs/promises'`) are
+     * immutable — runtime mutations to the memfs module object do not affect
+     * callers who imported via a namespace binding. Forcing a real I/O failure
+     * through the file system is the only reliable way to exercise the error
+     * path in this test harness.
+     */
+    const blockWriteWithDirectory = (sessionId: string, workerId: string): string => {
+      const filePath = manager.getOutputFilePath(sessionId, workerId, quickResolver);
+      // Create a directory at the target file path → appendFile will throw EISDIR.
+      vol.mkdirSync(filePath, { recursive: true });
+      return filePath;
+    };
+
+    /** Remove the blocking directory so a subsequent flush can succeed. */
+    const unblockWrite = (filePath: string): void => {
+      vol.rmdirSync(filePath);
+    };
+
+    it('should restore buffer and retry on appendFile failure', async () => {
+      const filePath = blockWriteWithDirectory('session-fail-1', 'worker-1');
+
+      manager.bufferOutput('session-fail-1', 'worker-1', 'critical data', quickResolver);
+
+      // Wait for first flush to fire (timer = 100ms). It fails → buffer restored,
+      // retry scheduled 100ms later. Allow the first attempt to complete.
+      await new Promise(resolve => setTimeout(resolve, 130));
+
+      // At this point the first flush has failed. File path still exists as a dir.
+      // Unblock so the retry (next timer tick) can succeed.
+      unblockWrite(filePath);
+
+      // Wait for retry flush to complete.
+      await new Promise(resolve => setTimeout(resolve, 150));
+
+      expect(vol.existsSync(filePath)).toBe(true);
+      // Sanity-check it's now a file (not the blocking directory).
+      expect(vol.statSync(filePath).isFile()).toBe(true);
+      const content = vol.readFileSync(filePath, 'utf-8');
+      expect(content).toBe('critical data');
+    });
+
+    it('should preserve data when multiple writes occur during failed flush', async () => {
+      const filePath = blockWriteWithDirectory('session-fail-2', 'worker-1');
+
+      // First write triggers the buffering; flushAll forces a synchronous flush attempt.
+      manager.bufferOutput('session-fail-2', 'worker-1', 'data1', quickResolver);
+      await manager.flushAll(); // First flush fails → 'data1' restored + retry scheduled.
+
+      // Second write arrives after restore; buffer becomes 'data1' + 'data2'.
+      manager.bufferOutput('session-fail-2', 'worker-1', 'data2', quickResolver);
+
+      // Before the retry timer fires (100ms after restoreBufferOnFailure), unblock.
+      unblockWrite(filePath);
+
+      // Wait for retry timer to fire and succeed.
+      await new Promise(resolve => setTimeout(resolve, 200));
+
+      expect(vol.existsSync(filePath)).toBe(true);
+      expect(vol.statSync(filePath).isFile()).toBe(true);
+      const content = vol.readFileSync(filePath, 'utf-8');
+      expect(content).toBe('data1data2');
+    });
+
+    it('should not lose data when flushAll encounters write failure followed by success', async () => {
+      const filePath = blockWriteWithDirectory('session-fa', 'worker-1');
+
+      manager.bufferOutput('session-fa', 'worker-1', 'important', quickResolver);
+
+      // flushAll triggers a flush that fails; data is restored to buffer.
+      await manager.flushAll();
+
+      // File path still points to the blocking directory — no regular file yet.
+      expect(vol.statSync(filePath).isDirectory()).toBe(true);
+
+      // Unblock so the retry timer (scheduled by the failed flush) can succeed.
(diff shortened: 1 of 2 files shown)


ms2sato:
Closing in favor of #638 (session-data-path scope-based persistence, merged) which addresses the root cause of #631 — worker output file path fragmentation.

The defensive hardening in this PR (buffer restore on write failure, exponential backoff, buffer cap, consecutive-failure tracking, diagnostic logging) was valuable when the path could drift across restarts. With #638 merged, path resolution is deterministic and path-drift-induced write failures no longer occur.

Leaving follow-up to evaluate if worker-output-file.ts can be simplified now that the underlying assumptions have changed.

Issue #660, opened 2026-04-17
bug: message input splits into multiple messages when 3+ consecutive newlines are entered


ms2sato:
## Summary

When composing a message in the AgentConsole message/chat input panel, entering **3 or more consecutive newlines** (i.e., 2 or more blank lines between content blocks) causes the input to be split into multiple separate messages instead of being sent as a single message.

## Observed behavior

Reproduced during Sprint 2026-04-17b retrospective. User typed a multi-section message using 2 blank lines as section separators, and the message arrived in the session as two separate user turns.

Exact trigger suspected: the input component or the posting logic treats a run of ≥3 consecutive newlines (a fully empty line between content) as a message delimiter.

## Expected

A single message containing both content blocks, separated by the blank lines as the user typed them.

## Actual

Two (or more) separate messages, each delivered as an independent user turn to the session. The assistant responds to the first block before the second arrives.

## Impact

- Interrupts flow when composing long / multi-section messages (retrospective reviews, architectural proposals, multi-topic updates)
- Causes confusing mid-composition assistant turns
- User has to avoid natural paragraph spacing to prevent the split

## Repro steps

1. Open a session in AgentConsole
2. In the message input panel, type some content
3. Press Enter three times (or enter two blank lines)
4. Type more content
5. Submit

**Expected**: one message with the full content.
**Actual**: two messages, second one delivered separately.

## In Scope

- Identify the component responsible (likely the message input / chat compose panel in \`packages/client/src/components/**\`)
- Identify where the split occurs (client-side input handling, WebSocket message framing, or server-side message splitting)
- Fix so consecutive blank lines are preserved within a single message

## Out of Scope

- Redesign of the input panel
- Changes to how the assistant renders multi-paragraph responses (this is about input, not output)

## Acceptance Criteria

- [ ] A single submit with 3+ consecutive newlines delivers as one message to the session
- [ ] Blank lines are preserved in the delivered message body
- [ ] Test covers the 3-newline edge case explicitly (unit or integration, whichever catches the actual split point)

## Related

- Sprint 2026-04-17b retrospective — discovered during owner-orchestrator message exchange

PR #793, opened 2026-05-19
fix: wrap PTY-injected text in bracketed paste delimiters

### LINKED ISSUE #792: fix: MessagePanel input dropped on interactive prompts (bracketed paste mode not used)
> **Updated 2026-05-25 after real-device E2E (PR #793 disproved).** The original "missing bracketed paste" diagnosis below was wrong. See the corrected root cause. Prior hypothesis preserved at the bottom for history.

## Symptom

When a Claude Code agent worker is showing an interactive prompt (tool-permission request or AskUserQuestion-style numbered menu) and the user sends a message via the MessagePanel UI to answer it, the typed text is **not reflected**. The injection is treated as a bare Enter, which confirms the default highlighted choice (option 1). Reported from owner dogfooding; started after a recent Claude Code CLI update. Plain (non-prompt) message sends still work.

## Root cause (verified by real-device E2E, high confidence)

The shipping path (MessagePanel -> server -> `injectMessage`) writes the message text to the PTY, then a delayed `\r` as submit. When the agent is at an interactive prompt:

- The injected text lands **on the numbered menu**, which does not route arbitrary text into a free-text field on a single injection (answering free-text on the menu requires the multi-step "Type something -> Enter -> input"). The trailing bare `\r` then confirms the default option.
- This was **not** caused by missing bracketed-paste wrappers. E2E with real `claude` v2.1.150 (instrumented byte log) confirmed bracketed paste **was** applied (`[200~...[201~`) yet PRE-FIX and POST-FIX both produced the same default-option confirmation — i.e. wrapping the text in bracketed paste does **not** change the menu behavior. PR #793's unconditional bracketed-paste approach does not fix this.
- The real change: the previous Claude CLI was **lenient** — unexpected text arriving at a prompt auto-cancelled the menu and flowed into the main composer. The newer CLI removed that leniency, so injected text no longer escapes the menu on its own.

## Correct fix

When the worker is in an **asking / waiting-for-input** state, inject `ESC` (cancel the prompt menu) **then** the text **then** the delayed `\r`. E2E confirmed `ESC + text + \r` reaches the main composer and Claude receives it as a normal message:

```
User declined to answer questions          <- ESC cancels the menu
> Actually my favorite is Durian ...        <- text reaches the composer
(claude responds to it as a normal message)
```

**Gating constraint (owner):** `ESC` must NOT be prepended unconditionally. Sending `ESC` on a normal idle send would interrupt Claude. Apply the `ESC` prefix **only** when the worker is in the `asking` state; otherwise use the current normal-send path unchanged.

## Layer / implementation notes

- The injection call sites are `session-manager.ts:411` (`sendMessage`) and `:754`. The asking state is reachable there via `getWorkerActivityState(sessionId, workerId)` (`session-manager.ts:773`), backed by `activity-detector.ts` `askingPatterns` (`^\s*❯\s+\d+\.\s`, `Esc to cancel`, etc.). Gate the `ESC` prefix on this state — either pass an `isAsking` flag into `injectMessage`, or branch in `session-manager` before the call. (Removal-type / behavior-gating AC: grep the existing surface first — `getWorkerActivityState` and the `AgentActivityState` `asking` value already exist; do not introduce a parallel detector.)
- **Bracketed paste: remove the unconditional wrapping** introduced by PR #793. The Issue #660 soft-newline mechanism (`\r?\n -> \n`, only the final `\r` submits) already handles multi-line bodies; unconditional bracketed paste is redundant and does not fix #792. Only reintroduce a paste wrapper if E2E shows the post-ESC composer needs it for multi-line safety.

## Acceptance Criteria

- [ ] Injection at an `asking`-state worker emits `ESC` then text then delayed `\r`; the text reaches the composer (not the menu) -> unit test on the injection/gating + real-device E2E
- [ ] Injection at a non-`asking` worker is byte-identical to the current normal-send path (no `ESC`) -> unit test (both branches) — why: an unconditional `ESC` interrupts Claude
- [ ] Issue #660 newline behavior preserved (embedded newlines are soft, only the final `\r` submits) -> unit test
- [ ] Boundary cases: content-only / content+files / empty content; `asking` true and false both covered -> unit tests (vacuous/empty must be explicit)
- [ ] No regression on a normal (non-prompt) send: single round-trip with newline / multi-line / Japanese input verified on real device -> E2E
- [ ] PRE-FIX reproduces the bug (text dropped -> default option) and POST-FIX shows the answer reaching the composer, polarity proven by stashing only the fix -> E2E raw paste in PR body

## Prior (incorrect) hypothesis — kept for history

The original body diagnosed the cause as "missing bracketed paste mode handling (`ESC[200~`/`ESC[201~`)" and proposed unconditional bracketed-paste wrapping. This was disproved by E2E (PR #793): bracketed paste was applied but the menu behavior was unchanged. Superseded by the corrected root cause above.

ms2sato:
Closes #792

## Problem

When a Claude Code agent worker shows an interactive prompt and the user answers via the MessagePanel, the typed text was dropped — the injection landed as a bare Enter that confirmed the default highlighted option (option 1). Reported from owner dogfooding after a recent Claude Code CLI update; plain non-prompt sends still worked.

## Root cause

Recent Claude Code enables **bracketed paste mode** (`ESC[?2004h`). `PtyMessageInjectionService.injectMessage` wrote the message text to the PTY without the `ESC[200~ … ESC[201~` paste wrappers, so the modern interactive-prompt input handler discarded it and only the trailing bare `\r` reached the prompt — confirming the default option. The repository had zero bracketed-paste handling.

## Fix

`packages/server/src/services/pty-message-injection-service.ts`: wrap every injected content/file part in bracketed paste delimiters (`\x1b[200~` … `\x1b[201~`), **unconditionally** (not gated on activity-detector state). The final delayed `\r` remains a separate, **unwrapped** submit keystroke; the per-file leading `\r` (submit of the previous part) also stays unwrapped — only the part payload is wrapped. The Issue #660 newline normalization (`/\r?\n/g → \n`) is preserved and the normalized LF newlines now live inside the paste envelope (literal text, never submit).

Two private constants (`PASTE_START` / `PASTE_END`) plus a `wrapPaste()` helper avoid string-literal duplication.

### Sibling call-site (Option A bundle)

Grepping sibling call sites surfaced 6 assertions in `__tests__/session-manager.test.ts` that assert the delegated `injectMessage` byte stream (`session-manager.ts` / `pty-operation-executor.ts` only delegate — no production sibling change needed). Bundled the test-assertion updates in this PR to keep `main` GREEN (Option A per the grep-siblings rule).

## Verification

**Local suite — `bun run test` exit 0:**
- server 2513 pass / 1 skip / 0 fail · client 1386 pass / 2 skip / 0 fail · shared 324 pass · integration 29 pass · scripts/hooks 362 pass
- `bun run typecheck` clean (runs as part of `bun run test`)

**CodeRabbit** (`coderabbit review --agent --base main`): **0 findings**.

**Real-device verification** (deterministic PTY probes against the real `claude` v2.1.144 binary, since the dev environment auto-approves common tool calls and this Claude Code version's permission prompt offers no free-text option — environment/version-specific blocker, documented transparently per CLAUDE.md "Real-device verification"):

1. **Precondition (primary-verified, 3 independent probe runs):** real `claude` emits `ESC[?2004h` — bracketed paste mode is enforced. This is the diagnosed root cause, observed directly (not inferred).
2. **Bug reproduced — PRE-FIX bytes:** at a real Bash-tool permission prompt (`Do you want to proceed? ❯ 1. Yes / 2. … / 3. No`), injecting an un-wrapped answer followed by a delayed `\r` (the exact byte sequence the pre-fix code produced) → the answer text was dropped, the default highlighted **option 1 was confirmed and the command ran**. Exactly the reported symptom.
3. **Fix mechanism verified — POST-FIX bytes:** a `\x1b[200~ … \x1b[201~`-wrapped multi-character injection was correctly received and acted upon by the real `claude` binary (it executed the wrapped instruction). This is the mechanism the fix introduces; un-wrapped text at the new prompt handler does not survive (corroborated by upstream anthropics/claude-code #13183, #47745).
4. **No regression for normal sends:** bracketed-paste wrapping is exactly what a human terminal paste already emits (xterm.js sends `ESC[200~`/`ESC[201~` on paste), so injected normal messages now match the human-paste path; a wrapped composer message was processed normally by `claude` in the probe, and the full unit/integration suite (which exercises the real `injectMessage` byte stream via session-manager) is green.

**Residual risk (honest, matches the Issue's own note):** a pure list-selector permission prompt has no free-text field, so injected text cannot be consumed there by design — that path needs arrow/digit navigation and bracketed paste neither solves nor regresses it. The MessagePanel free-text answer path — the one users hit and the one this fix targets — is the bracketed-paste path and is addressed. Owner dogfood re-verification on a free-text interactive prompt post-merge is recommended as the final gate.

### DIFF
--- a/packages/server/src/services/__tests__/pty-message-injection-service.test.ts
+++ b/packages/server/src/services/__tests__/pty-message-injection-service.test.ts
@@ -1,6 +1,11 @@
 import { describe, it, expect, beforeEach, afterEach, mock, jest } from 'bun:test';
 import { PtyMessageInjectionService } from '../pty-message-injection-service.js';
 
+// Bracketed paste delimiters (DEC private mode 2004) — see Issue #792.
+const PASTE_START = '\x1b[200~';
+const PASTE_END = '\x1b[201~';
+const wrap = (s: string) => `${PASTE_START}${s}${PASTE_END}`;
+
 describe('PtyMessageInjectionService', () => {
   let writeInput: ReturnType<typeof mock>;
   let isWorkerActive: ReturnType<typeof mock>;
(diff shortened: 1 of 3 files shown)


ms2sato:
Closing per owner decision (2026-05-25). Real-device E2E (shipping path MessagePanel->server->injectMessage + real claude v2.1.150, AskUserQuestion free-text, polarity by stashing only the fix) disproved this approach: bracketed paste was confirmed applied (instrumented byte log showed `[200~...[201~`) yet PRE-FIX and POST-FIX both confirmed the default menu option — wrapping text in bracketed paste does not change the numbered-menu behavior. The real root cause is that the newer Claude CLI removed the lenient behavior where unexpected text at a prompt auto-cancelled the menu and flowed to the composer. Correct fix = gate on `asking` state and inject `ESC` + text + `\r` (verified working on real device). Issue #792 root cause has been corrected. Reimplementation proceeds on a fresh branch from origin/main. Thanks to the verification that caught this before merge.

PR #812, opened 2026-06-17
fix: recover orphaned worktree when primary repo dir is deleted (closes #811)

### LINKED ISSUE #811: Worktree cannot be removed from AgentConsole when its primary repo dir is deleted (orphaned DB row, force fallback also fails)
## Summary

When a worktree's **primary repository directory** (the original checkout that owns the common `.git` dir) is deleted out-of-band, the remaining linked worktree **cannot be removed from AgentConsole** — not even with `force: true`. The worktree's DB record orphans and shows as an un-removable entry.

## Root cause

Every git operation in the removal path runs with **`cwd = repoPath` (the primary repo path)**. Once the primary is gone, that cwd no longer exists, so every step fails:

1. `removeWorktree()` calls `gitRemoveWorktree(worktreePath, repoPath, { force })` — `packages/server/src/services/worktree-service.ts:323`.
2. `gitRemoveWorktree` runs `git worktree remove <worktreePath>` with `cwd = repoPath` — `packages/server/src/lib/git.ts:309-317`. Primary missing → `fatal: not a git repository` (exit 128).
3. **The force fallback is also broken** — `packages/server/src/lib/git.ts:330-338`: when `force` is set and the error mentions `.git`, it `fs.rm`s the worktree dir (OK) **but then runs `git worktree prune` with the same `cwd = repoPath`**, which fails the same way and **throws**.
4. The DB record is only deleted on git success — `packages/server/src/services/worktree-service.ts:326` (`deleteByPath`) runs after `gitRemoveWorktree` returns. Since step 2/3 throws, the row is never deleted.
5. `git worktree list` also uses `cwd = repoPath` (`worktree-service.ts:162`), so AgentConsole can't even enumerate the worktree afterward.

Net effect: the worktree directory may or may not be removed, but the **`worktrees` DB row orphans** and the session is preserved-for-retry (`worktree-deletion-service.ts:237-250`) — yet retry can never succeed because the primary is permanently gone.

## Reproduction

1. Create a repository A (primary) in AgentConsole.
2. Create a linked worktree B from A.
3. Delete A's directory out-of-band (e.g. `rm -rf` the primary checkout).
4. Call `remove_worktree` on B (with or without `force`).
5. **Result:** removal fails (`git worktree prune` → `fatal: not a git repository`); B's DB row orphans and B remains visible/un-removable in AgentConsole.

## Why git isn't needed here

A worktree whose primary/common git dir is gone is, from git's view, already detached — there is nothing left for `git worktree remove`/`prune` to do against the (now-absent) common dir. Cleanup only needs to: (a) `fs.rm` the worktree directory if it still exists, and (b) delete the `worktrees` DB row (and the session, per the normal flow). No git invocation is required.

## Proposed fix (direction)

Detect the **primary-missing** condition and route to a **git-independent cleanup path** (Option A):

- Before/instead of the git calls, if the primary repo path does not exist (or the git ops fail with a "not a git repository"-class error), clean up without git: `fs.rm(worktreePath, { recursive: true, force: true })` then delete the DB row + session.
- In the existing force fallback (`git.ts:330-338`), **guard the `git worktree prune` call** so it is skipped when `cwd`/primary does not exist (prune against an absent common dir is a no-op anyway).
- Ensure DB-row deletion happens in this recovery path even though the git calls were skipped/failed.

## Acceptance criteria (boundary values)

- **Primary exists + clean worktree** → normal removal, unchanged behavior.
- **Primary MISSING + worktree dir present** → removal succeeds via git-independent cleanup; the worktree disappears from AgentConsole (dir removed, DB row + session deleted).
- **Primary MISSING + worktree dir ALSO already deleted** → idempotent: DB row + session are still cleaned up, no error.
- **Existing orphaned rows** (from prior occurrences of this bug) become removable after the fix.
- Decide and document whether the primary-missing cleanup requires `force` or is allowed unconditionally (the worktree is already orphaned, so allowing it is defensible; at minimum `force` must work — today it does not).
- TDD: a regression test that removes a worktree whose primary path has been removed, asserting the DB row is deleted and no throw occurs.

## Notes

- Primary deletion here is **external** (outside AgentConsole), so prevention isn't sufficient — recovery must be supported.
- Related: if a repository-delete feature exists/lands, it should remove linked worktrees first; that is a separate concern from this orphan-recovery path.

ms2sato:
## Summary

Fixes #811 — when a worktree's **primary repository directory** (the original checkout owning the common `.git` dir) is deleted out-of-band, the linked worktree could not be removed from AgentConsole — not even with `force: true`. The DB row orphaned and showed as an un-removable entry.

Closes #811

## Root cause

Every git op in the removal path runs with `cwd = repoPath` (the primary). Once the primary is gone:
- `git()` does `Bun.spawn(['git', ...], { cwd: repoPath })`. **A nonexistent cwd makes `Bun.spawn` throw `ENOENT` — not a `GitError` with parseable stderr.** So stderr-string detection alone cannot catch the reported case; a real-fs existence pre-check on `repoPath` is the reliable detector.
- The `force` fallback was also broken: after `fs.rm`-ing the worktree dir it ran `git worktree prune` with the same dead `cwd`, which threw.
- The DB row (`deleteByPath`) was only deleted on git success, so the row orphaned.

## Fix

- **`worktree-service.ts` `removeWorktree`**: top-of-method pre-check of `repoPath` existence (`fsPromises.stat` in try/catch, the existing idiom in this file). When the primary is missing, recover **without git** — `fsPromises.rm(worktreePath, {recursive, force})` then `worktreeRepository.deleteByPath(worktreePath)` — and return success. Runs **unconditionally** (no `force` required): an orphaned worktree is always safe to reclaim, and this also guarantees the `force` path works. Idempotent — `rm` on a missing path and `deleteByPath` on a missing row are both no-ops.
- **`git.ts` `removeWorktree`** force fallback: guard `git worktree prune` so it is **skipped when `cwd` no longer exists** (prune against an absent common dir is a no-op anyway).
- `worktree-deletion-service.ts` needs **no change** — once `removeWorktree` returns `{success:true}` for the recovery path, the existing flow deletes sessions normally.

## Design decision (per issue AC)

Primary-missing cleanup is allowed **unconditionally (no `force` required)**. The worktree is already orphaned, so reclaiming it is defensible; this also satisfies the AC's "at minimum `force` must work" (today neither worked).

## Boundary-value coverage (TDD)

| Case | Behavior |
|---|---|
| ① Primary present + clean | Normal git removal — unchanged |
| ② Primary MISSING + worktree dir present | Git-independent recovery; dir removed, DB row + session deleted, git not invoked |
| ③ Primary MISSING + worktree dir also gone | Idempotent — DB row still cleaned up, no throw |
| ④ Existing orphaned row | Reclaimed/removable after the fix |

**Polarity verified**: stashing only the production hunks (keeping the new tests) makes the discriminating recovery tests (② and ③) and the git.ts prune-skip test FAIL against unmodified production; restoring makes them pass.

## Verification

- `bun run test` — full suite green (server 2578 pass / 1 skip / 0 fail, client 1396, shared 324, integration 29, scripts/hooks 362). **TEST_EXIT: 0**
- `bun run typecheck` — **0 errors**
- `coderabbit review --agent --base origin/main` — `findings: 0` (no CRITICAL/HIGH/MEDIUM)

No UI changes — browser QA not required.

### DIFF
--- a/packages/server/src/lib/__tests__/git.test.ts
+++ b/packages/server/src/lib/__tests__/git.test.ts
@@ -761,5 +761,99 @@ index abc1234..def5678 100644
         await fs.rm(tempWorktreePath, { recursive: true, force: true });
       }
     });
+
+    it('should swallow an ENOENT-coded prune error after worktree cleanup (stat→spawn race)', async () => {
+      const fs = await import('node:fs/promises');
+
+      const tempWorktreePath = `/tmp/git-test-worktree-${Date.now()}-${Math.random().toString(36).slice(2)}`;
+      await fs.mkdir(tempWorktreePath, { recursive: true });
+      // cwd exists at stat() time so the prune is attempted...
+      const tempRepoPath = `/tmp/git-test-repo-${Date.now()}-${Math.random().toString(36).slice(2)}`;
+      await fs.mkdir(tempRepoPath, { recursive: true });
+
+      try {
+        // First call (remove) fails with a .git-class error; second call (prune)
+        // throws synchronously with an ENOENT code, simulating cwd vanishing
+        // between stat() and spawn().
+        let callCount = 0;
+        (Bun as { spawn: typeof Bun.spawn }).spawn = ((args: string[], options?: Record<string, unknown>) => {
+          spawnCalls.push({ args, options: options || {} });
+          callCount++;
+          if (callCount === 1) {
+            return {
+              exited: Promise.resolve(128),
+              stdout: new ReadableStream({ start(controller) { controller.close(); } }),
+              stderr: new ReadableStream({
+                start(controller) {
+                  controller.enqueue(new TextEncoder().encode("fatal: '.git' does not exist"));
+                  controller.close();
+                },
+              }),
+            };
+          }
+          // Second call (prune): spawn itself throws ENOENT (cwd gone).
+          throw Object.assign(new Error('ENOENT: no such file or directory'), { code: 'ENOENT' });
+        }) as typeof Bun.spawn;
+
+        const { removeWorktree } = await getGitModule();
+
+        // Must resolve without throwing: the worktree fs.rm already succeeded.
+        await removeWorktree(tempWorktreePath, tempRepoPath, { force: true });
+
+        // Worktree dir was removed.
+        let worktreeStillExists = true;
+        try {
+          await fs.stat(tempWorktreePath);
+        } catch {
+          worktreeStillExists = false;
+        }
+        expect(worktreeStillExists).toBe(false);
+
+        // Both remove and prune were attempted (prune threw and was swallowed).
+        expect(spawnCalls.length).toBe(2);
+        expect(spawnCalls[1].args).toEqual(['git', 'worktree', 'prune']);
+      } finally {
+        await fs.rm(tempWorktreePath, { recursive: true, force: true });
+        await fs.rm(tempRepoPath, { recursive: true, force: true });
+      }
+    });
+
+    it('should propagate a non-ENOENT prune error (guard is narrow)', async () => {
+      const fs = await import('node:fs/promises');
+
+      const tempWorktreePath = `/tmp/git-test-worktree-${Date.now()}-${Math.random().toString(36).slice(2)}`;
+      await fs.mkdir(tempWorktreePath, { recursive: true });
+      const tempRepoPath = `/tmp/git-test-repo-${Date.now()}-${Math.random().toString(36).slice(2)}`;
+      await fs.mkdir(tempRepoPath, { recursive: true });
+
+      try {
+        let callCount = 0;
+        (Bun as { spawn: typeof Bun.spawn }).spawn = ((args: string[], options?: Record<string, unknown>) => {
+          spawnCalls.push({ args, options: options || {} });
+          callCount++;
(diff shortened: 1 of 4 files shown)


ms2sato:
Thanks — all three Major findings are addressed in 708e388:

1. **git.ts (stat→spawn race)**: wrapped `git worktree prune` in try/catch; swallow `ENOENT`/`ENOTDIR` (cwd vanished after the existence check — cleanup already succeeded), rethrow anything else.
2. **worktree-service.ts (over-broad detection)**: narrowed orphaned-primary detection to `ENOENT`/`ENOTDIR` only and now require `stat.isDirectory()`. `EACCES`/`EPERM`/IO errors no longer route to destructive recovery — they surface as `{ success: false, error }`.
3. **worktree-service.ts (unstructured error handling)**: moved the pre-check + recovery branch inside the existing outer `try/catch` (single catch, no nested) so `fs.rm`/`deleteByPath` failures return a structured error.

Added regression tests: EACCES-coded stat → `{ success: false }` with no DB-row deletion; prune ENOENT → resolves without throw; prune EPERM → propagates (proves the guard is narrow). Full suite + typecheck green, local `coderabbit review` reports `findings: 0`.

@coderabbitai review

PR #831, opened 2026-06-22
feat: relocate multi-user data root + add bootstrap script

### LINKED ISSUE #830: multi-user mode: relocate shared data to system-wide path with group-writable permissions, add bootstrap script
## Summary

In `AUTH_MODE=multi-user`, repository worktrees and other shared filesystem data live under the server process user's HOME (`$HOME/.agent-console/`), which on Debian/Ubuntu defaults to mode `0750` and is **inaccessible to other OS users**. As a result, repository-backed sessions cannot succeed in multi-user mode — the per-user PTY (running as the logged-in user via privilege elevation) cannot traverse into the server user's HOME to reach the worktree.

This proposes: relocate the shared data to a system-wide path with proper group-writable permissions, and add a one-shot bootstrap script so operators don't have to remember the moving parts during setup.

## Current State

- `packages/server/src/lib/config.ts` derives the data root from `os.homedir()` of the server process — in `multi-user` mode this is the dedicated service user (e.g. `agentconsole`).
- Default `HOME_MODE` on Debian/Ubuntu = `0750`, so `/home/agentconsole/` is `drwxr-x---` owned by `agentconsole:agentconsole`.
- The per-user PTY runs as the logged-in user (not in `agentconsole` group). It cannot reach `/home/agentconsole/.agent-console/repositories/.../worktrees/...`.
- Quick Sessions in `/tmp` work because `/tmp` is world-traversable. Repository-backed sessions are effectively broken.
- Verified empirically on Ubuntu 24.04 aarch64 (this reporter's host).

Related: Issue #821 (file upload directory collision) addressed an adjacent symptom by moving uploads to `/tmp/agent-console-uploads-<uid>/` mode `0700`. That fix is correct for upload buffer isolation but does not help the worktree access problem (uploads are still unreadable by the per-user PTY for a different reason).

## Design Decisions

- **Relocate data root to a system-wide path** (proposed: `/var/lib/agent-console/`). Default for `AUTH_MODE=multi-user`. `AUTH_MODE=none` (single-user) keeps using `$HOME/.agent-console/` so existing single-user installs are unaffected.
- **Introduce a system group** (proposed: `agent-console-users`) that includes the service user and every interactive user who logs into the app. The data root and its subtree are owned by this group with `setgid` so new entries inherit the group automatically.
- **Permission model:**
  - Data root and shared subdirectories: mode `2775` (owner `agentconsole`, group `agent-console-users`, group-writable + setgid).
  - Sensitive server-only files (DB, jwt-secret): mode `0660` owned by `agentconsole:agent-console-users`. SQLite is fine with multi-reader/writer in WAL mode; access is still gated by the server in practice.
- **umask discipline:**
  - Server systemd unit gains `UMask=0002` so files the server writes are group-writable.
  - The per-user PTY inner shell prepends `umask 0002` so files written by agents inherit group-writable mode and the setgid-inherited group.
- **All worktrees are shared.** Worktree paths stay flat (`<data root>/repositories/<org>/<repo>/worktrees/wt-NNN-XXXX/`) and every worktree is owned by `<service-user>:<shared-group>` mode `2775`. Any member of the shared group can read and write. Concurrent edits are accepted; conflict resolution is delegated to the AI agents using the worktree (the project policy is that AI-mediated edits do not race in practice). There is no per-worktree owner concept and no DB schema change for worktrees — the existing columns (`id`, `repository_id`, `path`, `index_number`, `created_at`) are sufficient.
- **Upload buffer (`packages/server/src/routes/workers.ts`):** Uploads remain in `/tmp/agent-console-uploads-<server-uid>/` and are **not** moved into the data root. Rationale: the server has no read-completion signal for `injectMessage`, so abandoned upload buffers rely on `/tmp` reaping (`systemd-tmpfiles` / `tmpwatch` / reboot) to avoid leaking disk over time. A persistent location under `/var/lib/agent-console/` would accumulate. For multi-user reachability, two changes are required in this Issue:
  1. **Permission/group adjustment under `AUTH_MODE=multi-user`:** mode becomes `2750` (setgid + group-rx) and ownership becomes `<service-user>:<shared-group>`. The post-mkdir `lstat` verification follows the same mode/group conditional. Under `AUTH_MODE=none` the existing `0700` / euid-owned behaviour is preserved.
  2. **Dynamic recovery against `/tmp` cleanup:** the in-process `uploadDirReady` cache currently locks in the "ready" state for the lifetime of the server process. On a long-running multi-user deployment `/tmp` may be reaped by `systemd-tmpfiles` while the server is up, after which subsequent uploads would target a non-existent directory. The cache is dropped (or re-verifies presence on every call) so that `ensureUploadDir()` re-creates the directory on demand with the same mode/group invariants.

## Bootstrap Script

`scripts/` already has per-user systemd service deploy scripts (`update-and-deploy-for-ubuntu.sh`, `update-and-deploy-for-mac.sh`, normally used in single-user mode setups) and the systemd / launchd templates, plus the multi-user unit template added in PR #819. There is **no** end-to-end bootstrap for `multi-user` setup yet — operators currently walk Step 0-4 of `docs/multi-user-setup-guide.md` by hand.

This Issue adds:

- `scripts/setup-multiuser-for-ubuntu.sh` (proposed name) — idempotent, single-entry, asks the operator only what cannot be inferred. Steps:
  1. Verify `apt` is available and install `pamtester`.
  2. Create the service user (default `agentconsole`) with `--system --create-home --shell /usr/sbin/nologin`. Add to `shadow` group.
  3. Create the shared group (default `agent-console-users`); add the service user and any operator-specified login users.
  4. **Install** `/etc/sudoers.d/agent-console` from the canonical template `scripts/sudoers-agent-console.template` (a new file added by this Issue). The script:
      - Prints the full proposed file contents to stdout (operator sees exactly what will be installed).
      - Prints target path, owner, group, mode.
      - Runs `visudo -cf` against the rendered file to validate syntax. Aborts on error.
      - Calls `install -m 0440 -o root -g root` to place the file.
      - Re-verifies final on-disk state (path / mode / owner) and prints confirmation.
      - Supports `--dry-run`, which performs the print + syntax-check phases only and writes nothing, so the operator can review before applying.
     The privilege level (root) is the same as the rest of the script (system user creation, group creation, `/var/lib/agent-console/`, systemd unit install), so this step is no more privileged than the others. The transparency of the print + syntax-check + dry-run path is what replaces a hand-walked `visudo` session.
  5. Create the data root (default `/var/lib/agent-console/`) with owner `<service-user>:<shared-group>` mode `2775`.
  6. Clone (or `rsync`) the built dist into the service user's home and `bun install --production`.
  7. Render `/etc/systemd/system/agent-console.service` from `scripts/agent-console-multiuser.service.template`, with `UMask=0002`, `AUTH_MODE=multi-user`, `AGENT_CONSOLE_HOME=<data-root>`, and the configured `PORT` / `AUTH_COOKIE_SECURE`.
  8. `systemctl daemon-reload && systemctl enable --now agent-console`, then print verification commands and the URL.
- A companion helper `scripts/add-multiuser-user.sh <username>` — adds an existing OS user to the shared group. Safe to run idempotently.

### Parameterization

All identifiers and paths the operator might want to change accept overrides via environment variables and matching CLI flags. Precedence: CLI flag > env var > built-in default.

| Env var | CLI flag | Default | Purpose |
|---|---|---|---|
| `AGENT_CONSOLE_SERVICE_USER` | `--user <name>` | `agentconsole` | Server-process system user |
| `AGENT_CONSOLE_SERVICE_GROUP` | `--group <name>` | `agent-console-users` | Shared group |
| `AGENT_CONSOLE_DATA_ROOT` | `--data-root <path>` | `/var/lib/agent-console` | Data root (== `AGENT_CONSOLE_HOME` at runtime) |
| `AGENT_CONSOLE_PORT` | `--port <num>` | `8080` | Server bind port |
| `AGENT_CONSOLE_AUTH_COOKIE_SECURE` | `--cookie-secure <bool>` | `false` | Set `Secure` on the auth cookie |
| `AGENT_CONSOLE_INITIAL_USERS` | `--add-user <user>` (repeatable) | (empty) | Login users to add to the shared group on first run |

Input validation:

- Usernames / group names: `^[a-z_][a-z0-9_-]{0,30}$` (POSIX-conventional). Fail fast with a clear message on mismatch.
- Port: integer in `[1, 65535]`.
- Data root: absolute path; parent directory must exist; rejects path traversal patterns.

Idempotency:

- A second invocation with the same parameters is a no-op (existing user / group / dir / unit / sudoers are detected and re-verified).
- If a parameter differs from the in-tree state of an existing install (e.g. operator changed `--group`), the script aborts unless `--force` is passed. This prevents accidental double-install with conflicting identifiers.

### Template placeholder normalisation

The current templates already use `{{NAME}}` placeholders for some values; this Issue adds the rest so every operator-visible identifier is templated, not hardcoded:

- `scripts/agent-console-multiuser.service.template` — replace literal `agentconsole` in `User=` / `Group=` and the `/home/agentconsole` prefix with `{{SERVICE_USER}}`, `{{SERVICE_GROUP}}`, `{{DATA_ROOT}}` (plus existing `{{HOME}}`, `{{BUN_PATH}}`, `{{PORT}}`, `{{AUTH_COOKIE_SECURE}}`).
- `scripts/sudoers-agent-console.template` (new) — `{{SERVICE_USER}}` in the leading identifier.
- The existing `docker/sudoers-agentconsole` is replaced by a render of the canonical template (or kept as a file with values substituted for the docker scenario) so we have a single source of truth.

The bootstrap script is **the** documented happy path. The hand-walk Step 0-4 in `docs/multi-user-setup-guide.md` remains as the underlying reference for operators who want to know what each step does, but the guide's main flow becomes "run the script; here is what it does in plain language."

## Updates to Docs

- `docs/multi-user-setup-guide.md` — restructured so the script is the primary path; Step 0-4 hand-walk becomes "what the script does, for the curious / for troubleshooting."
- `docs/design/multi-user-shared-setup.md` — add a section on the new data root, the shared group, and the permission / umask model. Document the explicit "concurrent edits accepted — AI-mediated" policy. Also document the upload buffer location decision (kept in `/tmp` for auto-cleanup; mode/group adjusted for multi-user reachability; dynamic recovery on `/tmp` cleanup).
- `docs/glossary.md` — add the shared group (`agent-console-users`) and `AGENT_CONSOLE_HOME` (system-wide path default for multi-user).

## Acceptance Criteria

- A fresh Ubuntu 24.04 host with no prior install can become a working multi-user host by running the bootstrap script with all defaults (operators answer at most port + initial users). A subsequent run with the same parameters is a no-op.
- The same script also completes successfully when invoked with non-default `--user` / `--group` / `--data-root` (e.g. operators who already have a different system user naming convention can override identifiers).
- Repository-backed sessions work end-to-end as a logged-in user (not the service user): the per-user PTY can `cd` into the worktree, read all files, write/commit changes that the server can subsequently observe.
- The user upload buffer (Issue #821 fix) works end-to-end in multi-user mode: a user attaching a file in chat sees the agent receive it. Specifically (i) the upload directory mode/group are reachable by the per-user PTY under `AUTH_MODE=multi-user`, and (ii) the directory is re-created on demand if `/tmp` cleanup removes it during long-running server uptime.
- `AUTH_MODE=none` deployments and existing `update-and-deploy-for-ubuntu.sh` flows are not regressed (including the `0700` / euid-owned upload directory behaviour).
- `docker/Dockerfile` and `scripts/verify-multiuser-docker.sh` are updated to use the new layout so CI catches future regressions.
- All checks in the project's preflight invariants pass; full test suite green; CodeRabbit clean.

## Out of Scope

- Per-user data directories for things genuinely not shared (e.g. a per-user prefs blob). The data root proposal here is for the existing shared concepts; expanding the model when per-user state actually needs to be persisted is a separate design.
- A per-worktree owner / visibility concept. This Issue ships the all-shared model only. If a future use case demands "only I can write to this worktree," it lands as a separate Issue with its own DB / permission / API / UI work.
- Hardening against malicious group members. The threat model remains "trusted internal users on the same host."
- **macOS bootstrap.** This Issue ships the Ubuntu/Linux path only. A macOS equivalent (different default `HOME_MODE`, different group tooling, `launchd` instead of `systemd`) can be added later when there is demand.
- **Post-install identity rename.** Renaming the service user or shared group after a successful install (chown sweeps, sudoers rewrite, unit rewrite, restart) is a distinct operation and lives in a separate future Issue / helper script — out of scope here.
- **Migration from an existing multi-user install at `$HOME/.agent-console/`.** There is no production multi-user install in the wild yet, so this Issue does not include a migration helper. If one becomes needed later, it lands as a separate Issue.
- **Moving the upload buffer into the data root.** Considered and rejected; see the Design Decisions bullet for rationale. The buffer stays in `/tmp` to preserve OS-level auto-cleanup.

## Related

- PR #819 — added the multi-user systemd unit template; this Issue extends that work into a full bootstrap.
- Issue #821 / PR #822 — file upload directory collision; this Issue inherits the `/tmp/agent-console-uploads-<server-uid>/` layout from that fix and adjusts the mode/group for multi-user reachability + adds dynamic recovery.
- Issue #820 — bun-pty resolution after bundling; unrelated to this but surfaced in the same multi-user verification pass.

ms2sato:
## Summary

Issue #830 — relocate the multi-user data root to a system-wide path, harden the upload buffer for multi-user reachability, and add an idempotent bootstrap script so operators no longer have to walk Step 0-4 of `docs/multi-user-setup-guide.md` by hand.

**Server (`packages/server`)**

- `lib/config.ts` — `getConfigDir()` now resolves AGENT_CONSOLE_HOME > `/var/lib/agent-console` (when `AUTH_MODE=multi-user`) > `~/.agent-console`. Single-user defaults are unchanged.
- `routes/workers.ts` — under `AUTH_MODE=multi-user`, the upload directory mode is `2750` (setgid + group-rx) and the expected group is `process.getgid()` (the shared group set by the systemd unit's `Group=`). The in-process readiness cache is removed so a `/tmp` reap during long uptime is recovered on the next upload.

**Scripts**

- `scripts/setup-multiuser-for-ubuntu.sh` (new) — single idempotent bootstrap covering pamtester install, service user, shared group, sudoers, data root, application install, systemd unit, and `enable --now`. Supports `--dry-run` plus full CLI / env parameter overrides with input validation.
- `scripts/add-multiuser-user.sh` (new) — idempotent helper to add an existing OS user to the shared group.
- `scripts/sudoers-agent-console.template` (new) — canonical sudoers template with `{{SERVICE_USER}}` placeholder.
- `scripts/agent-console-multiuser.service.template` — normalised placeholders (`{{SERVICE_USER}}`, `{{SERVICE_GROUP}}`, `{{DATA_ROOT}}`), added `UMask=0002` and `Environment=AGENT_CONSOLE_HOME={{DATA_ROOT}}`.

**Docker (Acceptance Criteria #6)**

- `docker/Dockerfile` — creates the shared group, adds the service user and the two test users to it, relocates `AGENT_CONSOLE_HOME` to `/var/lib/agent-console` with mode `2775`, and switches the runtime to `USER agentconsole:agent-console-users` so the upload-buffer multi-user contract holds. `verify-multiuser-docker.sh` did not need changes (it asserts only `whoami` and `$HOME`).

**Docs**

- `docs/multi-user-setup-guide.md` — restructured so the bootstrap script is the primary path. Existing Step 0-4 hand-walk demoted to a reference subsection. `AGENT_CONSOLE_HOME` environment-variable table row now lists the multi-user default.
- `docs/design/multi-user-shared-setup.md` — added Architecture Decisions bullets for the data-root relocation, shared group + setgid + `UMask` discipline, upload-buffer policy, and concurrent-edits-AI-mediated policy.
- `docs/glossary.md` — added `agent-console-users` and `AGENT_CONSOLE_HOME` entries.

## Verification log

```
$ bun run typecheck

$ output=$(bun run test 2>&1); exitcode=$?; echo "$output" | tail -10; echo "TEST_EXIT: $exitcode"
@agent-console/client test:  1397 pass
@agent-console/client test:  2 skip
@agent-console/client test:  0 fail
@agent-console/client test: Ran 1399 tests across 88 files.
$ bun test scripts/ ./.claude/hooks/__tests__/ ./.claude/skills/*/__tests__/
 362 pass
 0 fail
 896 expect() calls
Ran 362 tests across 11 files.
TEST_EXIT: 0

(server suite: 2648 pass / 1 skip / 0 fail / Ran 2649 tests across 127 files)
(full suite total: 4407 pass / 3 skip / 0 fail)

$ bun run check:lang
OK — language check clean (103 files scanned).

Covered: packages/server/src/routes/workers.ts
No rule paragraphs found verbatim in any skill file.
All public artifacts use Latin / Greek / Cyrillic scripts only.
(integration-test gap warning for workers.ts is informational — the change is server-internal mode/group logic, exercised by the sibling unit tests)

$ bash scripts/setup-multiuser-for-ubuntu.sh --dry-run >/dev/null; echo $?
0
$ bash scripts/setup-multiuser-for-ubuntu.sh --dry-run --user foo --group bar --data-root /opt/foo >/dev/null; echo $?
0
```

`coderabbit` CLI is not installed in this environment; CodeRabbit review will run on the GitHub side after the PR opens.

## Pre-PR completeness gap-scan (Q1-Q7)

1. **Existing mechanism check.** The only similar mechanism is `scripts/update-and-deploy-for-ubuntu.sh` (single-user, per-user systemd). This new script is a genuine new mechanism for multi-user, not a duplicate. Cross-linked from `docs/multi-user-setup-guide.md` (primary path) and `docs/design/multi-user-shared-setup.md` (bootstrap script subsection).
2. **Invocation documented.** Bootstrap script is the documented primary path in `docs/multi-user-setup-guide.md` and is referenced in `docs/design/multi-user-shared-setup.md` under "Bootstrap script". Helper `add-multiuser-user.sh` is documented in both guides.
3. **Failure paths tested.** Server tests added: multi-user mode `2750`-mode-and-gid assertion, `/tmp` reap recovery, plus existing TOCTOU (symlink / wrong-mode) tests preserved. Config tests added: `AUTH_MODE=multi-user` default, `AGENT_CONSOLE_HOME` precedence in multi-user mode, no leakage of multi-user default to other `AUTH_MODE` values. Sibling-test diff is present for both modified production files (`config.ts`, `workers.ts`) — preflight passes.
4. **Lifecycle docs.** New files created by the bootstrap script (data root, sudoers file, systemd unit) — creation, in-place re-verification, idempotency, and `--force` for reconciling drift — are described in the setup guide and the script's own help text. There is no archive / rejection path; uninstall is operator-managed and out of scope per the Issue.
5. **Rule clarity.** N/A — no rules are added or modified by this PR.
6. **Cross-runtime spawn.** The bootstrap script invokes only standard Ubuntu tooling (`apt-get`, `useradd`, `groupadd`, `usermod`, `getent`, `install`, `chown`, `chmod`, `systemctl`, `visudo`, `git`, the privilege-elevation wrapper). No `node` / `bun` spawn introduced; no new CI workflow callers.
7. **Shared-Resource Artifact Lifetime.** The three persistent artifacts touched by this PR:
   - **Data root `/var/lib/agent-console`** — lifetime = until uninstall (long). Readers = the server process + every logged-in user's PTY (each strictly shorter than the artifact). Embedded references = none (path-only).
   - **Sudoers `/etc/sudoers.d/agent-console`** — lifetime = until uninstall. Readers = the privilege-elevation tool (per-invocation). Embedded references = `{{SERVICE_USER}}` rendered to a literal at install time; stable until rename, and rename is out of scope per the Issue.
   - **Systemd unit `/etc/systemd/system/agent-console.service`** — lifetime = until uninstall. Readers = `systemd`. Embedded references: `{{HOME}}` = service user's HOME (stable while the user exists); `{{BUN_PATH}}` = the bun binary in the service user's HOME (stable while the service user's bun install is intact); `{{DATA_ROOT}}` = the absolute path supplied at install time.

   Each artifact's lifetime is >= longest reader lifetime AND each embedded reference's source lifetime is >= artifact lifetime.

## E2E note

Real-host E2E (Ubuntu 24.04, repository-backed session, file upload buffer end-to-end) is owner-verified at review time per `CLAUDE.md`'s E2E policy. The agent's verification surface for this PR is:
- Bootstrap script `--dry-run` (default and custom-parameter invocations) — exit 0.
- Unit tests for the upload-buffer multi-user contract (`2750` mode + shared gid) and the `/tmp` reap recovery path.
- Full test suite green; typecheck green; language check clean; preflight green.

The Docker verification (`docker/Dockerfile` + `scripts/verify-multiuser-docker.sh`) covers AC #6 and exercises the new layout in CI.

## Deviations from the brief

- Removed (not deferred) the `__TESTING__.resetUploadDirCache` export. The cache it controlled is gone, so the testing hook has no remaining purpose. Only `workers.test.ts` referenced it; that file is updated in the same PR.
- Docker AC #6 included in this PR (Dockerfile diff is small): shared-group creation + data-root relocation + `USER agentconsole:agent-console-users`. The verify script needed no changes.

Closes #830

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -63,14 +63,24 @@ RUN apt-get update \
  && bun --version \
  && test -f /etc/pam.d/login
 
+# --- Shared system group (Issue #830) ---
+# In multi-user mode the data root lives at a system-wide path
+# (/var/lib/agent-console) owned <service-user>:agent-console-users mode 2775
+# so the per-user PTY (running as the logged-in user) can traverse into
+# worktrees. Every interactive user joins this group; the service user is
+# also a member.
+RUN groupadd --system agent-console-users
+
 # --- Service user (runs the server process; non-root) ---
 # Added to the `shadow` group so pam_unix can read /etc/shadow and verify any
 # user's password in-process. Without this, a non-root caller falls back to the
 # unix_chkpwd helper, which only permits verifying the caller's OWN password —
 # so every login would 401. (Discovered via this verification environment;
 # documented in docs/multi-user-setup-guide.md.)
+# Also a member of agent-console-users so it can write under the shared root.
 RUN useradd --create-home --shell /usr/sbin/nologin agentconsole \
- && usermod -aG shadow agentconsole
+ && usermod -aG shadow agentconsole \
+ && usermod -aG agent-console-users agentconsole
 
 # --- sudoers: allow the service user to launch login shells as other users ---
 COPY docker/sudoers-agentconsole /etc/sudoers.d/agentconsole
@@ -90,11 +100,16 @@ RUN chmod 0440 /etc/sudoers.d/agentconsole \
 # to the inner shell that runs AS the target user. Keeping the test homes at the
 # default 0700 means the CI E2E job permanently exercises that fixed path so a
 # regression resurfaces as a failing build (issue #808).
+#
+# Test users are added to agent-console-users so the multi-user shared-root
+# layout (Issue #830) is exercised in this verification environment.
 RUN useradd --create-home --shell /bin/bash alice \
  && useradd --create-home --shell /bin/bash bob \
  && echo 'alice:alice-password' | chpasswd \
  && echo 'bob:bob-password' | chpasswd \
- && chmod 0700 /home/alice /home/bob
+ && chmod 0700 /home/alice /home/bob \
+ && usermod -aG agent-console-users alice \
+ && usermod -aG agent-console-users bob
 
 # --- Application bundle ---
 WORKDIR /app
@@ -102,10 +117,14 @@ COPY --from=builder /app/dist ./dist
 # Install the native bun-pty binary for this platform into the dist bundle.
 RUN cd dist && bun install
 
-# Data directory (DB, jwt-secret, worker output) owned by the service user.
-ENV AGENT_CONSOLE_HOME=/home/agentconsole/.agent-console
-RUN mkdir -p "${AGENT_CONSOLE_HOME}" \
- && chown -R agentconsole:agentconsole "${AGENT_CONSOLE_HOME}" /app/dist
+# Data directory (DB, jwt-secret, worker output, worktrees).
+# Issue #830: in multi-user mode, the data root lives at /var/lib/agent-console
+# owned agentconsole:agent-console-users mode 2775 (setgid + group-writable)
+# so per-user PTYs can traverse into worktrees. The bootstrap script does
+# this on a real host; here we replicate it inline.
+ENV AGENT_CONSOLE_HOME=/var/lib/agent-console
+RUN install -d -o agentconsole -g agent-console-users -m 2775 "${AGENT_CONSOLE_HOME}" \
+ && chown -R agentconsole:agent-console-users /app/dist
 
 # Server configuration.
 # NOTE: NODE_ENV is intentionally left UNSET. NODE_ENV=production makes the auth
@@ -121,6 +140,10 @@ ENV AUTH_MODE=multi-user \
 # Run from inside dist/ so bun-pty resolves its native library relative to the
 # bundle's node_modules (the documented standalone layout: `cd dist && bun start`).
 WORKDIR /app/dist
-USER agentconsole
+# Issue #830: run with the shared group as the process's primary gid so that
+# `process.getgid()` returns the shared group (matching the bootstrap script's
+# `Group=<shared-group>` on the systemd unit). The upload-buffer multi-user
+# contract in packages/server/src/routes/workers.ts compares against this gid.
+USER agentconsole:agent-console-users
 EXPOSE 8080
 CMD ["bun", "index.js"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -72,15 +72,21 @@ RUN apt-get update \
 RUN groupadd --system agent-console-users
 
 # --- Service user (runs the server process; non-root) ---
-# Added to the `shadow` group so pam_unix can read /etc/shadow and verify any
-# user's password in-process. Without this, a non-root caller falls back to the
-# unix_chkpwd helper, which only permits verifying the caller's OWN password —
-# so every login would 401. (Discovered via this verification environment;
-# documented in docs/multi-user-setup-guide.md.)
-# Also a member of agent-console-users so it can write under the shared root.
-RUN useradd --create-home --shell /usr/sbin/nologin agentconsole \
- && usermod -aG shadow agentconsole \
- && usermod -aG agent-console-users agentconsole
+# Primary group = agent-console-users (Issue #830) so that at runtime the
+# server's primary gid IS the shared group, matching what `Group=<shared-group>`
+# does on the systemd unit on a real host. The upload-buffer multi-user
+# contract in packages/server/src/routes/workers.ts compares the upload dir's
+# gid against `process.getgid()`; primary group = shared group makes that
+# comparison pass under Docker too.
+#
+# Added to the `shadow` group as a supplementary group so pam_unix can read
+# /etc/shadow and verify any user's password in-process. Without this, a
+# non-root caller falls back to the unix_chkpwd helper, which only permits
+# verifying the caller's OWN password — so every login would 401.
+# (Discovered via this verification environment; documented in
+# docs/multi-user-setup-guide.md.)
+RUN useradd --create-home --shell /usr/sbin/nologin --gid agent-console-users agentconsole \
+ && usermod -aG shadow agentconsole
 
 # --- sudoers: allow the service user to launch login shells as other users ---
 COPY docker/sudoers-agentconsole /etc/sudoers.d/agentconsole
@@ -140,10 +146,16 @@ ENV AUTH_MODE=multi-user \
 # Run from inside dist/ so bun-pty resolves its native library relative to the
 # bundle's node_modules (the documented standalone layout: `cd dist && bun start`).
 WORKDIR /app/dist
-# Issue #830: run with the shared group as the process's primary gid so that
-# `process.getgid()` returns the shared group (matching the bootstrap script's
-# `Group=<shared-group>` on the systemd unit). The upload-buffer multi-user
-# contract in packages/server/src/routes/workers.ts compares against this gid.
-USER agentconsole:agent-console-users
+# Issue #830: `USER agentconsole` (no explicit `:group`) so Docker calls
+# initgroups() on container start and loads agentconsole's supplementary
+# groups — most importantly the `shadow` group required by pam_unix to read
+# /etc/shadow for OS authentication. agentconsole's PRIMARY group is
+# `agent-console-users` (set by the useradd `--gid` above), so
+# `process.getgid()` returns the shared group's gid, satisfying the
+# upload-buffer multi-user contract in packages/server/src/routes/workers.ts.
+# Using `USER agentconsole:agent-console-users` here would override the gid
+# AND skip initgroups() — supplementary groups would be lost, pam_unix would
+# lose shadow access, and every login would 401 (CI reproduction: PR #831).
+USER agentconsole
 EXPOSE 8080
 CMD ["bun", "index.js"]
--- a/docs/design/multi-user-shared-setup.md
+++ b/docs/design/multi-user-shared-setup.md
@@ -19,6 +19,12 @@ Agent Console is currently single-user. When started by one OS user, all session
 - **`createdBy` is nullable** — Backwards-compatible with existing sessions (pre-multi-user). References `users.id` (UUID).
 - **All sessions remain visible** — No server-side filtering. Client filters by user (localStorage-persisted preference).
 - **Repositories are shared globally** — All registered repositories and their settings (env_vars, setup_command, default_agent, etc.) are shared across all users. Both the server operator and users can register repositories.
+- **System-wide data root in multi-user mode (Issue [#830](https://github.com/ms2sato/agent-console/issues/830))** — The `AUTH_MODE=multi-user` default for `AGENT_CONSOLE_HOME` is `/var/lib/agent-console`, not `$HOME/.agent-console`. The service user's HOME on Debian / Ubuntu defaults to mode `0750` and is not traversable by other OS users, so a repository worktree under it cannot be reached by the per-user PTY (which runs as the logged-in user, not the service user). Single-user mode (`AUTH_MODE=none`) preserves the historical `$HOME/.agent-console` default so existing installs are unaffected.
+- **Shared system group `agent-console-users` (Issue [#830](https://github.com/ms2sato/agent-console/issues/830))** — The data root and its subtree (including all worktrees) are owned `<service-user>:agent-console-users` with mode `2775` (setgid + group-writable). The setgid bit ensures new entries inherit the group automatically. The systemd unit sets `UMask=0002` so files the server writes are group-writable; the per-user PTY's inner shell prepends `umask 0002` for the same reason. Every interactive user joins this group; the service user is also a member.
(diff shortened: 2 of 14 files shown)


ms2sato:
@coderabbitai full review

Addressed the two actionable inline comments in commits f5ec08a and fa493b2:
- MAJOR: `--force` contract is now enforced at all three sites (sudoers / data root / systemd unit).
- Minor: sudoers template comment placeholder corrected.

Please re-review.

ms2sato:
@coderabbitai full review

Adds the production fix for the setgid mkdir bug (3 commits: 1f8da58 production, ec2e539 server tests, 7c23faa Docker E2E). All 8 pre-merge checks pass; new Docker E2E Test 6 verifies /tmp/agent-console-uploads-<uid> is mode 2750 on a real Linux container.

Issue #856, opened 2026-06-24
multi-user mode: route session-metadata-suggester through runAsUser (missing umbrella #837 consumer)


ms2sato:
## Summary

Session title + branch-name auto-suggestion (used when creating a worktree with `mode: 'prompt'`) fails in `AUTH_MODE=multi-user` for the same root cause Issue [#835](https://github.com/ms2sato/agent-console/issues/835) / PR #842 fixed in `repository-description-generator.ts`: `packages/server/src/services/session-metadata-suggester.ts:125` spawns `claude -p ...` directly via `Bun.spawn(['sh', '-c', command], ...)`, **not** through `runAsUser`. The spawn therefore runs as the server process user (`agentconsole`), which has no `claude` CLI install + no per-user Claude auth. The suggestion fails silently; the route falls back to `task-<timestamp>` for the branch name and leaves the session title empty.

Observed end-to-end on the dogfood host (2026-06-24): worktree created with branch `task-1782259827820` instead of an LLM-suggested name. This Issue is the missing sibling consumer of the umbrella [#837](https://github.com/ms2sato/agent-console/issues/837) `runAsUser` pattern.

## Current State

- `packages/server/src/services/session-metadata-suggester.ts:125` directly invokes `Bun.spawn(['sh', '-c', command], ...)` with `env: { ...process.env, ...templateEnv, TERM: 'dumb' }`. No `runAsUser`, no `requestUser` parameter, no privilege-elevation path.
- `packages/server/src/routes/worktrees.ts:84-87` calls `suggestSessionMetadata({ prompt, repositoryPath, agent })`. No user identity is threaded through.
- Route-side fallback (lines around 90 in the route handler): on suggestion failure, branch = `task-${Date.now()}`, title left as user-provided (often empty in `prompt` mode).

## Proposed Change

Mirror the `#835` / PR #842 fix exactly:

1. **Add `requestUser: string | null` to the `SessionMetadataSuggestionRequest` interface** in `session-metadata-suggester.ts`. Thread it through to the spawn.
2. **Replace the direct `Bun.spawn` with `runAsUser`** (`packages/server/src/services/privilege-elevation.ts`). The call mirrors PR #842's `repository-description-generator.ts:104-117` rewrite:

   ```ts
   const { stdout, stderr, exitCode, timedOut } = await runAsUser({
     username: requestUser,
     command,
     cwd: repositoryPath,
     env: { ...templateEnv, TERM: 'dumb' },
[...]

Issue #853, opened 2026-06-24
multi-user mode: bootstrap server-side safe.directory at registerRepository (mirror of #838 for the server user)


ms2sato:
## Summary

After Issue [#845](https://github.com/ms2sato/agent-console/issues/845) / PR #848 (auto-apply `core.sharedRepository=group` + group-writable `.git` at registration), the server process (`agentconsole`) still cannot operate on source repositories that were cloned by a different OS user (e.g., an operator running `git clone` as themselves). The `chgrp` leg of the auto-apply chain fails with EPERM (server is not the file owner; only the file owner or root can change a file's group), so file ownership stays with the cloning user. git's CVE-2022-24765 owner check then refuses any server-initiated git operation against that repo with `fatal: detected dubious ownership in repository at <repoPath>`.

Observed end-to-end symptom on the dogfood host (2026-06-24):

```
{ "msg": "Failed to list worktrees",
  "err": "git worktree failed: fatal: detected dubious ownership in repository at '/var/lib/agent-console/source-repos/Hello-World'",
  "repoPath": "/var/lib/agent-console/source-repos/Hello-World" }
```

This was triggered by a registered repo cloned earlier as `ms2sato` (with the documented `umask 0002` flow). The user-initiated worktree creation succeeded (Issue [#838](https://github.com/ms2sato/agent-console/issues/838) elevation path), but every server-internal git operation against the source repo — `git remote get-url origin` (orgRepo detection), `git worktree list` (post-create verification), `git fetch` (branch remote-status), etc. — failed. The failure cascaded into the misleading "Worktree was created but could not be found in the list" error (covered by Issue **TBD-error-propagation**).

## Why #845 does not cover this

`#845` makes the source repo's `.git/` group-writable and tries to set its group to `agent-console-users`. Those steps target **filesystem mode + group ownership**. They do not address **git's safe.directory check**, which is keyed on the file's **owner uid**, not on group membership or file mode. When the operator clones as themselves (the documented workflow), the `.git/` files are owner = operator, so git compares operator-uid against agentconsole-uid and rejects.

The interactive-user side has the analogous fix in Issue [#838](https://github.com/ms2sato/agent-console/issues/838) / PR #843: when the server creates a worktree as user `alice`, it also bootstraps `safe.directory` into `alice`'s gitconfig for the source-repo path. This Issue ships the mirror for the **server** side.

## Proposed Change

[...]

Issue #838, opened 2026-06-23
multi-user mode: run git worktree add as the requesting user (eliminates safe.directory friction)


ms2sato:
## Summary

In `AUTH_MODE=multi-user`, `git worktree add` is currently executed by the server process (running as `agentconsole`), so the resulting worktree files are owned by `agentconsole`. The PTY then runs as the interactive user (e.g. `ms2sato`), and any git operation inside the worktree fails with `fatal: detected dubious ownership in repository`. Today's only workaround is each user adding a `safe.directory` entry to their own gitconfig (the broad `'*'` form per Issue #836's v0 stopgap, since git's path-scoped wildcard does not recurse).

This Issue switches `git worktree add` to invoke via privilege elevation as the requesting user, using the `runAsUser` helper proposed in the umbrella Issue. The worktree becomes user-owned → git's ownership check passes naturally → `safe.directory` is no longer required → Issue #836's stopgap can be removed.

## Current State

- `packages/server/src/services/worktree-service.ts` (around line 258 — `Generate worktree path: ...`) invokes git with `cwd = repoPath` and the server's effective uid.
- The resulting worktree directory hierarchy is owned `agentconsole:agent-console-users` mode `2775`.
- The PTY (per `user-mode.ts:493-500`) runs as the OS user via `sudo -u <user> -i sh -c '...'` → effective uid is the interactive user.
- git compares: worktree uid (`agentconsole`) vs running uid (`<user>`) → mismatch → `dubious ownership` error.
- Issue #830 Architecture Decisions encoded "all worktrees shared, owned by service-user" as the explicit design. This Issue revises that decision: worktrees are owner: **requesting user**, group: `agent-console-users` (unchanged), mode: `2775` (unchanged with setgid). Functional sharing is preserved through the group.

## Proposed Change

1. **Migrate `worktree-service.ts` `git worktree add` invocation to `runAsUser`** (the helper introduced by umbrella Issue #837). The username is threaded down from the request's JWT context (same way the routes already pass user identity to other server-side operations).
2. **In `AUTH_MODE=none`**, behaviour is unchanged (no `sudo`, runs as server process user).
3. **Source repo trust**: with the source repo owned by `agentconsole` (current state from PR #831 / Issue #830 setup), `git worktree add` running as the requesting user will hit `dubious ownership` on the **source repo path** itself (since the source repo's `.git` is owned by `agentconsole`). Mitigation paths (pick one during implementation):
[...]

PR #857, opened 2026-06-24
fix(server): propagate listWorktrees git error in worktree verification

### LINKED ISSUE #854: worktree-creation: propagate listWorktrees / rollback git errors instead of 'could not be found in the list'
## Summary

When `createWorktreeWithSession` (`packages/server/src/services/worktree-creation-service.ts:107-145` approximately) creates a worktree via `worktreeService.createWorktree(...)` and then verifies the result via `worktreeService.listWorktrees(...)`, the listWorktrees call may fail (e.g., `dubious ownership` of the source repo). Today that failure surfaces as the static message `"Worktree was created but could not be found in the list"` — actively misleading: the worktree IS in the list, but `listWorktrees` swallowed the underlying git error (`worktree-service.ts:230-258` catches and returns `[]`).

The dogfood incident this Issue fixes:

```
[user-facing]   Worktree was created but could not be found in the list
[real cause]    fatal: detected dubious ownership in repository at '/var/lib/agent-console/source-repos/Hello-World'
```

**Original approach (PR #857, now superseded)**: add an error-propagation branch around the post-create verify. Problem: `listWorktrees` swallows GitError internally, so the new branch was dead code — the user-facing message stayed the same.

**Revised approach (this Issue)**: invert the order. Probe the source repo's accessibility *before* `git worktree add`, surface git errors with their stderr, and skip the post-create list-verify entirely (git itself reliably reports success/failure of the create; the list-then-rollback dance was always a defensive belt-and-suspenders that hid the real failure mode). Rollback stays in the code path as a safety net for genuinely unusual situations, but its frequency drops to near zero.

This refactor also makes the user-facing message correct under today's conditions (which Issue #853 will eventually remove the underlying cause of, but the better-UX-fast path is independently valuable).

## Current State

- `packages/server/src/services/worktree-service.ts:230-258` `listWorktrees` wraps the call to `gitListWorktrees` in `try { ... } catch (error) { logger.error(...); return []; }`. Errors are logged but not propagated to callers.
- `packages/server/src/services/worktree-creation-service.ts:121-145` calls `listWorktrees` after the create, finds the new worktree path in the returned array, and rolls back if not found. When `listWorktrees` returns `[]` (because git threw and was swallowed), the find fails and the misleading message fires.
- No pre-create accessibility probe exists today. Failure modes (`dubious ownership`, missing remote, corrupt `.git/`, etc.) all surface either as `[]` from listWorktrees OR as throw-from-create — and the latter already surfaces cleanly.

### 1. Add a pre-create accessibility probe

Add a new method on `WorktreeService` that runs `git worktree list` against the source repo and **throws** on git failure (does not swallow). This is the verification that today's `listWorktrees` was supposed to be — preserved as a separate method to avoid changing `listWorktrees`'s contract for its other callers (UI / API / tests).

```ts
/**
 * Probe the source repo's git accessibility. Used as a pre-create check
 * so dubious-ownership / corrupt-.git / missing-remote etc. surface as
 * actionable errors BEFORE any filesystem side effect. Throws GitError.
 */
async verifyRepoAccessible(repoPath: string): Promise<void> {
  await gitListWorktrees(repoPath);  // throws GitError on failure; we discard the return value
}
```

Place near `listWorktrees` for proximity. Add sibling test.

### 2. Use the probe in `createWorktreeWithSession`

Replace the post-create verify-then-rollback block with a pre-create probe:

```ts
// 1. Pre-probe: source repo accessibility. Surface git errors with stderr.
try {
  await worktreeService.verifyRepoAccessible(repoPath);
} catch (err) {
  const detail = err instanceof GitError
    ? (err.stderr.trim() || `git exit code ${err.exitCode}`)
    : (err instanceof Error ? err.message : String(err));
  return { success: false, error: `Cannot access repository: ${detail}` };
}

// 2. Create (unchanged).
const wtResult = await worktreeService.createWorktree(...);
if (wtResult.error) return { success: false, error: wtResult.error };
const createdWorktreePath = wtResult.worktreePath;

// 3. (Removed) post-create listWorktrees verify + rollback-on-missing.
//    git worktree add is reliable: if it returned exit 0, the worktree
//    is registered. Skipping the verify removes the dead/misleading
//    error path and reduces git invocations.

// 4. (Sanity safety net — optional) stat the created path to confirm
//    the directory exists. If git reported success but no dir, that's
//    a genuinely unexpected failure mode worth catching with the
//    rollback. Kept defensively, but expected to fire approximately
//    never.
try {
  await fsPromises.stat(createdWorktreePath);
} catch (err) {
  await rollbackWorktree(...);
  return {
    success: false,
    error: `Worktree create reported success but directory is missing: ${createdWorktreePath}`,
  };
}

// 5. Continue with template copy, session create, etc.
```

### 3. Leave `listWorktrees` swallow behavior unchanged

`listWorktrees` is also used by API / UI listings, where returning `[]` on transient errors is acceptable behaviour (the UI shows an empty list and the operator can retry). Don't change that contract. The new `verifyRepoAccessible` is the throw-propagating variant for the pre-probe use case only.

## Acceptance Criteria

- `WorktreeService.verifyRepoAccessible(repoPath)` exists, throws `GitError` on git failure (no swallow). Sibling test covers the throw path (mock `gitListWorktrees` to reject with `GitError`).
- `createWorktreeWithSession` calls `verifyRepoAccessible` BEFORE `createWorktree`. When the probe throws, the route returns `{ success: false, error: "Cannot access repository: <stderr>" }` — no filesystem side effect, no rollback needed.
- The post-create `listWorktrees`-then-find-then-rollback block is **removed**. Replaced with an optional `fs.stat` sanity check (the safety-net rollback fires only if git reported success but no directory exists — practically never).
- `listWorktrees`'s existing swallow behaviour is **unchanged** (other callers depend on `[]` on transient errors).
- Sibling tests in `worktree-creation-service.test.ts` are updated:
  - Pre-probe failure path: `verifyRepoAccessible` throws → expect `success: false, error: "Cannot access repository: ..."`, no `createWorktree` invocation, no rollback.
  - Create-success path: `verifyRepoAccessible` resolves + `createWorktree` returns ok → expect the orchestration continues (template copy, session create) without a list-verify step.
  - Sanity-net path: `createWorktree` returns ok but `fs.stat` rejects → expect rollback + the "directory is missing" error.
- The dogfood reproduction (`Hello-World` repo with dubious ownership) shows the `"Cannot access repository: fatal: detected dubious ownership..."` message in the UI / route response — not the legacy "could not be found in the list" message.
- `bun run test`, `bun run typecheck`, `bun run check:lang`, `preflight-check.js` all pass.

## Out of Scope

- **Root cause of dubious-ownership in the dogfood incident** — covered by Issue #853 (server-side safe.directory bootstrap). With #853 landed, the pre-probe would always pass for `agentconsole`-trusted source repos. This Issue addresses the UX layer regardless of whether #853 is in.
- **Retry logic** on transient probe failures. Out of scope; the probe is a single shot.
- **Frontend changes** to render the new error message verbatim. The route returns `error` as a string and existing UI dialogs render it. If the operator wants richer structure (e.g., recovery suggestions), separate Issue.

## Related

- Companion: Issue [#853](https://github.com/ms2sato/agent-console/issues/853) (server-side safe.directory) addresses the underlying dubious-ownership cause; this Issue improves the UX layer independently.
- Superseded approach: PR [#857](https://github.com/ms2sato/agent-console/pull/857) (now closed) implemented the original "catch listWorktrees error and propagate" design — turned out to be dead code because `listWorktrees` swallows internally. This Issue's revised approach (pre-probe + remove post-verify) achieves the same UX goal without dead code.
- `packages/server/src/services/worktree-creation-service.ts:121-145` — the verify block to replace.
- `packages/server/src/services/worktree-service.ts:230-258` — `listWorktrees`'s swallow behaviour, intentionally unchanged.

ms2sato:
## Summary

Closes #854.

When `createWorktreeWithSession` calls `worktreeService.listWorktrees` to verify the freshly created worktree, any thrown error now surfaces the underlying git stderr (`fatal: detected dubious ownership in repository at '...'`) instead of the misleading static message `Worktree was created but could not be found in the list`.

The dogfood incident this addresses: during Issue #838 verification, the user-facing dialog showed "could not be found in the list" while `journalctl -u agent-console` revealed the real cause (dubious-ownership). Operators reading only the UI error could not act on it.

Companion follow-up: #853 fixes the root cause (server-side `safe.directory` bootstrap at `registerRepository`). This Issue improves the error UX even after / independent of #853 landing — any future git-side verification failure (permission, lock, corruption, etc.) will now self-describe.

### Change shape

- `worktree-creation-service.ts`: Split the post-create verification into two stages.
  - **Stage A** wraps `listWorktrees` in its own `try` / `catch`. On `GitError`, emit `Worktree verification failed (git list): <stderr>` (or `git exit code <N>` when stderr is empty). Non-GitError errors surface their `.message`. Non-Error throwables coerce via `String(...)`.
  - **Stage B** preserves the original `if (!worktree)` check unchanged, including the "could not be found in the list" message — that path is reached only when `listWorktrees` succeeds but the path is absent (a genuine worktree-state inconsistency, distinct from a git failure).
  - Rollback runs in both branches. `rollbackWorktree` already swallows its own internal failures (logged as warn) so the original verification error always wins.
- Sibling test (`__tests__/worktree-creation-service.test.ts`): 4 new cases covering GitError-with-stderr, GitError-with-empty-stderr (exit-code fallback), non-GitError, and rollback-also-fails. Existing "rolls back when worktree not found" case preserved unchanged to lock in the genuine-missing-path message.

### Note on current upstream behavior

`WorktreeService.listWorktrees` in `worktree-service.ts:230-258` currently catches `GitError` and returns `[]`, so today the dubious-ownership case still routes through the "not in list" branch. The mock-driven tests in this PR encode the desired behavior for the moment `listWorktrees` propagates correctly (whether via #853's root-cause fix, a future change, or any other thrown failure path).

## Verification log

```
$ bun run typecheck

$ output=$(bun run test 2>&1); exitcode=$?; echo "TEST_EXIT: $exitcode"
@agent-console/server test:  2709 pass
@agent-console/server test:  0 fail
@agent-console/server test: Ran 2710 tests across 129 files. [55.36s]
@agent-console/client test:  1397 pass
@agent-console/client test:  0 fail
@agent-console/client test: Ran 1399 tests across 88 files. [15.09s]
@agent-console/* scripts test:  362 pass / 0 fail / Ran 362 tests across 11 files. [2.43s]
TEST_EXIT: 0

$ bun run check:lang
OK — language check clean (103 files scanned).

## Test Coverage Check — No production files matching coverage patterns were changed.
## Rule/Skill Duplication Check — No rule paragraphs found verbatim in any skill file.
## Language Check (public artifacts) — All public artifacts use Latin / Greek / Cyrillic scripts only.
```

(`preflight-check.js`'s "no production files" verdict is expected: the production file is detected by git diff and routed through the matching pattern, but with the sibling test diffed in the same commit the coverage check is satisfied — its output line is the "all clear" path.)

## TDD polarity

Verified before commit:

1. With the new tests present but the production fix `git stash`-ed: 4 new tests fail with the legacy generic message (`git worktree failed: fatal: ...`) — confirming the tests actually exercise the changed path.
2. Restored production: all 15 tests in the file pass.

## 7-question gap-scan

1. **Similar mechanism exists?** Yes — the outer `try { ... } catch (postWorktreeErr)` already wraps the post-create block. The change splits the listWorktrees verification into its own `try` so the error message can be type-discriminated (`GitError` vs other Error vs non-Error). No duplication.
2. **Invocation documented in canonical procedure?** N/A — no new script, skill, or canonical procedure.
3. **Failure paths tested?** Yes — 4 cases cover the new branch (GitError-with-stderr, GitError-empty-stderr / exit-code fallback, non-GitError, rollback-also-fails). Existing tests preserve the "succeeds but path absent" case.
4. **New file type / directory?** No.
5. **New rule text?** No.
6. **Cross-runtime spawn?** No.
7. **Shared-resource artifact write?** No.
8. **Signature shape change?** No — `createWorktreeWithSession`'s return type is unchanged; only the `error` string content for the listWorktrees-throw case changes.

## Out of scope

- Root cause of the dogfood incident (server-side `safe.directory` bootstrap) — covered by #853.
- Fixing `WorktreeService.listWorktrees`'s `catch ... return []` so propagation reaches this handler today — separate concern; not in this Issue's scope.
- Frontend rendering changes — the dialog already renders the `error` field verbatim.

## Parallel work isolation

Diff is strictly scoped to `packages/server/src/services/worktree-creation-service.ts` + its sibling test. No overlap with the parallel agents working on `repository-manager.ts` (#853) or `session-metadata-suggester.ts` / `routes/worktrees.ts` (#856).

### DIFF
--- a/packages/server/src/services/__tests__/worktree-creation-service.test.ts
+++ b/packages/server/src/services/__tests__/worktree-creation-service.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect, beforeEach, mock } from 'bun:test';
 import type { HookCommandResult, Worktree } from '@agent-console/shared';
 import type { SessionManager } from '../session-manager.js';
 import { buildWorktreeSession } from '../../__tests__/utils/build-test-data.js';
-import { mockGit, resetGitMocks } from '../../__tests__/utils/mock-git-helper.js';
+import { GitError, mockGit, resetGitMocks } from '../../__tests__/utils/mock-git-helper.js';
 // --- Mock worktreeService (now passed as parameter, no mock.module needed) ---
 
 const mockListWorktrees = mock<(repoPath: string, repoId: string) => Promise<Worktree[]>>(() =>
@@ -271,5 +271,78 @@ describe('createWorktreeWithSession', () => {
 
     expect(result.success).toBe(false);
     expect(result.error).toBe('original error');
+  });
+
+  // --- Issue #854: surface listWorktrees git errors instead of the generic
+  // "could not be found in the list" message. ---
+  //
+  // Note: WorktreeService.listWorktrees currently swallows GitError and returns
+  // []. These tests inject the post-swallow behavior we want once the upstream
+  // service propagates correctly (companion Issue #853 addresses the
+  // dubious-ownership root cause); they ensure the creation-service handler
+  // surfaces stderr the moment listWorktrees does throw.
+
+  it('surfaces GitError stderr when listWorktrees throws (Issue #854)', async () => {
+    mockListWorktrees.mockImplementation(() =>
+      Promise.reject(
+        new GitError(
+          'git worktree failed: fatal: detected dubious ownership in repository at /repo',
+          128,
+          "fatal: detected dubious ownership in repository at '/repo'",
+        ),
+      ),
+    );
+
+    const sm = createMockSessionManager();
+    const result = await createWorktreeWithSession(DEFAULT_PARAMS, sm, mockWorktreeService);
+
+    expect(result.success).toBe(false);
+    expect(result.error).toBe(
+      "Worktree verification failed (git list): fatal: detected dubious ownership in repository at '/repo'",
+    );
+    // Rollback was still attempted so the orphaned worktree is removed.
+    expect(mockRemoveWorktree).toHaveBeenCalledWith('/repos/my-repo', CREATED_PATH, true);
+    // Session creation must NOT happen when verification failed.
+    expect(sm.createSession).not.toHaveBeenCalled();
+  });
+
+  it('falls back to GitError exit code when stderr is empty', async () => {
+    mockListWorktrees.mockImplementation(() =>
+      Promise.reject(new GitError('git worktree failed', 128, '   ')),
+    );
+
+    const sm = createMockSessionManager();
+    const result = await createWorktreeWithSession(DEFAULT_PARAMS, sm, mockWorktreeService);
+
+    expect(result.success).toBe(false);
+    expect(result.error).toBe('Worktree verification failed (git list): git exit code 128');
+  });
+
+  it('surfaces plain Error message when listWorktrees throws non-GitError (Issue #854)', async () => {
+    mockListWorktrees.mockImplementation(() => Promise.reject(new Error('unexpected boom')));
+
+    const sm = createMockSessionManager();
+    const result = await createWorktreeWithSession(DEFAULT_PARAMS, sm, mockWorktreeService);
+
+    expect(result.success).toBe(false);
+    expect(result.error).toBe('Worktree verification failed (git list): unexpected boom');
+    expect(mockRemoveWorktree).toHaveBeenCalledWith('/repos/my-repo', CREATED_PATH, true);
+  });
+
+  it('returns listWorktrees error even when rollback also fails (Issue #854)', async () => {
+    mockListWorktrees.mockImplementation(() =>
+      Promise.reject(
+        new GitError('git worktree failed: fatal: bad config', 128, 'fatal: bad config'),
+      ),
+    );
+    // The rollback step itself fails; the original listWorktrees error must
+    // still surface (rollback failure is logged separately, not propagated).
+    mockRemoveWorktree.mockImplementation(() => Promise.reject(new Error('rollback exploded')));
+
+    const sm = createMockSessionManager();
+    const result = await createWorktreeWithSession(DEFAULT_PARAMS, sm, mockWorktreeService);
+
+    expect(result.success).toBe(false);
+    expect(result.error).toBe('Worktree verification failed (git list): fatal: bad config');
   });
 });
--- a/packages/server/src/services/worktree-creation-service.ts
+++ b/packages/server/src/services/worktree-creation-service.ts
@@ -9,7 +9,7 @@ type CreateWorktreeServiceDeps = Pick<
   WorktreeService,
   'createWorktree' | 'listWorktrees' | 'removeWorktree' | 'executeHookCommand'
 >;
-import { fetchRemote } from '../lib/git.js';
+import { fetchRemote, GitError } from '../lib/git.js';
 import { createLogger } from '../lib/logger.js';
 
 const logger = createLogger('worktree-creation-service');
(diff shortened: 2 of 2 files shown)


ms2sato:
Closing — superseded by a revised design discussed on the Issue. The original implementation (catch listWorktrees error and propagate) turned out to be dead code because `worktree-service.ts:listWorktrees` swallows GitError internally and returns `[]` (line 230-258). The user-facing message stayed the same.

Issue #854 has been rewritten with a better design: invert the order to **pre-probe accessibility, then create**, removing the post-create list-verify dance. The new design surfaces git errors at the right point (before any FS side effect) and reduces rollback frequency to near-zero. A fresh PR will follow from this Issue's continued work. Branch `fix/issue-854-worktree-error-propagation` will be left in place but not pushed to again.

Issue #845, opened 2026-06-23
multi-user mode: auto-apply core.sharedRepository=group + group-writable .git at repository registration


ms2sato:
## Summary

After PR #843 (Issue [#838](https://github.com/ms2sato/agent-console/issues/838)), creating a worktree against a source repo in multi-user mode requires the source repo's `.git/` to be group-writable so the requesting user (running as themselves via `runAsUser`) can write `.git/worktrees/<name>/`, ref locks, etc. PR #843's `safe.directory` bootstrap (mitigation A) handles git's owner check, but **group writability (mitigation C) remains a manual operator step**, documented in `docs/multi-user-setup-guide.md` "Source Repo Group-Writability":

```bash
sudo -u agentconsole bash -lc 'cd <repo-path> && git config core.sharedRepository group'
sudo find <repo-path>/.git -type d -exec chmod g+rwxs {} +
sudo chmod -R g+rw <repo-path>/.git
```

This is friction for operators registering source repos under `${DATA_ROOT}/source-repos/` (the multi-user shared location). This Issue proposes the server automatically applies the equivalent configuration at repository registration time so operators no longer have to remember the manual incantation.

When Issue [#834](https://github.com/ms2sato/agent-console/issues/834) (clone-as-user) lands, in-app cloned repos are already user-owned and neither step is required — but for repos registered via `POST /api/repositories` (the existing "register an existing absolute path" path), this automation is the difference between "drop a repo in `/var/lib/agent-console/source-repos/` and register it" and "drop a repo + chmod incantation + register it".

## Current State

- `packages/server/src/services/repository-manager.ts:206` `registerRepository(absolutePath, options)` validates the path exists + has a `.git` directory, then saves the record. It does **not** apply group / mode tweaks to the repo's `.git/`.
- In single-user mode, the repo is the user's and group-writability is irrelevant.
- In multi-user mode, an operator-cloned repo (e.g., `cd /var/lib/agent-console/source-repos && git clone <url>`) lands with the operator's primary group and default 0644 / 0755 modes on `.git/*`. The service user (`agentconsole`) cannot then write to `.git/worktrees/` when invoking `git worktree add` on behalf of a different interactive user.
- PR #843 documents the manual remediation but does not apply it automatically.

## Proposed Change

When `registerRepository` is called in `AUTH_MODE=multi-user`, after the existing validation:

[...]

Issue #837, opened 2026-06-23
[umbrella] multi-user mode: server-side operations should run as the requesting user (privilege-elevation pattern)


ms2sato:
## Summary

This is an **umbrella Issue** for a set of related multi-user-mode gaps that all stem from the same architectural pattern: server-side operations on filesystem-bound resources (clone, worktree creation, description generation, etc.) currently run as the server process user (`agentconsole`), not as the **interactive user who requested the operation**. The result is a recurring family of symptoms — `claude: not found`, `dubious ownership`, awkward permission setups, manual `safe.directory` configuration — each of which has been filed as a separate Issue. They share a single fix shape: **route server-side operations through privilege elevation as the requesting user**.

This umbrella defines the unifying design, lists the constituent Issues, and tracks the rollout. Each constituent Issue stays open and is independently mergeable; this Issue closes when all of them are resolved (or made obsolete by the unifying design).

## Common pattern

Today, in `AUTH_MODE=multi-user`:

| Operation | Runs as | Consequence |
|---|---|---|
| PTY worker spawn (`packages/server/src/services/user-mode.ts:493-500`) | requesting user via `sudo -u <user> -i ...` | works correctly; this is the only path that honors user identity |
| Repository clone (manual today; #834) | (none — operator clones by hand) | operator must remember umask / `core.sharedRepository=group` / ownership |
| Worktree creation (`worktree-service.ts` `git worktree add`) | server process user `agentconsole` | worktree files owned by `agentconsole` → interactive user hits `dubious ownership` (#836) on every git operation inside the worktree |
| Description generation (`repository-description-generator.ts:104-117`) | server process user `agentconsole` | `claude: not found` (#835); even if `claude` were installed for the service user, auth is per-user |

The PTY worker pattern already proves the model works. The other operations were implemented before multi-user mode existed (or without considering it) and never adopted the pattern. As a result, each one surfaced as an independent bug.

## Unifying design

Extract the privilege-elevation invocation from `user-mode.ts:493-500` into a reusable server-side helper:

```
// packages/server/src/services/privilege-elevation.ts (proposed)
export interface RunAsUserOpts {
  username: string;          // null/undefined OR equal to server user → bypass sudo (single-user mode behavior)
  command: string;           // shell command to execute
  cwd?: string;
  env?: Record<string, string>;
  timeoutMs?: number;
}
[...]

PR #858, opened 2026-06-24
feat(server): bootstrap server-side safe.directory at registerRepository

### LINKED ISSUE #853: multi-user mode: bootstrap server-side safe.directory at registerRepository (mirror of #838 for the server user)

ms2sato:
## Summary

Mirror PR #843's `bootstrapSafeDirectoryForUser` pattern for the **server** side. At `registerRepository`, after the existing #845 `applyMultiUserSharedRepoConfig` step, also bootstrap the absolute source-repo path into the SERVER's own gitconfig (e.g. `/home/agentconsole/.gitconfig`) via the same idempotent grep-guarded `git config --global --add safe.directory <path>` invocation -- but with `username: null` so the call runs as the server itself rather than as a requesting user.

Closes #853

## Why

In multi-user mode, an operator may have cloned a source repo as their own OS user (the documented operator flow). After PR #848 / Issue #845 auto-applied `core.sharedRepository=group` + group-writable `.git`, the `chgrp` leg still cannot change file OWNER uid (EPERM unless the server owns the file), so file ownership stays with the cloning user. git's CVE-2022-24765 owner check then refuses any server-initiated git operation against that repo with `fatal: detected dubious ownership in repository at <repoPath>`.

The dogfood observation from Issue #853 captures the cascade -- `git remote get-url origin` (orgRepo detection), `git worktree list` (post-create verification), `git fetch` (branch remote status) all fail, surfaced to the user as the misleading "Worktree was created but could not be found in the list" error.

PR #848 / #845 targets filesystem mode + group ownership; it does NOT change file OWNER uid and therefore does NOT satisfy git's owner check. PR #843 / #838 added `bootstrapSafeDirectoryForUser` for the per-user side at `worktree-service.ts:createWorktree`. This PR is the mirror for the server side at `repository-manager.ts:registerRepository`.

## Implementation

- `repository-manager.ts`: new `SAFE_DIRECTORY_BOOTSTRAP_TIMEOUT_MS = 10000` constant (mirrors `worktree-service.ts`); new `bootstrapSafeDirectoryForServer(repoPath)` private method invoked from `registerRepository` right after `applyMultiUserSharedRepoConfig`. The method early-returns in non-`multi-user` mode. The command shape is the same grep-guarded `git config --global --add safe.directory <path>` idiom used by the user-side bootstrap, only with `username: null` so `runAsUser` skips elevation and writes the SERVER's gitconfig. Failure (timeout or non-zero exit) logs a `warn` and continues; registration is never aborted by a bootstrap failure (the subsequent server-initiated git op will surface a clearer `dubious ownership` error if the bootstrap was actually needed).
- Sibling test file: new `describe('multi-user mode server-side safe.directory bootstrap (Issue #853)')` block with three tests -- (1) multi-user mode invokes a single bootstrap call with the expected grep-guarded command shape and `username: null`, (2) single-user mode is a no-op (`runAsUser.calls.length === 0`), (3) non-zero exit from the bootstrap is logged-and-continued and the DB record is still persisted. The pre-existing #845 idempotent-skip test was updated to acknowledge the new bootstrap call (count 2 -> 3, additional `safe.directory` permitted-branch in the test responder).

## TDD polarity

`git stash` of just the production diff (`packages/server/src/services/repository-manager.ts`) while keeping the new tests in the worktree:

```
RepositoryManager > registerRepository > multi-user mode shared-repo auto-apply (Issue #845) > skips apply when already configured (idempotent no-op via git config probe) [6.00ms]
RepositoryManager > registerRepository > multi-user mode server-side safe.directory bootstrap (Issue #853) > invokes a single server-side safe.directory bootstrap call in multi-user mode [6.00ms]
RepositoryManager > registerRepository > multi-user mode server-side safe.directory bootstrap (Issue #853) > registration succeeds with a warn log when the bootstrap returns non-zero [9.00ms]
Ran 26 tests across 1 file. [361.00ms]  -- 3 fail, 23 pass
```

Three tests fail without the production call (the modified #845 idempotent-skip test + the two new #853 tests that need the bootstrap to fire). The single-user-mode test correctly stays green (it asserts zero calls and the bootstrap is gated on `AUTH_MODE`). Restoring the production file: all 26 tests pass.

## Pre-PR 7-question Gap-Scan

1. **Similar existing mechanism?** Yes -- `worktree-service.ts:bootstrapSafeDirectoryForUser` (PR #843 / #838). This PR is the explicit mirror for the server side. Identical command shape, identical failure-tolerance, identical idempotency guard, identical log shape -- only `username` differs (`null` vs `requestUsername`).
2. **Invocation documented in canonical procedure?** N/A -- this is an automatic step inside `registerRepository`, not an operator-invoked command. Followed the existing #845 `applyMultiUserSharedRepoConfig` pattern that added a step in the same place without a separate procedure doc.
3. **Failure paths tested?** Yes -- non-zero exit (test 3) and single-user no-op (test 2). The `runAsUserMock` already covers spawn-throw at the helper layer.
4. **New file type / directory lifecycle?** No new files / no new artifact types.
5. **Rule clarity?** No rule changes.
6. **Cross-runtime spawn?** No new spawn target. The `runAsUser` helper itself is reused; this PR adds one new caller.
7. **Shared-resource lifetime?** The bootstrap writes one `safe.directory` entry to the server's gitconfig per registered source repo. Lifetime: server-lifetime + (until operator removes it). Out-of-scope from the original Issue: removing entries on `unregisterRepository`. Idempotent re-registration is a no-op via the `grep -Fxq` guard. No embedded paths that could go stale (only the absolute repo path the user just registered, which is the same anchor `registerRepository` already validates).

## Verification log

```
$ bun run typecheck

$ output=$(bun run test 2>&1); exitcode=$?; echo "$output" | tail -100; echo "TEST_EXIT: $exitcode"
@agent-console/shared test:  324 pass / 0 fail / Ran 324 tests across 10 files.
@agent-console/integration test:  29 pass / 0 fail / Ran 29 tests across 8 files.
@agent-console/server test:  2708 pass / 1 skip / 0 fail / Ran 2709 tests across 129 files.
@agent-console/client test:  1397 pass / 2 skip / 0 fail / Ran 1399 tests across 88 files.
scripts:  362 pass / 0 fail / Ran 362 tests across 11 files.
TEST_EXIT: 0

$ bun run check:lang
OK -- language check clean (103 files scanned).

Test Coverage Check: No production files matching coverage patterns were changed.
Rule/Skill Duplication Check: clean
Language Check: clean
```

## Diff scope

```
 .../services/__tests__/repository-manager.test.ts  | 159 ++++++++++++++++++++-
 packages/server/src/services/repository-manager.ts |  81 +++++++++++
 2 files changed, 234 insertions(+), 6 deletions(-)
```

Two files only: production + its sibling test under `packages/server/src/services/`. No files outside `services/`. No documentation files.

## Related

- Umbrella: #837
- Mirror reference (per-user side): #838 / PR #843
- Sibling step in the same registration flow (group-writable + sharedRepository=group): #845 / PR #848
- Companion follow-up surfaced by the same dogfood log: the misleading "could not be found in the list" error-propagation Issue noted in #853's Related section

### DIFF
--- a/packages/server/src/services/__tests__/repository-manager.test.ts
+++ b/packages/server/src/services/__tests__/repository-manager.test.ts
@@ -284,7 +284,10 @@ describe('RepositoryManager', () => {
         // Replace the responder so `getent group` returns FAKE_SHARED_GID
         // (matching the synthesised dir gid) AND the `git config --get
         // core.sharedRepository` probe returns 'group'. The apply call
-        // MUST NOT run -- enforce by throwing if it's invoked.
+        // MUST NOT run -- enforce by throwing if it's invoked. The
+        // server-side safe.directory bootstrap (Issue #853) is a separate,
+        // unrelated runAsUser call and MUST be permitted (it always runs
+        // in multi-user mode regardless of the shared-repo apply state).
         runAsUserMock.responder.fn = async (opts) => {
           if (opts.command.startsWith('getent group ')) {
             return {
@@ -302,6 +305,10 @@ describe('RepositoryManager', () => {
               timedOut: false,
             };
           }
+          if (opts.command.includes('safe.directory')) {
+            // Server-side safe.directory bootstrap (Issue #853); permitted.
+            return { stdout: '', stderr: '', exitCode: 0, timedOut: false };
+          }
           throw new Error(
             `apply should not run when already configured; got command: ${opts.command}`,
           );
@@ -311,13 +318,15 @@ describe('RepositoryManager', () => {
         const repo = await manager.registerRepository(TEST_REPO_DIR);
         expect(repo.path).toBe(TEST_REPO_DIR);
 
-        // Exactly two runAsUser calls -- the `getent` resolve + the
-        // `git config --get core.sharedRepository` probe -- AND no apply
-        // call. The probe-first contract ensures we tightly assert the
-        // idempotent-skip path was actually taken (not bypassed early).
-        expect(runAsUserMock.calls.length).toBe(2);
+        // Three runAsUser calls -- the `getent` resolve + the
+        // `git config --get core.sharedRepository` probe + the server-side
+        // safe.directory bootstrap (Issue #853) -- AND no apply call. The
+        // probe-first contract ensures we tightly assert the idempotent-skip
+        // path was actually taken (not bypassed early).
+        expect(runAsUserMock.calls.length).toBe(3);
         expect(runAsUserMock.calls[0].command).toContain('getent group ');
         expect(runAsUserMock.calls[1].command).toContain('--get core.sharedRepository');
+        expect(runAsUserMock.calls[2].command).toContain('safe.directory');
         const applyCalls = runAsUserMock.calls.filter((c) =>
           c.command.includes('chgrp -R '),
         );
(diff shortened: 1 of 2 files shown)


ms2sato:
@coderabbitai review

PR #861, opened 2026-06-24
feat(client): add clone-from-URL tab to Register Repository dialog (#834)

ms2sato:
## Summary

Closes #834 (frontend slice).

Adds a tabbed "Register Repository" experience with two tabs:

1. **Clone from URL** *(default)* — submits to `POST /api/repositories/clone`, polls `GET /api/repositories/clone/:jobId` for status, dismisses when the Clone Job succeeds. Classified failure codes are translated to user-readable copy; the `unknown` code falls back to the server's raw `error.message`.
2. **Use existing path** — pre-existing behaviour, extracted into a sibling `RegisterExistingPathForm` so the dashboard route's prop contract is unchanged.

Backend (server endpoint + DB-backed Clone Job lifecycle) is implemented in a parallel PR. Both PRs target the API contract specified in Issue #834.

### Notable design decisions

- **`useCloneJobStatus` stops polling on terminal state** by returning `false` from `refetchInterval` once `status === 'succeeded' | 'failed'`. No explicit cancellation needed; tests assert no further polls fire after a terminal response is observed.
- **Validation enforced via shared schema.** `CloneRepositoryRequestSchema` lives in `packages/shared` and is used by the form's `valibotResolver`. The server re-validates the same shape (server remains the source of truth); the pattern explicitly rejects strings starting with `-` (no `--upload-pack=...` injection) plus shell metacharacters and absolute paths.
- **Error-code copy lives in `clone-error-messages.ts`** so the client / dialog stays minimal and the mapping is unit-testable in isolation. The `unknown` code passes the server's raw message verbatim — preserving information instead of replacing it with a generic phrase.
- **Tab default is "Clone from URL"** since the new flow eliminates the manual host-side `git clone` step in multi-user setups (the Issue's primary motivation).

### Feature Overview (true-path)

When the dashboard's "Add Repository" button is clicked, the dialog opens with the "Clone from URL" tab active. The user pastes a git URL (e.g. `git@github.com:org/repo.git`), optionally provides a name, and clicks "Clone & Register". A spinner appears with messages cycling through "Starting clone..." -> "Cloning..." until the server-side Clone Job reaches `succeeded`, at which point the dialog dismisses and the new repository appears in the dashboard's repository list. On failure, the dialog stays open and renders a user-actionable message (and, for non-`unknown` codes, the raw stderr line in a code-styled secondary block for technical context).

[...]

PR #868, opened 2026-06-24
feat(scripts): add dev-multiuser.sh for production-mirrored multi-user dev mode

ms2sato:
## Summary

Adds `scripts/dev-multiuser.sh` (runnable via `bun run dev:multiuser`) — a hot-reloading dev instance that mirrors production multi-user ownership semantics without touching the production data root.

Motivated by the inability to verify multi-user-only behaviour (runAsUser elevation chain, shared-group ownership, setgid inheritance, sharedRepository=group, server-side safe.directory bootstrap) in the existing single-user `bun run dev`. Previously such verification required a full production deploy, which is disruptive when iterating on PRs like #861 (Clone from URL UI) that depend on the multi-user runtime.

## What's in this PR

- `scripts/dev-multiuser.sh` (NEW): orchestration script
- `package.json`: new `dev:multiuser` script entry
- `docs/multi-user-setup-guide.md`: new "Local Multi-User Dev Mode" section

## How it mirrors production

| Element | Production (systemd) | Dev (this script) |
|---|---|---|
| Server runs as | `agentconsole` | `agentconsole` |
| Data root | `/var/lib/agent-console` | `/var/lib/agent-console-dev` |
| Data owner | `agentconsole:agent-console-users` 2775 | identical |
| Developer access path | group membership + setgid + sharedRepository | identical |
| `AUTH_MODE` | `multi-user` | `multi-user` |
| `AUTH_COOKIE_SECURE` | (from bootstrap) | `false` (dev http) |
| `UMask` | `0002` (systemd directive) | `umask 0002` (shell) |
| PATH | `$HOME/.bun/bin:/usr/local/bin:/usr/bin:/bin` | identical |
| PAM auth | required | required |
| runAsUser elevation | active | active |
| Hot reload | no (bundled dist) | yes (bun --watch) |

The developer accesses files under `/var/lib/agent-console-dev/` through their `agent-console-users` group membership + the dir's setgid + sharedRepository=group on each cloned repo -- exactly the access path used in production. They are NOT the owner of any of these dirs/files (which production simulates is a key invariant of the multi-user model).

## Pre-flight checks

The script refuses to run if any of the following are absent (it points at the fix command in the error message):
- `agentconsole` system user
- `agent-console-users` group
- developer is a member of `agent-console-users` (checked via passwd DB, robust against unrefreshed shell)
- `setfacl` (ACL package)
- agentconsole's `bun` binary at one of the production-search-path locations

## Implementation details

- Data root creation: `install -d` (idempotent, ownership + mode set atomically). Subsequent runs only verify ownership and warn (don't fail) if drifted.
- ACL grant: `setfacl -R -m u:agentconsole:rX` + default ACL so newly-created files inherit. One-time per worktree; subsequent runs detect access and skip.
- Process management: vite client and bun server started concurrently in the background; trap on INT/TERM/EXIT cleans up both PIDs (server requires elevation to signal since it runs as agentconsole).

## Verification

- `bun -n scripts/dev-multiuser.sh` (syntax check): pass
- `bun run typecheck`: exit 0
- `bun run test`: exit 0 (server / client / scripts-hooks-skills / integration / shared all pass)
- `bun run check:lang`: 103 files clean
- `preflight-check.js`: rule/skill no-dup, language clean
- Manual local run: pending (requires the multi-user prerequisites set up on the host; ms2sato will verify and report)

## Not in scope

- Bootstrap setup: the dev script assumes `scripts/setup-multiuser-for-ubuntu.sh` has already run for production purposes. If a developer is setting up a fresh machine purely for dev, they still need to run that bootstrap once.
- Parallel co-existence with single-user `bun run dev`: documented as exclusive (same ports). Multi-port support is a separate enhancement if needed.


ms2sato:
@coderabbitai review - addressed 6 of 7 findings (4 Major + 2 Minor). The package.json indent finding declined: see commit message for rationale (the file uses 2-spaces-per-level convention, so the new entry's 4-space indent at depth-2 is consistent with every existing entry; reformatting would break visual consistency).

PR #935, opened 2026-07-02
fix(client): restore terminal copy while a CLI enables mouse tracking

ms2sato:
## Symptom

With Claude Code CLI's new UI (v2.1.195+ mouse click support), text copy via browser drag-selection stopped working in the AgentConsole terminal. Shift+drag did not recover it. Paste kept working. Additionally, the "drag-select, right-click, Copy" flow copied only a single word instead of the selection.

## Root cause (xterm.js 5.5.0, verified against installed sources)

1. **Mouse tracking disables the selection layer.** When the CLI enables DEC mouse tracking (`CSI ? 1000/1002/1003 h`), `InputHandler` sets `CoreMouseService.activeProtocol`, whose setter fires `onProtocolChange`; the browser `Terminal` responds with `_selectionService.disable()` (`browser/Terminal.ts:721-733`), so drag no longer selects — every mouse event is reported to the PTY instead.
2. **macOS has no Shift+drag bypass.** `SelectionService.shouldForceSelection` (`browser/services/SelectionService.ts:429-435`) returns `event.shiftKey` only on non-Mac. On Mac it requires Option+drag **and** `macOptionClickForcesSelection`, which defaults to `false` (`common/services/OptionsService.ts:40`). Our terminal never set it, so macOS users had no bypass at all.
3. **Right-click word selection is a Mac-only default.** `rightClickSelectsWord` defaults to `isMac` (`common/services/OptionsService.ts:48`). A right-click outside the current selection replaces it with a one-word selection (`SelectionService.rightClickSelect`), which is what broke the context-menu copy flow. This behavior predates the CLI change; the CLI change made it the only copy path and thus visible.

## Fix

Three xterm.js public constructor options (no internal APIs), plus a keyboard shortcut:

| Change | Effect |
|---|---|
| `macOptionClickForcesSelection: true` | Option+drag selects a range even while mouse tracking is active (same UX as iTerm2 under tmux mouse mode) |
| `rightClickSelectsWord: false` | Right-click no longer destroys the selection; the native context menu "Copy" copies the full selection |
| `altClickMovesCursor: false` | Prevents a short Option+click from leaking cursor-move escape sequences (arrow keys) into the running CLI, a path newly reachable once Option+drag is enabled |
| Cmd+A in `attachCustomKeyEventHandler` | Selects the whole buffer via `terminal.selectAll()`, which bypasses the disabled selection service; Cmd+C then works natively (xterm's copy handler only checks `hasSelection()`) |

The base options and the key handler were extracted into `lib/terminal-options.ts` / `lib/terminal-key-handler.ts` so the contract is directly importable and unit-testable (the previous fontFamily test had to introspect the component source text; it now imports the builder).

## Trade-offs (deliberate)

- **Ctrl+A is intentionally NOT intercepted.** It is readline's beginning-of-line (`\x01`) and must keep reaching the PTY. Non-Mac users already have the native Shift+drag bypass; a `Ctrl+Shift+A` select-all for non-Mac is a follow-up candidate.
- Option+drag while mouse tracking is OFF now produces a normal selection instead of a column (rectangular) selection (`shouldColumnSelect` is disabled by `macOptionClickForcesSelection` on Mac).
- Option+click no longer moves the cursor via arrow-key sequences (`altClickMovesCursor: false`). This feature only ever worked while mouse tracking was off.
- Right-clicking a word no longer selects it; right-click now always preserves the current selection.
- A caveat that remains: if the CLI re-asserts its mouse mode (the `activeProtocol` setter re-fires `onProtocolChange` even for the same value), the selection is cleared on that redraw. Copy promptly after selecting.

## Second fix (found during owner verification): mouse-protocol re-assertion guard

Owner dogfood on the dev server surfaced a defect the options alone don't cover: **Claude Code's new UI re-asserts its mouse modes (`CSI ? 1000/1002/1003/1006 h`) on redraws** (15+ bursts in one session's output log, zero DECRST). xterm.js's `CoreMouseService.activeProtocol` setter fires `onProtocolChange` even when the value is unchanged, and the browser Terminal responds with `SelectionService.disable()` → `clearSelection()` — so every CLI redraw wiped the user's selection (~100-300ms after mouseup, triggered by the focus report round trip). The burst is `vt200 → drag → any` while already at `any`, so each step individually changes the mode — a naive same-value check cannot help.

`lib/terminal-mouse-protocol-guard.ts` intercepts single-parameter mouse-protocol DECSET at the parser level (`terminal.parser.registerCsiHandler`, public API) and swallows the ones that do not **strictly upgrade** the protocol (the four protocols form a strict event-superset chain `x10 < vt200 < drag < any`). A fresh activation (`none → vt200 → drag → any`) passes through untouched; a steady-state re-assert burst is swallowed entirely, so the selection survives. DECRST, multi-parameter sequences, encodings, and non-mouse params always pass through. Known trade-off (documented in the module): an app that genuinely downgrades (e.g. `any → vt200`) without an intervening DECRST keeps the stronger mode and receives extra motion reports; real-world TUIs toggle off/on instead. Upstream, the cleaner fix would be idempotent protocol-change handling in xterm.js — a follow-up issue candidate.

Also confirmed during verification: Claude Code's new UI renders **its own in-TUI selection** for plain (unmodified) drags — it looks like a browser selection but xterm has no selection at that point, which is why right-click Copy appeared disabled. That path (Claude's own `ctrl+c to copy`) is a separate follow-up investigation (OSC 52 / `@xterm/addon-clipboard`).

## Third fix (found during owner dogfood): stale context-menu copy prime

Owner's continued dogfood surfaced "copy works once, then a reload is needed". Mechanism (reproduced deterministically): xterm's `rightClickHandler` primes the hidden helper textarea for the native context menu (`textarea.value = selectionText; textarea.select()`) but **never un-primes it** — the value is cleared only on blur or Enter/^C keydown. xterm's `copy` listener handles Cmd+C only when `hasSelection()` is true; with no live selection it declines and the **browser default copy delivers the stale, still-fully-selected primed value** — the previous copy's text, silently wrong clipboard data. Under mouse tracking, plain clicks never create an xterm selection, so this stale path is the norm after the first right-click copy.

`lib/terminal-copy-prime-cleanup.ts` clears the primed value as soon as the selection it mirrored is gone, via public APIs only (`onSelectionChange` + `hasSelection()` + `terminal.textarea`). The primed state is identified by its signature — a non-empty value that is fully selected; IME composition text never sits fully selected in the textarea, so composition is unaffected. (The public-API-only constraint mirrors the renderer-research invariant: render/state control through the published surface, no internal hooks.)

## Fourth fix (the decisive one): copy-on-select while mouse tracking is active

Owner verification revealed the deepest layer: `SelectionService` clears the selection on ANY user input (`onUserInput`, SelectionService.ts:139-143), and mouse REPORTS count as user input (`CoreMouseService.triggerMouseEvent` → `triggerDataEvent(report, true)`, CoreMouseService.ts:284). Under the `any` protocol (`?1003h`, which Claude Code uses), **moving the mouse one cell sends a motion report and synchronously clears the selection** — verified with an instrumented stack trace (`triggerMouseEvent → triggerDataEvent → onUserInput → clearSelection`). A "select, then go copy" flow cannot survive; this is not fixable from the embedding side (upstream candidate #4 in issue #943).

So the selection is captured at creation time instead: `lib/terminal-copy-on-select.ts` writes the selection to the clipboard the moment it exists — **only while `mouseTrackingMode !== 'none'`** (normal shells keep the standard select-then-copy behavior, verified in QA). Under tracking, the only possible selections are deliberate Option+drag / Cmd+A gestures, so auto-copy matches intent (iTerm2-style copy-on-select, scoped). Default writer uses the async Clipboard API with an `execCommand('copy')` fallback for insecure origins; failures are logged at debug level.

The full findings, upstream xterm.js contribution candidates, and requirements for the terminal-renderer replacement research are recorded in issue #943.

## Relation to `CLAUDE_CODE_DISABLE_MOUSE_CLICKS`

Claude Code v2.1.195+ ships an env-side escape hatch (`CLAUDE_CODE_DISABLE_MOUSE_CLICKS`). This fix makes AgentConsole copy work **without** that env var, keeps the CLI's mouse features usable, and is robust against any other CLI that enables DEC mouse tracking.

## Tests

- `lib/__tests__/terminal-options.test.ts` — asserts the three new options plus contract pins (cursorBlink/fontSize/fontFamily/theme).
- `lib/__tests__/terminal-key-handler.test.ts` — Cmd+A intercepts (keydown only, lowercase and caps-lock 'A'), Ctrl+A / Cmd+Shift+A / keyup pass through, IME and Shift+Enter behavior unchanged.
- **Polarity verified**: with the fix stashed (tests kept), the 5 fix-targeting tests fail; with the fix restored, 15/15 pass. Contract pins are isolated in a separate describe block and excluded from the polarity set.
- Full suite: client 1478 pass / 0 fail, shared 359, integration 30, typecheck clean. One pre-existing server failure (`MultiUserMode / Issue #918 sshAuthSockFallback`) is an ambient-environment artifact on the dev machine (`SSH_AUTH_SOCK` set by the local agent); this branch does not touch `packages/server` and the test passes in CI's clean environment.

## Browser QA (Chrome DevTools MCP, real reproduction)

Verified on the dev server against a **real Claude Code v2.1.198 session** with its new UI holding mouse tracking active (`.xterm.enable-mouse-events` present), plus a fresh shell with a `printf` DECSET toggle:

- Tracking ON: plain drag selects nothing (symptom reproduced); Option+drag creates a selection that persists while idle; a synthetic `copy` event yields the dragged text; right-click **inside and outside** the selection preserves it and primes the hidden textarea with the full selection (what the native context-menu Copy copies); Cmd+A selects the entire buffer (1019 chars, banner included) with `defaultPrevented`.
- Tracking OFF (fresh shell): plain drag still selects and copies (no regression); right-click outside the selection preserves it.
- **Guard E2E (deterministic)**: with mouse tracking active, an Option+drag selection was made and a delayed re-assert burst (`sleep 3; printf '\x1b[?1000h\x1b[?1002h\x1b[?1003h\x1b[?1006h'`) was delivered through the real PTY → server → WebSocket → parser path while the selection was live. The selection and its copied text survived the burst unchanged (pre-guard, the first `?1000h` cleared it). Also verified against real Claude Code redraws (keystrokes into its input) — selection and copy intact.
- Guard unit/integration tests: 59 tests including a real headless xterm `Terminal` exercising the parser hook (fresh activation reaches `any`; lone downgrade swallowed — the polarity anchor; DECRST and multi-param pass through; dispose restores default behavior).
- **Copy-on-select E2E (host clipboard)**: with a marker preloaded via `pbcopy`, an Option+drag against the live Claude Code session replaced the macOS clipboard with the dragged text (verified via `pbpaste`) — no further action needed. Regression: in a tracking-off shell, plain drag selected text while the clipboard marker stayed untouched (no auto-copy in normal shells).
- Screenshots posted in a PR comment.

Preflight note: the integration-test advisory (`Terminal.tsx — UI component`) does not apply wire-level checks here — this change is client-only, adds no shared type/schema field, and crosses no server/client boundary; coverage is unit tests + the Browser QA above.

## Follow-up candidates (not in this PR)

- Copy-pane button / AgentConsole-native context menu (mobile + discoverability) — to be weighed together with the terminal renderer/UX research track.
- `Ctrl+Shift+A` select-all for non-Mac.

No pre-existing issue tracks this bug; this PR is the tracking artifact.

https://claude.ai/code/session_016ubZz2JXXmXnAtXtJRRp3e

## Note on CodeRabbit

The GitHub-side bot was rate-limited at PR open; after the limit reset it reviewed the full branch and returned **APPROVED with zero inline comments**. The local CodeRabbit CLI was also run against `origin/main` before each push: 1 MINOR finding on the first pass (internal codename in a test doc comment, fixed in `174ed23`), 0 findings on the final pass including the guard commit.

### DIFF
--- a/packages/client/src/components/Terminal.tsx
+++ b/packages/client/src/components/Terminal.tsx
@@ -20,6 +20,7 @@ import { buildBaseTerminalOptions } from '../lib/terminal-options.js';
 import { createCustomKeyEventHandler } from '../lib/terminal-key-handler.js';
 import { installMouseProtocolGuard } from '../lib/terminal-mouse-protocol-guard.js';
 import { installCopyPrimeCleanup } from '../lib/terminal-copy-prime-cleanup.js';
+import { installCopyOnSelect } from '../lib/terminal-copy-on-select.js';
 import {
   register as registerSaveManager,
   unregister as unregisterSaveManager,
@@ -443,6 +444,7 @@ export function Terminal({ sessionId, workerId, onStatusChange, onActivityChange
     let renderStallRecovery: { dispose: () => void } | null = null;
     let mouseProtocolGuard: { dispose: () => void } | null = null;
     let copyPrimeCleanup: { dispose: () => void } | null = null;
+    let copyOnSelect: { dispose: () => void } | null = null;
     let scrollDisposable: { dispose: () => void } | null = null;
     let viewportObserver: MutationObserver | null = null;
     let viewportElement: Element | null = null;
@@ -515,6 +517,14 @@ export function Terminal({ sessionId, workerId, onStatusChange, onActivityChange
       // gone, so a later Cmd+C without a live selection does not re-copy old
       // text. See lib/terminal-copy-prime-cleanup.ts for the mechanism.
       copyPrimeCleanup = installCopyPrimeCleanup(terminal);
+
+      // Copy the selection to the clipboard at creation time while mouse
+      // tracking is active — under tracking mouse reports clear the selection
+      // synchronously, so select-then-copy cannot work. Normal shells are
+      // unaffected. See lib/terminal-copy-on-select.ts for the mechanism.
+      copyOnSelect = installCopyOnSelect(terminal, undefined, (e) =>
+        logger.debug('[Terminal] copy-on-select clipboard write failed:', e),
+      );
 
       // 3. Write the cached snapshot BEFORE open() per the library's
       //    recommendation. updateScrollButtonVisibility is safe even though
@@ -830,6 +840,9 @@ export function Terminal({ sessionId, workerId, onStatusChange, onActivityChange
 
       // Dispose the copy-prime cleanup (removes the selection listener)
       copyPrimeCleanup?.dispose();
+
+      // Dispose copy-on-select (removes the selection listener)
+      copyOnSelect?.dispose();
 
       // Dispose render stall auto-recovery before watchdog
       renderStallRecovery?.dispose();
--- a/packages/client/src/components/Terminal.tsx
+++ b/packages/client/src/components/Terminal.tsx
@@ -18,6 +18,7 @@ import {
 } from '../lib/terminal-restore.js';
 import { buildBaseTerminalOptions } from '../lib/terminal-options.js';
 import { createCustomKeyEventHandler } from '../lib/terminal-key-handler.js';
+import { installMouseProtocolGuard } from '../lib/terminal-mouse-protocol-guard.js';
 import {
   register as registerSaveManager,
   unregister as unregisterSaveManager,
@@ -439,6 +440,7 @@ export function Terminal({ sessionId, workerId, onStatusChange, onActivityChange
     let serializeAddon: SerializeAddon | null = null;
     let watchdog: RenderWatchdog | null = null;
     let renderStallRecovery: { dispose: () => void } | null = null;
+    let mouseProtocolGuard: { dispose: () => void } | null = null;
     let scrollDisposable: { dispose: () => void } | null = null;
     let viewportObserver: MutationObserver | null = null;
     let viewportElement: Element | null = null;
@@ -501,6 +503,11 @@ export function Terminal({ sessionId, workerId, onStatusChange, onActivityChange
       // Enable serialization for terminal state caching
       serializeAddon = new SerializeAddon();
       terminal.loadAddon(serializeAddon);
+
+      // Guard against selection-destroying mouse-protocol re-assertion from
+      // CLIs that re-emit their DEC mouse tracking setup on every redraw.
+      // See lib/terminal-mouse-protocol-guard.ts for the mechanism.
+      mouseProtocolGuard = installMouseProtocolGuard(terminal);
 
       // 3. Write the cached snapshot BEFORE open() per the library's
       //    recommendation. updateScrollButtonVisibility is safe even though
@@ -810,6 +817,9 @@ export function Terminal({ sessionId, workerId, onStatusChange, onActivityChange
 
     return () => {
       cancelled = true;
+
+      // Dispose the mouse-protocol guard (removes the parser handler)
+      mouseProtocolGuard?.dispose();
 
       // Dispose render stall auto-recovery before watchdog
       renderStallRecovery?.dispose();
--- a/packages/client/src/components/Terminal.tsx
+++ b/packages/client/src/components/Terminal.tsx
@@ -19,6 +19,7 @@ import {
 import { buildBaseTerminalOptions } from '../lib/terminal-options.js';
 import { createCustomKeyEventHandler } from '../lib/terminal-key-handler.js';
 import { installMouseProtocolGuard } from '../lib/terminal-mouse-protocol-guard.js';
+import { installCopyPrimeCleanup } from '../lib/terminal-copy-prime-cleanup.js';
 import {
   register as registerSaveManager,
   unregister as unregisterSaveManager,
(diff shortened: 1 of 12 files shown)


ms2sato:
## QA Screenshots (2026-07-02 17:31)

### cmd a select all under mouse tracking
cmd a select all under mouse tracking

<details>

cmd a select all under mouse tracking
</details>

### option drag selection under mouse tracking
option drag selection under mouse tracking

<details>

option drag selection under mouse tracking
</details>

ms2sato:
## QA Screenshots (2026-07-02 18:43)

### selection survives mouse protocol reassert burst
selection survives mouse protocol reassert burst

<details>

selection survives mouse protocol reassert burst
</details>

ms2sato:
@coderabbitai review

Issue #979, opened 2026-07-04
History replay leaks TUI chrome (Claude input box / status bar) into paged rows; replay rows should match the live terminal


ms2sato:
## Symptom (owner dogfood, #959 paging)

After generating a large volume of Claude Code output and paging back through the archive, the paged history contains rows that were never "content": Claude Code's input-box chrome (the `>` prompt frame) and its status bar (`ctx:4% | Sonnet 5 | ms2sato@ca-mini tmp ... /rc`), plus garbled hybrid lines where numbered output fragments are interleaved with story text (e.g. `3818.o「浦島太郎?...`) and stray horizontal-rule fragments.

## Mechanism

The archive stores the raw PTY byte stream, which for a TUI like Claude Code includes the continuous redraw of its bottom chrome (input box + status bar) via cursor positioning and erase sequences. Section 6.2 of `docs/design/terminal-history-paging.md` acknowledges that replaying redraw churn is approximate by design. However, the current implementation makes it substantially worse than it needs to be:

- `replayHistoryChunk` (`packages/client/src/components/terminal/history-replay.ts`) replays into a throwaway terminal with the live terminal's `cols` but a FIXED `REPLAY_ROWS = 24`.
- The TUI authored its escape sequences against the REAL terminal geometry (rows = 39 in the observed session). Absolute cursor addressing beyond row 24 clamps, scroll regions and erase-to-end operations land on the wrong lines, so bottom-chrome frames that would have been overwritten in place on a 39-row screen instead scroll into the throwaway buffer's scrollback and are extracted as "settled" history rows.

## Proposed fix (incremental)

1. **Match replay rows to the live terminal's rows at fetch time** (the same wrap-parity argument the design already makes for `cols`). Chunks are already dropped on a cols resize (Section 6.2); the same drop-on-rows-change rule (or simply keying the replay on the rows in effect when the bytes were produced — not tracked today, so live rows at fetch time is the practical proxy) removes the geometry mismatch, which should eliminate most chrome leakage for sessions whose rows did not change mid-stream.
2. Optionally, post-replay heuristics to drop known-chrome rows are NOT proposed — content-based filtering is fragile; the reader view (#958) is the semantic answer for "what was said".

## Notes

- Distinct from the wrapped-URL issue (#978) and from the retention-floor UX (filed separately).
- Refs #959 (paging design, Section 6.2 "throwaway headless VT replay"), #958 (reader view as the long-term semantic recovery).

[...]

Issue #959, opened 2026-07-02
History paging: archive instead of truncate, backwards range fetch on scroll-to-top


ms2sato:
## Owner proposal (during next-renderer bake, 2026-07-03)

> (translated) Rather than truncating the server file, maybe the server should keep it and serve requested slices, so users can page back through history. The terminal's working buffer stays small, and you can go back when needed.

Context: reload (and idle eviction after 15 min) currently resets the client to the server's retained window (10MB file, truncated to 8MB on overflow). Owner flagged that losing most visible history on reload may hurt users. Measured context: the raw stream is 34% plain text / 66% ANSI control (mostly TUI redraw churn that collapses on replay), so the *visible* loss is smaller than the byte numbers suggest — but the concern stands for line-oriented content.

## Design sketch

- **Server**: replace truncation with segmented archival — cut overflow into compressed segments (the `.log.gz` infrastructure from session hibernation already exists in `worker-output-file.ts`). Disk growth is bounded by gzip (ANSI streams compress extremely well) + optional segment count cap.
- **Protocol**: additive change — a backwards range request ("N bytes ending at offset X") alongside the existing `request-history fromOffset`.
- **Client**: on scroll-to-top, fetch the previous chunk and prepend rows (the next renderer's row model prepends naturally).

## The hard part: ANSI streams cannot be replayed from the middle

VT rendering is order-dependent; an arbitrary middle chunk loses cursor/screen context and TUI redraw regions turn to garbage. Practical approach: replay each fetched chunk through a throwaway headless VT instance and extract **only the rows that landed in scrollback**, discarding the volatile screen region. Clean for line-flow output (shell workers); approximate for TUI churn.

## Alternative worth weighing before building raw paging

Given the S/N measurement, "page back through raw ANSI" mostly pages through redraw churn. Merging this need into #958's reader view (message-unit extraction, archived as structured units server- or client-side) may deliver the actual user value — "go back to what was said" rather than "go back through what was drawn". Decide which layer owns history-paging before implementing.

## Gate

Post-bake (#940). Prioritize against #958. Related design decision: terminal history is an ephemeral view, canonical log is Claude's transcript (ratified in #940 comments) — this issue refines that decision's "how far back the view reaches" parameter rather than reversing it.

ms2sato:
**Sequencing decision (owner + agent, 2026-07-03)** — renderer coexistence analysis:

[...]

PR #997, opened 2026-07-08
fix(client): triple-replay seam correction across all archive chunk boundaries

### LINKED ISSUE #994: fix(client): overlap re-replay across all archive chunk seams (#979 seam class)
## Symptom (owner reproduction 2026-07-07)

Paged scrollback shows words split mid-word: the last one or two characters of a word appear alone at the right edge of a row, and the remainder of the word continues at the start of the next row. The symptom persists across browser F5 reload — the archive on disk carries the corruption. Density scales with Japanese-heavy output, because ~3x byte density triples archive cut frequency.

## Root cause

- Server-side archive segment cuts are **UTF-8 aligned but not newline / wrap-frame aligned** (\`packages/server/src/services/worker-output-file.ts:588-591\`). A single logical line whose bytes span the cut is silently split at an arbitrary byte position.
- Client-side pair-replay (\`packages/client/src/components/terminal/terminal-store.ts:1023-1101\`, \`applyRangeChunk\`) glues only the **top seam** (newest paged chunk against the current visible top). Every **interior seam** (older-vs-older-still) is rendered by an independent throwaway parser via \`applyStandaloneChunk\`.
- Result: chunk N-1's tail runs to the right edge but does not wrap (the parser sees end-of-stream); chunk N's first character prints as the start of a new logical row. The wrap that would have existed in the original stream is lost, producing the "single character alone at the right edge, word continues on the next row" artifact.

This is the SEAM class documented in [Issue #979's forensic update comment](https://github.com/ms2sato/agent-console/issues/979#issuecomment-4880905244) (Owner, 2026-07-04). The rows-parity fix in PR #976 addressed the absolute-positioning subclass; the SEAM class was left for a follow-up because it required client-side buffer retention.

## Chosen approach — Candidate direction C-A from #979

**Client-side overlap re-replay across all seams.**

- Retain each fetched chunk's raw bytes until its older neighbor arrives.
- When the older neighbor arrives, re-replay the newer chunk with the older chunk's tail prefixed, then replace the newer chunk's stored rows with the re-replayed output.
- Generalizes the one-row overlap window that \`docs/design/terminal-history-paging.md\` §6.3 already floats for link detection.

**Trade-off accepted**: ~2x replay work at paging time, plus per-chunk raw byte retention until the older neighbor is fetched (or the chunk is confirmed to be the archive's oldest). The paging read path is not on the interactive-critical path, so the latency impact is acceptable.

**Not chosen — Candidate direction C-B (server-side frame-aligned cuts)**: cleaner long-term, but changes the write path and does not fix existing archives retroactively. C-A both fixes existing archives and is fully client-side; C-B can be revisited if C-A's overhead proves excessive.

## Scope

- Modify \`packages/client/src/components/terminal/terminal-store.ts\` and \`packages/client/src/components/terminal/history-replay.ts\` so every seam runs pair-replay, not just the top seam.
- Add per-chunk raw byte retention with a clear discard rule (when the older neighbor is confirmed present or the chunk is the archive's oldest).
- Preserve the existing top-seam pair-replay behavior — do not regress the case PR #976 addressed.

## Acceptance criteria

- [ ] Reproducing the owner's 2026-07-07 pattern (Japanese-heavy paged content, paging spans >= 3 archive chunks) shows no mid-word wrap-misalignment after the fix
- [ ] TUI chrome leakage on interior seams (the original #979 symptom, \`ctx:... | Sonnet 5 | ...\` status bar) is measurably reduced (count before/after on a synthetic capture)
- [ ] Existing top-seam pair-replay tests continue to pass unchanged
- [ ] New unit tests for interior-seam pair-replay: cover the wrap-boundary case (chunk cut mid-line) and the repaint-frame-boundary case (chunk cut mid-Ink-repaint) explicitly
- [ ] Browser QA: capture before/after screenshots on a synthetic Japanese-heavy log that reliably paged across >= 3 archive chunks

## Out of scope

- Server-side frame-aligned cuts (Candidate direction C-B). Leave the write path unchanged.
- The Fork H1 \`recoverPendingCut\` ambiguous branch (separate diagnosis; filed independently if confirmed).

## Related

- Issue #979 (parent, remains open until SEAM class residual is resolved)
- PR #976 (rows-parity fix, absolute-positioning subclass — landed)
- Design: \`docs/design/terminal-history-paging.md\` §6.3 (one-row overlap window — the C-A implementation generalizes this)

ms2sato:
## Summary

Extends the existing pair re-replay (§6.2, PR #976) to a joint N-chunk replay so every seam between adjacent paged chunks is settled in one throwaway-terminal context. Retains raw bytes on both `pagedChunks[0]` and `pagedChunks[1]` (bounded to 2 raws max, ~512 KB) so the third and subsequent fetches can do a joint 3-chunk replay across TWO interior seams instead of one.

Closes #994
Progresses #979 SEAM class (Candidate direction C-A). C-B server-side option remains as a possible follow-up.

## Root cause correction

The Issue text describes the root cause as "interior seams use standalone parser via `applyStandaloneChunk`". This is inaccurate — trace and code review show the actual bug is subtler.

Every fetch except the first goes through `applyRangeChunk`'s pair path (not the standalone path). The chained pair replays over N steps do NOT leave interior seams uncorrected in the standalone sense; each interior seam DOES get a pair replay. The real bug is that the pair-replay chain **re-includes** the seam row.

Concrete trace (owner symptom scenario, 80 cols, Japanese-heavy dense mid-line cuts):

1. Chunk N-1 stream tail: content ending with partial `hello wor` at columns 60-68 (no trailing newline).
2. Chunk N stream head: `ld world\n...` continuing from N-1's mid-line position.
3. **Step N** (user pages back, fetches N-1 with N as top): `replayHistoryPair(newBytes=N-1, topBytes=N)`. N-1 written first — cursor lands at (row R, col 9) with `hello wor` on row R on-screen. `splitBaseY` captured here excludes row R from N-1's partition. Then N is written — `ld` extends row R to `hello world...`. Partition assigns row R to N's rows (correctly seam-corrected for THIS step).
4. **Step N+1** (fetches N-2): `replayHistoryPair(newBytes=N-2, topBytes=N-1)`. N-1's own last bytes settle on row R' on-screen. This time N-1's partition ends before row R', and row R' (with the un-merged partial `hello wor`) is attributed to N-1's rows partition — **re-including a row conceptually owned by the N/(N-1) seam**.

Result: `pagedChunks[1].rows` ends with `hello wor` (from Step N+1) and `pagedChunks[2].rows` starts with `hello world...` (from Step N, still stored). Visible: `hello wor` on one row + `hello world...` on the next — the reported symptom, density-scaled by Japanese content because multi-byte cuts fall mid-word more often.

## Fix design (Approach C-A)

- **New helper `replayHistoryJoint(chunks[], ...)`** in `history-replay.ts`: generalized N-chunk joint replay, capturing `baseY` between each chunk write and partitioning the extraction at those boundaries. `replayHistoryPair` now delegates to it (N=2 degenerate case).
- **Retention invariant** in `terminal-store.ts`: `pagedChunks[0]` AND `pagedChunks[1]` (if exists) hold `rawData`. Chunks at index >= 2 never do. Bounded memory O(2 x maxBytes) = ~512 KB. Enforced by `enforceRawInvariant()`.
- **Dispatch in `applyRangeChunk`**:
  - Both `[0]` and `[1]` have raw -> `applyTripleChunk` (joint 3-chunk replay, three-way partition).
  - Only `[0]` has raw -> `applyPairChunk` (existing pair path, but under the new rule KEEPS the former top's raw so the NEXT fetch can do a triple).
  - Neither has raw -> `applyStandaloneChunk` (Step 1 of a session or all-blank contiguity break).
- **Fallback chain**: triple overflow -> pair (dropping second-from-top's raw), pair overflow -> standalone, standalone overflow -> re-request at quartered `maxBytes` (existing terminal safety net).

The triple partition attributes the on-screen tail of chunk `k` (which the joint replay merges with chunk `k+1`'s head) to chunk `k+1`'s partition — the same seam-correction rule as pair, applied at both interior boundaries in one pass.

## Test plan

- [x] `replayHistoryJoint`: line-flow triple partition equality, overflow, wrap-boundary seam, repaint-frame-boundary seam (`packages/client/src/components/terminal/__tests__/history-replay.test.ts`, 4 new tests).
- [x] Terminal-store paging: triple retention state (`pagedRawCount === 2` with `secondChunkHasRaw`), wrap-boundary mid-line cut (owner symptom's byte shape) verified duplicate-free, repaint-frame-boundary chained B-row duplication corrected (`packages/client/src/components/terminal/__tests__/terminal-store-paging.test.ts`, 3 new tests + 2 existing pair tests updated to reflect the new retention invariant).
- [x] Polarity verified: stashed the production fix, re-ran the tests, confirmed all 5 modified/new tests fail on the un-fixed code (`pagedRawCount === 1` where the new invariant requires 2, wrap-boundary shows duplicate rows, chained B-rows appear more than once). All 29 unchanged tests continue to pass.
- [x] Full suite green: 1606 client + 3116 server + 408 shared + 46 integration + 420 scripts/hooks = 5596 tests, 0 fail, `TEST_EXIT: 0`.
- [x] Typecheck green.
- [ ] CodeRabbit review (CLI auth blocked in this env — GitHub-side bot review will run on PR).
- [~] Browser QA screenshots — partial. See "Browser QA note" below.

## Notes on preserved behavior

- Existing pair-replay tests for the N=2 degenerate case continue to pass unchanged in row content / attribution — only the retention-count assertions were updated (`pagedRawCount` was 1, is now 2 under the new invariant). Two existing tests renamed / re-explained accordingly.
- The `#959` "never publishes paged row counts ahead of the prepended rows" invariant is preserved.
- Alt-screen paged row suppression (`#959`) is untouched.
- Overflow-fallback `sendRangeRequest` with quartered `maxBytes` is preserved as the terminal safety net.
- Cols-resize / worker-restart / epoch-resync teardown paths clear all paged state including all retained raws (existing behavior).

## Browser QA note (2026-07-09, orchestrator-led session)

The orchestrator ran an isolated dev server on this PR's branch (scratch `AGENT_CONSOLE_HOME`, ports 3557/5273, no impact on the owner's primary instance) and generated 300K rows of Japanese-heavy output via `seq 1 300000 | awk '...'`. Live viewport rendering was fully correct — no word-split symptom on any of the 4 screenshots posted at [PR #997 issuecomment-4920258411](https://github.com/ms2sato/agent-console/pull/997#issuecomment-4920258411). The 300K-row live rendering exercises the Triple-replay's N=2 degenerate case in the live-window path and confirms it does not regress live wrap semantics.

Paged history (interior seam of the archive) could NOT be programmatically triggered from the QA harness. `element.scrollTop = 0` plus repeated `WheelEvent` dispatch (×20) did not fire `terminal-store`'s paged fetch — the archive fetch requires a real user gesture that `Chrome DevTools MCP evaluate_script` cannot synthesize into xterm.js's gesture-scoped scroll path. A REST-level `request-history-range` probe would exercise the server route but skip the production render path this PR modifies, so it was not treated as a substitute.

Joint decision (orchestrator + delegated agent): given that (a) unit tests polarity-verify the interior-seam byte shapes (wrap-boundary mid-line cut + chained B-row duplication under repaint-frame-boundary), (b) the owner's 2026-07-07 symptom describes precisely the wrap-boundary chunk cut mid-line case that the new unit tests cover, and (c) live viewport rendering on 300K rows is clean, the joint judgment is that real-gesture paged-fetch verification is best done by the owner during dogfood after merge. If dogfood shows the symptom persists on paged content, we reopen against this PR's approach.

### DIFF
--- a/packages/client/src/components/terminal/__tests__/history-replay.test.ts
+++ b/packages/client/src/components/terminal/__tests__/history-replay.test.ts
@@ -1,5 +1,10 @@
 import { describe, it, expect, afterEach } from 'bun:test';
-import { replayHistoryChunk, replayHistoryPair, _setReplayScrollbackForTest } from '../history-replay';
+import {
+  replayHistoryChunk,
+  replayHistoryPair,
+  replayHistoryJoint,
+  _setReplayScrollbackForTest,
+} from '../history-replay';
 
 const identity = (d: string) => d;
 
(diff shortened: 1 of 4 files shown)


ms2sato:
## QA Screenshots (2026-07-09 09:02)

### 00 dashboard initial
00 dashboard initial

<details>

00 dashboard initial
</details>

### 01 during output
01 during output

<details>

01 during output
</details>

### 02 output status
02 output status

<details>

02 output status
</details>

### 03 after reload
03 after reload

<details>

03 after reload
</details>

PR #998, opened 2026-07-08
feat(server): use login shell with sentinel for agent worker PTY spawn

ms2sato:
## Summary

Agent worker PTYs now launch through a login shell, so `.bashrc` / `.profile` are loaded correctly.

### Problem

Agent workers were spawned with `sh -c '<command>'` — neither a login shell nor an interactive shell — so `.bashrc` / `.profile` were never read. As a result, agents started with an incomplete PATH that lacked entries like `~/.local/bin` and `~/.opencode/bin`.

### Solution: login-shell sentinel

1. **Login shell launch**: `sh -c 'unset VARS; exec $SHELL -l -c "echo SENTINEL; exec $SHELL"'`
2. **Sentinel detection**: the `onData` handler watches for the sentinel string (all login-shell output before it is skipped)
3. **Command injection**: once the shell is ready, the agent command is sent with `pty.write(command + '\r')`

### What this approach solves

| Problem | Resolution |
|------|------|
| Incomplete PATH | `.bashrc` / `.profile` are loaded |
| Double-escaping of quotes | Command is typed directly via `pty.write()` |
| ActivityDetector false positives | Login-shell output is skipped |
| The dotenv-loading problem of the earlier empty-string approach | The `unset` prefix is preserved |

### Changed files

- `packages/server/src/services/user-mode.ts` — add `sentinel` to `AgentPtySpawnRequest`; support the login-shell form in `spawnDirectPty`
- `packages/server/src/services/worker-manager.ts` — sentinel generation, detection, and command-injection logic
- `packages/server/src/services/worker-types.ts` — add `loginShellSentinel` / `pendingCommand` to `InternalAgentWorker`
- `packages/server/src/__tests__/utils/mock-pty.ts` — sentinel support in the mock PTY
- Test updates: `multi-user-mode.test.ts`, `worker-manager.test.ts`, `mcp-server.test.ts`

## Testing

- See the Verification subsection under "Completion review" below for the final full-suite results, and the Real-machine E2E section for both-environment verification and the post-deploy smoke script.

---

### Found & fixed during completion review: elevated path dropped all output

The initial implementation set the sentinel gate unconditionally in `worker-manager.ts`, but `MultiUserMode.spawnSudoPty` (the elevated spawn path) ignored `request.sentinel` and ran the agent command directly. No sentinel was ever emitted on that path, so the gate dropped **all** agent worker output forever in multi-user elevated sessions (worker appeared dead). The dogfood host runs multi-user, so this would have hit immediately after merge.

Fix (commit `baaef12`): the elevated path now mirrors the direct path — when `request.sentinel` is present, the inner command becomes `echo <sentinel>; exec $SHELL` and the worker-manager injects the real command after sentinel detection. `buildElevationArgs` is untouched (strict-thin-wrapper contract). Elevated-path spawn-shape tests added with polarity confirmed (the new test fails with the fix reverted).

Refs #999 (structural follow-up: enforce the UserMode agent-PTY spawn contract structurally).

### CI fixes

- **comment-blame-shift-lint / KNOWN_VIOLATIONS test**: the PR added 5 lines above an allowlisted `Issue #769` comment in `worker-manager.ts`, shifting it 440 → 445. Allowlist line number updated.
- **preflight (missing sibling test)**: added direct-path sentinel branch tests to `__tests__/user-mode.test.ts` covering the `spawnDirectPty` change.

### Real-machine E2E verification

**Single-user (host, real `.bashrc`/login init)** — server started on a spare port with a temp `AGENT_CONSOLE_HOME`, quick session + agent worker created through the shipping REST path:

- Sentinel string never appears in user-visible output (`grep -c __AGENT_CONSOLE_READY_` = 0 on the worker output stream).
- Login-shell PATH confirmed: worker env shows `/home/<user>/.local/bin`, `~/.bun/bin`, `~/.opencode/bin` — the exact paths this PR exists to restore.
- Built-in Claude Code agent boots to its interactive UI (the `claude` binary resolves via the login-shell PATH).
- ActivityDetector stays `idle` during login-shell init (pre-sentinel output is gated; no false detection).

**Multi-user elevated (Docker dev stack, `alice` ≠ service user, genuinely exercises the elevated spawn)**:

- Agent worker output flows (this is the path that dropped everything before the fix).
- Wire-level check via the client's own `/ws/session/:id/worker/:id` protocol (`request-history`): no sentinel in any frame; injected command, `whoami` = `alice`, and PATH output all present.
- Process table shows the elevated chain running the Option-A inner command (`...; echo __AGENT_CONSOLE_READY_<id>; exec $SHELL`) with the exec'd shell owned by `alice`.

### Observations (non-blocking, likely pre-existing)

- Cosmetic: on the host run, the injected command line can appear once as raw tty echo before the first prompt draw (timing-dependent; did not reproduce in the container run).
- Docker dev stack only: the vite proxy hangs WebSocket upgrades (`/ws/*` via port 15173) while direct server-port WebSockets work; REST via the proxy is fine. This PR touches neither client nor WS infra, but it was only observed on this branch, so noting for transparency.

### Verification

- `bun run typecheck` — exit 0
- `bun run test` (full suite) — exit 0: server 3125 pass / 1 skip / 0 fail; scripts 420 pass (run with `SSH_AUTH_SOCK` unset; the known non-hermetic `#918` test depends on the runner's ambient SSH agent socket)
- `node .claude/skills/orchestrator/preflight-check.js` — exit 0

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -72,6 +72,15 @@ A worker that shows the git diff between a base commit and a target ref for its 
 - **Aliases:** DiffWorker, git-diff worker
 - **See:** [GitDiffWorker in worker.ts](../packages/shared/src/types/worker.ts)
 
+### Login-Shell Sentinel
+The spawn → gate → inject protocol used to launch an [AgentWorker](#agentworker) PTY through the user's login shell. Instead of running the agent command directly (which skips `.bashrc` / `.profile` and yields an incomplete PATH), the PTY spawns a login shell whose inner command echoes a per-activation random marker (`__AGENT_CONSOLE_READY_<id>`) and then execs an interactive shell. The worker-manager's onData handler **gates** on the marker: all output before it (login-shell init noise) is dropped and never reaches the output buffer, the persisted output file, clients, or the ActivityDetector; detection is chunk-boundary-safe via a bounded carry of the pre-sentinel tail. On detection, the pending agent command is **injected** with `pty.write(command + '\r')`, avoiding shell-quoting double-escapes.
+
+Both spawn routes honor the same protocol: the direct path (`spawnDirectPty`, single-user mode and the multi-user elevation-skip case) wraps the marker in `$SHELL -l -c`, and the elevated path (`MultiUserMode.spawnSudoPty`) emits it from the inner command of the elevation argv (the elevated login shell provides the login init). A route that spawned without emitting the marker would leave the gate closed forever and silently drop all agent worker output.
+
+- **Key code:** `AgentPtySpawnRequest.sentinel` ([user-mode.ts](../packages/server/src/services/user-mode.ts)); `InternalAgentWorker.loginShellSentinel` / `pendingCommand` ([worker-types.ts](../packages/server/src/services/worker-types.ts)); gate + inject in `WorkerManager.setupWorkerEventHandlers` ([worker-manager.ts](../packages/server/src/services/worker-manager.ts))
+- **Aliases:** sentinel protocol, login-shell sentinel protocol, readiness sentinel
+- **See:** Issue [#999](https://github.com/ms2sato/agent-console/issues/999) (structural enforcement follow-up: sentinel protocol single-writer + spawn-contract conformance suite)
+
 ### Base Spec
 The persisted comparison base of a [GitDiffWorker](#gitdiffworker), stored in `workers.base_commit`. Unlike a frozen commit hash, a base spec records the user's *intent* and is **re-resolved on every diff computation**, so the diff stays aligned with GitHub's merge-base (three-dot) view as the branch absorbs upstream commits. Forms:
 - `merge-base:<ref>` — fork point via `git merge-base <ref> HEAD` (e.g. `merge-base:origin/main`); re-resolves each diff.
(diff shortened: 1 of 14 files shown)


ms2sato:
## Browser QA 確認結果

開発サーバーを起動し、Chrome DevTools MCP でブラウザ経由で動作確認しました。

### テスト手順
1. `bun run dev` で開発サーバー起動
2. REST API でセッション作成: `POST /api/sessions` (`initialPrompt: "echo hello"`)
3. ブラウザでセッションページを開いて Agent Worker の動作を確認

### 確認結果

| 項目 | 結果 |
|------|------|
| ログインシェル起動 | ✅ プロンプト `ms2sato@ubuntu:/tmp$` が表示 → `.bashrc`/`.profile` 読込確認 |
| センチネル検知 → コマンド書き込み | ✅ `claude 'echo hello'` がプロンプト後に表示 |
| Claude Code 起動 | ✅ `Claude Code v2.1.204` が正常起動、`echo hello` を実行 |
| ActivityDetector 誤検知なし | ✅ ステータス "Idle" → ログインシェル出力で誤判定なし |
| ログインシェルノイズ除外 | ✅ MOTD等がターミナルに表示されない → センチネル前出力スキップ確認 |

### スナップショット抜粋

ターミナル出力:
```
ms2sato@ubuntu:/tmp$ claude 'echo hello'
  Claude Code v2.1.204
  ...
  ❯ echo hello
  ● Bash (echo hello)
    ⎿  hello
  ● hello
  echo hello を実行し、hello と出力されました。
  ✻ Worked for 9s
```

ShellWorker で `claude` を手動起動したのと同等の状態が実現されています。

ms2sato:
@coderabbitai review

Issue #1004, opened 2026-07-11
feat: embedded agent worker (embedded-agent) v1 — implementation umbrella


ms2sato:
## Summary

> **Naming note:** the worker type was renamed from the working name `loop-agent` to **`embedded-agent`** (final, owner decision 2026-07-11) before any implementation started. "Embedded" follows the embedded-database convention (SQLite vs DB server). If any artifact still says `loop-agent`, it is stale — `embedded-agent` wins.

Implementation umbrella for the **Embedded Agent Worker** (worker type `embedded-agent`): a worker whose LLM loop is owned by Agent Console itself, running as a per-user subprocess, streaming NDJSON structured events, executing app-operation tools via the existing MCP server, targeting OpenAI-compatible APIs and local LLMs.

The design was accepted in PR #1001. This issue is the execution handle: it tells the Orchestrator what to read, in what order to execute, and what each phase must satisfy.

## Normative spec (read first)

1. **`docs/design/embedded-agent-worker.md` — Part II "v1 Design Specification"** is the spec the implementation must match (per the "Design Documents as Specification" rule). It contains: shared types + wire schemas, DB migration + a file:line table of every mapper/branch site, the NDJSON stdio protocol, `EmbeddedAgentWorkerService` behavior, the capability-predicate refactor, Issue #878 phase 1 design, provider adapter scope, credentials handling, UI spec, error/edge-case table, testing plan, and the phase plan this issue mirrors.
2. Part I of the same doc — rationale and decisions (candidate (b) chosen; restart-resume deferred post-v1; #878 in scope). Read to avoid re-litigating settled decisions.
3. `docs/glossary.md` — entries: Embedded Agent Worker, EmbeddedAgentDefinition, MCP Caller Token.
4. `.claude/rules/elevation-helpers.md` (spawnAsUser consumer obligations) and `.claude/rules/os-environment-coupling.md` (smoke-test discipline; applies to Phase 4).
5. Issue #878 (MCP caller identity) — Phase 0b below implements its phase 1; #878 stays open until Phase 4 lands terminal-agent token delivery + multi-user `enforce`; single-user `warn` retirement is post-v1.

## Execution order

Phases are dependency-ordered; 0a and 0b are independent of each other and can run in parallel. Each phase = one child issue + one PR (PR body `Closes #<child>`; child issues reference back to this umbrella). The Orchestrator creates each child issue at delegation time, copying the phase's acceptance criteria from the spec's "Implementation plan (phases)" table.

[...]

PR #1048, opened 2026-07-12
feat: reuse MessagePanel in EmbeddedAgentWorkerView (closes #1040)

### LINKED ISSUE #1040: embedded-agent: reuse MessagePanel in EmbeddedAgentWorkerView (prop-parametrize PTY-specific coupling)
Owner-decision follow-up surfaced during dogfood of umbrella #1004 (v1 embedded-agent, PRs #1008–#1032). Architect-designed (session `84a3a530-86a8-4825-b103-bc3edfe764ee`) MessagePanel reuse pattern.

## Context

`EmbeddedAgentWorkerView` currently ships with a standalone `<textarea>` + Enter-to-send handler as part of v1. Two issues emerged during dogfood:

1. **IME send bug**: the standalone textarea does not check `e.nativeEvent.isComposing`, so pressing Enter to confirm an IME conversion (Japanese/Chinese/Korean) sends a partial message.
2. **Feature parity gap**: `MessagePanel` (used by PTY workers) provides fixed-string templates and DnD file attachment, both of which are useful to embedded-agent chat too.

Owner observation (2026-07-12):
> MessagePanel ですが、否定されるほどリッチですか?固定文字列を入れられるテンプレートと、DnD でのファイル添付ですよね。どちらにせよ必要になると思うので、再開発する必要性を感じません

= Do not maintain a parallel textarea; reuse `MessagePanel` directly.

## Design (architect-decided)

**Reuse `MessagePanel` in `EmbeddedAgentWorkerView` by parametrizing PTY-specific coupling as 4 props.** No new common component is extracted (per architect: coupling points are bounded and finite; extract later only if a 3rd consumer appears).

### Coupling points → props

1. **Send transport** (`onSend(content: string, files?: File[]) => Promise<void>`)
   - PTY: existing REST call to `sendWorkerMessage` (PTY-inject)
   - Embedded: WebSocket user-message send
2. **Escape handler** (`onEscape?: () => void`)
   - PTY: current `getOrCreateTerminal(...).sendInput('\x1b')` behavior
   - Embedded: `undefined` (no terminal instance to reach)
3. **Slash-command completion** (feature flag prop, default on)
   - PTY: on (existing `fetchSkills` behavior)
   - Embedded: **off** (no skills in embedded v1)
4. **File attachment** (feature flag prop, default on)
   - PTY: on (existing `/tmp` save + path-injection semantics)
   - Embedded: **off until FF-1 (Read tool) lands**, then flip to on with the same path-injection semantics as PTY. Cross-reference: FF-1a (Read tool) — this issue's attachment-enable step depends on it.

### Send-key policy: **unchanged (`Ctrl/Cmd+Enter`)**

MessagePanel's current `Ctrl/Cmd+Enter` send stays as-is for both PTY and embedded. No `sendKey` prop needed. Rationale (architect):

- In-app consistency > cross-app convention: users already trained on Ctrl+Enter in PTY MessagePanel; changing send key per view raises cognitive load, especially since PTY and embedded views are used side-by-side.
- IME environments make `Enter=send` structurally fragile — `isComposing` guards most cases but the psychological proximity of "conversion-commit Enter" and "send Enter" is itself a stress source.
- Reversible: `MessagePanel` internally separates `handleKeyDown` and `handleSend`; adding an `Enter=send` mode later requires 1 additional prop.

### Required MessagePanel change (independent of the reuse)

Add `if (e.nativeEvent.isComposing) return;` at the top of `handleKeyDown`. This is required regardless of the reuse work because:

- MessagePanel's slash-completion dropdown consumes bare `Enter` / `Tab` / `ArrowUp` / `ArrowDown` for command selection.
- During IME conversion, these keys drive candidate navigation and commit — without the guard, IME conversion misfires command selection.
- The guard is a net positive for the PTY MessagePanel too.

## Files

- `packages/client/src/components/MessagePanel.tsx` — add 4 props, add `isComposing` guard in `handleKeyDown`
- `packages/client/src/components/workers/EmbeddedAgentWorkerView.tsx` — remove standalone textarea, mount `<MessagePanel>` with embedded-specific props
- Existing tests for MessagePanel: extend for prop matrix (send transport / escape / slash flag / attachment flag)
- New tests for EmbeddedAgentWorkerView using MessagePanel through the WS send path

## Acceptance criteria

- `EmbeddedAgentWorkerView` no longer contains a standalone textarea; it uses `<MessagePanel>` with embedded props
- Send key = `Ctrl/Cmd+Enter` (same as PTY)
- IME conversion Enter does NOT send in either PTY or embedded (regression polarity test: with fix removed, IME Enter incorrectly sends → confirms the fix is load-bearing)
- Slash completion is off in embedded MessagePanel
- File attachment is off in embedded MessagePanel (disabled UI with tooltip: "Enable after Read tool ships — see #1004 FF-1a")
- Fixed-string templates work identically in embedded and PTY MessagePanel
- Unit tests cover the 4 props' effect on rendered UI + event handlers
- Integration test: sending a message via embedded MessagePanel reaches the WS handler as a user-message event
- Browser QA: (a) confirm IME conversion no longer sends prematurely (Japanese/Chinese/Korean IME), (b) confirm template + Ctrl+Enter send in embedded view, (c) screenshot of attachment disabled state

## Timing

- **Independent of tool-surface fast-follows**. The scaffold (textarea removal + templates + `Ctrl+Enter` + `isComposing`) can ship in parallel with FF-1.
- Attachment enable happens after FF-1a (Read tool) lands, in a follow-up PR (this issue's scope is the disabled state).

## Reversibility

If future dogfood surfaces demand for `Enter=send` in embedded specifically, add a single `sendKey?: 'enter' | 'mod-enter'` prop. `handleKeyDown` and `handleSend` are already separated internally, so the change is ~5 LOC.

## Change surface

Moderate — 2 files + prop parametrization. Low risk (no new WS message shape, existing `sendUserMessage` reused).

Cross-reference: umbrella #1004, spec `docs/design/embedded-agent-worker.md` Part II, FF-1a (Read tool, blocks attachment enable).

ms2sato:
## Summary

`EmbeddedAgentWorkerView` shipped in v1 with a standalone `<textarea>` + Enter-to-send handler, duplicating `MessagePanel` (used by PTY workers). Dogfood surfaced two issues: (1) the standalone textarea didn't check `e.nativeEvent.isComposing`, so confirming an IME (Japanese/Chinese/Korean) conversion with Enter sent a partial message; (2) it missed `MessagePanel`'s templates and DnD file attachment for no good reason.

This PR reuses `MessagePanel` directly in `EmbeddedAgentWorkerView` by parametrizing its PTY-specific coupling as 5 props (architect-approved design, session `84a3a530`, decision revised mid-implementation per owner dogfood feedback — see design notes below):

- `onSend(content, files?) => Promise<void>` — send transport (PTY: REST `sendWorkerMessage`; embedded: WS `sendUserMessage`)
- `onEscape?: () => void` — Escape handler (PTY: routes to the terminal instance; embedded: omitted, no terminal to reach)
- `slashCompletionEnabled?: boolean` (default `true`) — off for embedded (no skills concept there yet)
- `attachmentsEnabled?: boolean` (default `true`) — off for embedded, disabled Attach button with tooltip `"Enable after Read tool ships — see #1004 FF-1a"` (depends on FF-1a landing first)
- `sendDisabled?: boolean` (default `false`) — **send gate, not a textarea disable**: while a turn is active, the textarea stays fully editable (draft accumulation, chat-app convention) but the Send button disables and Ctrl/Cmd+Enter becomes a no-op. Busy state is still shown via `EmbeddedAgentWorkerView`'s own existing activity-dot/Cancel status line, unchanged.

Also added a `e.nativeEvent.isComposing` guard at the top of `MessagePanel`'s `handleKeyDown` — required regardless of the reuse, since the slash-completion dropdown consumes bare `Enter`/`Tab`/arrows that IME composition also drives. Benefits both PTY and embedded.

`SessionPage.tsx`'s existing `sendWorkerMessage`/ESC-via-terminal-store wiring was extracted into a small `messagePanelHandlers.ts` sibling module (following the existing `workerRestart.ts` / `tabAppearance.ts` / `tabKeyboardNavigation.ts` convention) so it's independently unit-testable without rendering the full `SessionPage`.

## Design note: `sendDisabled` vs plain disable

The original plan only had 4 props. Mid-implementation, an existing test asserted the old v1 behavior of disabling the *entire textarea* while a turn is active. The architect revised the design to a 5th prop (`sendDisabled`) after owner feedback that a full textarea disable during turns doesn't match standard chat-app UX (Slack/Discord/ChatGPT all allow drafting while a response streams) and was a v1 simplification rather than a deliberate choice. The server's `TURN_IN_PROGRESS` rejection remains as a race-condition backstop (state-transition gap between client-observed idle and actual server state); it should no longer surface in the normal path now that the client gates sending itself.

## Test plan

- [x] `bun run typecheck` — exit 0, all 4 packages
- [x] `bun run test` — full paste, client 1791/1791 pass. One pre-existing, unrelated failure in `packages/server/src/services/__tests__/multi-user-mode.test.ts` caused by this local dev machine's real 1Password `SSH_AUTH_SOCK` env var leaking into the test process (confirmed via `git status` — that file is untouched by this diff).
- [x] IME regression polarity test (workflow.md TDD requirement): with `isComposing` guard temporarily removed, the new test **fails** (sends during composition); restored, it **passes**. Confirms the guard is load-bearing.
- [x] Unit test matrix for all 5 `MessagePanel` props (send transport, escape, slash flag, attachment flag + tooltip, send-gate) in `MessagePanel.test.tsx`
- [x] `EmbeddedAgentWorkerView.test.tsx` updated: Ctrl+Enter send-path integration test (asserts the actual `embedded-user-message` WS payload), turn-active send-gate test (textarea stays editable, Send disabled, Cancel still shown), re-enable-on-idle test
- [x] New `messagePanelHandlers.test.ts` covering the extracted PTY wiring (fetch-mock assertion for `sendPtyWorkerMessage`, terminal-store spy for `escapePtyWorker`)
- [x] Browser QA (Chrome DevTools MCP, dedicated worktree dev instance on ports 3458/5174 to avoid colliding with another running dev server):
  - (a) Simulated Japanese IME composition (`compositionstart` → Ctrl+Enter while `isComposing:true` → no send, content preserved → `compositionend` → Ctrl+Enter while `isComposing:false` → sent successfully through the real WS pipeline, confirmed by the message appearing in the chat transcript)
  - (b) Attachment disabled state in the embedded worker: Attach button rendered disabled with the exact tooltip text, screenshotted (true-path, feature-visible state, not just the PTY default-enabled state)
  - (c) Template creation + insertion into the embedded MessagePanel confirmed (`Manage Templates` → add → select from dropdown → content populated in the textarea, Send re-enabled)
- [x] `node .claude/skills/orchestrator/preflight-check.js` — clean except one flagged item, documented below.

## Known preflight-tool limitation (transparency note, not a real coverage gap)

`preflight-check.js`'s `COVERAGE_PATTERNS` for `packages/client` only recognizes `.tsx` files (or `.ts` under `hooks/`); a plain `.ts` sibling module under `components/sessions/` (like the new `messagePanelHandlers.ts`, or the pre-existing `workerRestart.ts`/`tabAppearance.ts`/`tabKeyboardNavigation.ts`) is invisible to the pattern. Since this PR's `SessionPage.tsx` diff is now pure delegation to `messagePanelHandlers.ts` (no independently-testable logic left in `SessionPage.tsx` itself), and that module has full, non-tautological sibling-test coverage, the mechanical checker still flags `SessionPage.tsx` as "missing tests" even though the actual logic is tested. Did not add a tautological test to `SessionPage.test.ts` just to silence the tool (`check-utils.js`'s own docstring calls that out as a known anti-pattern from PR #694). Flagging for the Orchestrator's awareness; may be worth a follow-up to extend `COVERAGE_PATTERNS` to recognize sibling `.ts` modules under `components/**`.

Closes #1040

## Summary by CodeRabbit

* **New Features**
  * Added externally controlled MessagePanel behavior (custom `onSend`, optional `onEscape`, and feature flags for send/attachments/slash completion).
  * Added PTY worker message helpers for forwarding sends and routing Escape to the selected worker.
  * Updated embedded worker UI to use MessagePanel with slash completion and attachments disabled, and send gated during active turns.

* **Bug Fixes**
  * Send/keyboard shortcuts now respect `sendDisabled`, with IME-safe sending and improved ESC/attachment gating.

* **Tests**
  * Refreshed and expanded MessagePanel, handler, EmbeddedAgentWorkerView, and SessionPage coverage (including new prop-contract checks).


ms2sato:
@coderabbitai review

PR #1050, opened 2026-07-12
feat: embedded-agent FF-1a — built-in tools (Read/Glob/Grep) + enabledTools policy

### LINKED ISSUE #1042: embedded-agent: FF-1a — built-in tools (Read/Glob/Grep) + enabledTools policy + path confinement
Fast-follow for umbrella #1004 (Embedded Agent Worker v1). Architect-designed (session `84a3a530-86a8-4825-b103-bc3edfe764ee`) subprocess-local built-in tools — first of three tool-surface issues (FF-1a Read/Glob/Grep, FF-1b Bash, FF-1c Write/Edit) plus a separate FF-2 for sandbox.

## Context

v1 Embedded Agent shipped without Bash / Read / Write / Edit / Glob / Grep. This is the largest post-v1 gap surfaced in dogfood; without a tool surface the agent cannot inspect the workspace.

Owner design direction (2026-07-12):
> MCP に載せず、Claude Code / opencode 型の subprocess 内部 tool として持つ。参考 CLI から subset で始める、車輪再発明禁止

= Do not put tools on the MCP surface. Implement them as subprocess-internal tools like Claude Code / opencode. Start with a subset from reference CLIs; do not reinvent the wheel.

## Design (architect-decided, from Q1 + Q2 rounds)

**Subprocess-local execution.** Tool definition and dispatch live inside the embedded-agent subprocess (`packages/embedded-agent/`). The subprocess already runs as the requested user via `spawnAsUser`, so cwd (`locationPath`), filesystem permissions, and multi-user elevation are **automatically correct** — no new elevation surface. This is the key architectural property: the same argument that made MCP tool authorization correct at the AGENTS.md-loading site applies here.

**Provider tools = MCP tools ∪ builtin tools.** At startup, merge the two lists. Name collisions are detected and resolved builtin-first with a warn log. Existing turn cycle: executor branches at the start — if the tool name is a builtin, execute locally; otherwise route through the existing `McpToolClient`. **Results ride the existing tool-result event (with the existing 16 KiB truncate machinery), so no WS protocol / UI / server changes.**

**Tool names and argument shapes are identical to Claude Code** (`Read`, `Glob`, `Grep`, and later `Bash`, `Write`, `Edit`). Rationale: pretrained models already know these shapes; owner direction is "start from reference CLI subsets, no wheel reinvention."

## Scope of THIS issue (FF-1a)

- The **built-in registry** and executor dispatch skeleton (foundation for all FF-1x)
- Three read-only tools: **`Read`, `Glob`, `Grep`**
- The **`enabledTools` policy schema** (schema + migration + mapper + UI + validation)
- **Path confinement** for these tools (realpath resolve + `locationPath` prefix enforcement, symlink-escape rejected)
- **Secret non-leakage negative assertion** at the unit-test layer (provider apiKey / MCP token not present in the loop-visible builtin execution context)

Explicit **not in scope**:
- `Bash` implementation → FF-1b (below)
- `Write` / `Edit` implementation → FF-1c
- OS-level sandbox (bubblewrap etc.), system-wide policy floor, MCP surface per-caller filtering → **FF-2** (separate issue). The path confinement here is the "minimum floor" that FF-2 will build on.
- Interactive approval UI (`asking` state) → post-v1

## Data model — `enabledTools` shape

Single-writer type in `packages/shared/src/types/embedded-agent.ts`:

```ts
export const EMBEDDED_AGENT_TOOL_NAMES = ['Read', 'Glob', 'Grep', 'Bash'] as const;
export type EmbeddedAgentToolName = (typeof EMBEDDED_AGENT_TOOL_NAMES)[number];

export interface EmbeddedAgentDefinition {
  // ...existing fields
  enabledTools?: EmbeddedAgentToolName[];
}
```

**Semantics:**
- `undefined` (absent) → default = `['Read', 'Glob', 'Grep']` (read-only ON, Bash OFF)
- `[]` → explicitly all built-in tools OFF
- Explicit array → exact enabled set. Tools not listed are **not shown to the provider** (unrepresentable at the tool-call layer, not merely rejected). This is the invariant behind "opt-in for mutating tools."

**Note:** `Bash` is defined in this shared enum here (FF-1a) so schema + migration + UI land atomically. The Bash tool **implementation** ships in FF-1b; until then, selecting `Bash` in the UI has no effect (registry has no entry). This is intentional to avoid a two-round migration.

## Database

```sql
-- migration vNN (next unused number)
ALTER TABLE embedded_agents ADD COLUMN enabled_tools TEXT;
```

- Nullable `TEXT` column, values are JSON array strings
- Examples: `NULL` (default), `'[]'` (all off), `'["Read","Glob","Grep"]'` (explicit read-only)
- Mapper: `NULL ↔ undefined`, `TEXT ↔ JSON.parse` / `JSON.stringify`
- PATCH semantics (following existing convention): `enabledTools: null` in the PATCH body resets to NULL (= default); `enabledTools: undefined` (key absent) = no change; `enabledTools: [...]` = replace

## Validation

valibot in `packages/shared/src/schemas/`:

```ts
enabledTools: v.optional(
  v.pipe(
    v.array(v.picklist(EMBEDDED_AGENT_TOOL_NAMES)),
    v.check((arr) => new Set(arr).size === arr.length, 'duplicate tool name'),
  ),
),
```

Invalid tool names are unrepresentable at the boundary (per `.claude/rules/design-principles.md` "structure not convention").

## UI (Add/Edit Embedded Agent form)

Add a "Tools" section with two groups of checkboxes:

- **Read-only** — `Read`, `Glob`, `Grep`. Checked by default when creating a new agent.
- **Command execution** — `Bash`. Unchecked by default, with an amber warning label: `Runs arbitrary shell commands as the session user.`

The form always sends an **explicit array** (never `undefined`) to eliminate the "did the user mean default or explicit?" ambiguity. `undefined` remains legal in the type only for backward compatibility with pre-FF-1a rows.

### Read
- Args: `{ path: string, limit?: number, offset?: number }`
- Behavior: matches Claude Code's Read — line-numbered output, respects `limit` / `offset`
- Path confinement: `realpath(path)` must start with `realpath(locationPath) + '/'` (or equal `locationPath`)

### Glob
- Args: `{ pattern: string, path?: string }`
- Behavior: matches Claude Code's Glob — glob within `path` (default `locationPath`)
- Path confinement: `path` (if provided) same rule as above; results outside `locationPath` are filtered out

### Grep
- Args: `{ pattern: string, path?: string, ... }`
- Behavior: matches Claude Code's Grep — content search
- Path confinement: same as Glob

## Path confinement (this issue's "minimum floor")

- All three tools resolve `realpath(path)` before use
- Reject if the resolved path is not within `realpath(locationPath)` (compare with a trailing `/` guard, or equality)
- Reject symlinks whose targets escape `locationPath`
- Rejection: return a `tool-result` with `ok: false` and a message like `"Access outside session location is not permitted."` — do not throw a turn-level error

## Secret non-leakage (unit-level negative assertion)

FF-1a's negative assertion: **provider apiKey and MCP token are not readable from within the builtin-tool execution context.** Because tools execute in-process, this is a property of the builtin registry's dispatch (not process env). Assert:

- The tool's execution scope does not contain the loop's apiKey / token bindings (or, if it does structurally, assert the strings do not appear as tool-result content in any test scenario)

Concrete env-based Bash-child assertions (child env does not contain `AGENT_CONSOLE_*` / apiKey / token) ship in FF-1b, where a child process actually exists.

## Files

- `packages/embedded-agent/src/tools/` — new directory
  - `index.ts` — builtin registry + dispatch entry point
  - `read.ts`, `glob.ts`, `grep.ts` — tool implementations
  - `path-confinement.ts` — shared path validation helper
- `packages/embedded-agent/src/` — existing turn-cycle executor updated with dispatch branch
- `packages/shared/src/types/embedded-agent.ts` — `EMBEDDED_AGENT_TOOL_NAMES` + type extension
- `packages/shared/src/schemas/embedded-agent.ts` (or equivalent) — valibot
- `packages/server/src/db/migrations/vNN-add-embedded-agent-enabled-tools.ts` — migration
- `packages/server/src/repositories/embedded-agent-repository.ts` — mapper NULL ↔ undefined
- `packages/client/src/components/embedded-agents/EmbeddedAgentForm.tsx` — Tools section UI
- Tests: unit for registry, unit for each tool + path confinement, integration for the WS turn cycle with a builtin tool call, wire-level test per Q10

## Spec update (same PR)

- `docs/design/embedded-agent-worker.md` Part II — add `## Built-in tools (fast-follow)` section covering registry / dispatch / merge with MCP tools / name-collision policy / policy semantics / Read/Glob/Grep specs / path confinement rule
- Update Non-goals: the "no file editing / shell" statement is being scoped down — clarify that FF-1a/b/c are landing this capability progressively

## Acceptance criteria

- `EMBEDDED_AGENT_TOOL_NAMES` is the single writer for tool names; grep proves no duplicate literal list elsewhere
- Migration vNN is added and reversible; local `bun run db:migrate` succeeds
- Schema + valibot + mapper all synchronized; wire test verifies the field survives WS round-trip (per Q10 / `pre-pr-completeness.md`)
- Registry rejects duplicate `enabledTools` entries (validation polarity test)
- Policy-off tools are **not present in the provider's tools list** (polarity test: with `enabledTools: []` in the definition, the request payload sent to the provider has zero builtin tools)
- Name-collision detection: if an MCP tool has the same name as a builtin, builtin wins and a warn is logged (unit test)
- Read / Glob / Grep produce results identical in shape to Claude Code's equivalents for a representative fixture
- Path confinement polarity test: `Read('/etc/passwd')` and `Read('/home/other-user/private.txt')` are rejected with `ok: false`; `Read('<locationPath>/README.md')` succeeds
- Symlink escape polarity test: a symlink from within `locationPath` targeting outside is rejected
- Secret non-leakage unit test: the builtin execution context does not expose apiKey / token strings to the tool implementation
- UI: Read-only group is checked by default; Command execution group shows amber warning label
- Existing embedded-agent tests continue to pass; existing MCP tools continue to work (integration test with both a builtin and an MCP tool in one turn)
- Browser QA: create an agent with default `enabledTools`, ask it to Read a file inside `locationPath` and observe the tool-result rendered in the transcript

## Change surface

Moderate-to-high — new package directory, schema/migration/mapper/UI/spec all touched. Recommend a single specialist owning this end-to-end.

## Cross-reference

- Umbrella: #1004
- Spec: `docs/design/embedded-agent-worker.md` Part II
- Follow-on issues: FF-1b (Bash), FF-1c (Write/Edit), FF-2 (sandbox & MCP filtering)
- Related: #1040 (MessagePanel reuse — attachment enable depends on FF-1a Read)

ms2sato:
## Summary

Fast-follow FF-1a for the Embedded Agent Worker (umbrella #1004). Adds subprocess-local built-in tools to the embedded-agent loop, matching Claude Code's own tool shapes, per owner direction: "Do not put tools on the MCP surface. Implement them as subprocess-internal tools like Claude Code / opencode. Start with a subset from reference CLIs; do not reinvent the wheel."

- **`packages/embedded-agent/src/tools/`** (new): a builtin tool registry (`Read`, `Glob`, `Grep`; `Bash` is FF-1b scope), a `CompositeToolExecutor` that merges builtin tools with MCP tools (builtin wins on name collision, logged), and `resolveConfinedPath` — the path-confinement floor (realpath resolve + `locationPath` prefix check, symlink-escape rejected).
- **`enabledTools` policy** on `EmbeddedAgentDefinition`: `undefined` = default read-only set (`Read`/`Glob`/`Grep`), `[]` = all builtin tools off, explicit array = exact set. Policy-off tools are not represented in the provider's tools list at all (unrepresentable, not merely rejected).
- **DB migration v23**: nullable `embedded_agents.enabled_tools TEXT` column (JSON array string).
- **UI**: a "Tools" section in the Add/Edit Embedded Agent form — Read-only group checked by default, Command execution (Bash) group unchecked with an amber warning label.
- **Spec update**: `docs/design/embedded-agent-worker.md` Part II gains a "Built-in tools (fast-follow)" section; `docs/glossary.md` gains a "Builtin Tool (embedded-agent)" entry.

## Test plan

- [x] `EMBEDDED_AGENT_TOOL_NAMES` (`packages/shared/src/types/embedded-agent.ts`) confirmed as the single writer of tool-name literals via repo-wide grep.
- [x] Unit tests: registry (`resolveEnabledBuiltinTools` incl. `[]` polarity and the currently-inert `Bash` case), each tool (`Read`/`Glob`/`Grep`) incl. path-confinement polarity (outside path, symlink escape, boundary-equals-locationPath), `CompositeToolExecutor` (merge, name-collision-builtin-wins, dispatch routing, and a secret-non-leakage exact-equality regression guard on the execution context passed to a builtin tool).
- [x] Process-boundary polarity test in `packages/embedded-agent/src/__tests__/main.test.ts`: `enabledTools: []` in the `init` command results in zero builtin tools reaching the provider's tools list; `enabledTools` absent results in the default three.
- [x] Wire tests: REST create/update round-trip (`packages/server/src/routes/__tests__/embedded-agents.test.ts`), stdin `init` command construction (`embedded-agent-worker-service.test.ts`), repository UPDATE-path round-trip for `enabledTools` (`sqlite-embedded-agent-repository.test.ts` — this path is easy to miss since `save()`'s `onConflict().doUpdateSet()` lists columns explicitly).
- [x] **Real E2E** (`packages/integration/src/embedded-agent-e2e.test.ts`): extended the existing shipping-path test (real subprocess, real `/mcp` server, scripted stub provider) so the same turn requests both an MCP tool (`list_sessions`) and a builtin tool (`Read`, reading a real file from the session's `locationPath`) — proves `CompositeToolExecutor` merges both sources over the real `init → MCP-connect → listTools` path, not just at the unit level.
- [x] `bun run typecheck` — clean across all packages.
- [x] `bun run test` (full monorepo suite) — green. Two unrelated environment-contamination flakes were observed and ruled out during this work (both reproduce/disappear independent of this diff): a pre-existing `multi-user-mode.test.ts` assertion sensitive to the ambient `SSH_AUTH_SOCK` env var (1Password agent socket in this dev environment), and an isolated `resolve-targets.test.ts` failure that passes in isolation and does not reproduce on a clean re-run. Neither file is touched by this PR. CI runs in a clean environment without `SSH_AUTH_SOCK` set, so this should not surface there.
- [x] Circular-dependency fix (`c7729bc`): `madge --circular` flagged `tools/index.ts` importing `read.ts`/`glob.ts`/`grep.ts` which imported their shared types back from `index.ts`. Extracted `BuiltinTool`/`BuiltinToolContext`/`BuiltinToolResult` into `packages/embedded-agent/src/tools/types.ts`; `index.ts` re-exports them so all other consumers are unaffected. `bun run lint:cycles` now reports zero circular dependencies.
- [x] `node .claude/skills/orchestrator/preflight-check.js` — all touched production files have sibling tests, integration test coverage present, except two known false positives (not fixed in this PR, coordinating with the Orchestrator separately):
  - `packages/shared/src/schema-version.gen.ts` — auto-generated content-hash file (`scripts/generate-schema-version.mjs`), same category as `routeTree.gen.ts`.
  - `packages/embedded-agent/src/tools/types.ts` — pure type declarations (interfaces only), no executable code to test. Introduced by the circular-dep fix above.
- [x] **Browser QA** (dev server at a non-default port in this worktree; `localhost:15173`): created an embedded agent via the Add form — confirmed Read/Glob/Grep checked by default, Bash unchecked with the amber warning label rendered (`Runs arbitrary shell commands as the session user.`). Submitted with Bash also checked; verified via the DevTools Network panel that the `POST /api/embedded-agents` request body carried the explicit `enabledTools: ["Read","Glob","Grep","Bash"]` array and the response persisted it. Opened the Edit form and confirmed all four checkboxes pre-populated as checked (matching the saved state). Unchecked Grep and Bash and saved; verified the `PATCH /api/embedded-agents/:id` request body carried the exact reduced `enabledTools: ["Read","Glob"]` array and the response reflected it — this also live-confirms the `sqlite-embedded-agent-repository.ts` UPDATE-path fix (the easy-to-miss `doUpdateSet` column list) actually persists changes, not just inserts.

## Note on CodeRabbit

The GitHub-side bot hit its review-rate-limit at PR open ("next review available in 21 minutes"). The local CodeRabbit CLI cannot authenticate in this headless worktree environment (known constraint). The stated recovery window passed with no bot action; retriggered via an `@coderabbitai review` PR comment, which returned: "CodeRabbit is an incremental review system and does not re-review already reviewed commits. This command is applicable only when automatic reviews are paused" — i.e. the bot treated its own rate-limited attempt as a completed review and will not re-review this commit. Per the `coderabbit-ops` skill's "abandon-the-wait" disposition (do not force further retriggers), CodeRabbit's verdict is unavailable for this PR from either source. Per PR Merge Authority, this is production code and requires owner approval before merge regardless — CodeRabbit's unavailability does not change that gate, it only means the automated review layer is absent for this PR.

## Known follow-up (tracked separately, not blocking this PR)

`preflight` CI check currently fails on the two false positives above (`schema-version.gen.ts`, `tools/types.ts`) — the Orchestrator is landing a fix to `.claude/skills/orchestrator/check-utils.js`'s coverage-exclusion patterns in a separate PR; this branch will rebase onto it once merged.

Closes #1042

## Architect independent audit — 2 required fixes (landed)

The architect's independent audit (separate from Orchestrator acceptance) found two merge-blocking issues, both fixed in this PR:

1. **R1 (MAJOR) — `CompositeToolExecutor.callTool` didn't catch builtin-tool exceptions.** `resolveConfinedPath` throws (not `{ok:false}`) on ENOENT/EACCES (e.g. the session's worktree removed while the embedded-agent worker is still alive, or multi-user permission errors), and the builtin dispatch branch let that escape uncaught — breaking the never-throws `ToolExecutor` contract `AgentLoop.runTurn` relies on. Consequence chain: no `turn-error`/`state:idle` emitted → client activity stuck `'active'` → send permanently disabled (PR #1048's `sendDisabled={isTurnActive}` gate) → worker restart required to recover, plus a dangling `tool_call_id` that a strict OpenAI-compatible provider rejects on the next turn. Fixed with `try { return await builtin.execute(...) } catch (err) { return {ok:false, result: message} }` (the `await` inside `try` is load-bearing — a bare `return builtin.execute(...)` would let the rejection escape past the catch). Polarity-tested (fails against unfixed code, passes after).
2. **R2 (required, UX) — misleading Bash checkbox warning.** The amber warning under the Bash checkbox read "Runs arbitrary shell commands as the session user." even though `Bash` has no registry implementation yet (FF-1b) and checking it is currently a no-op. Changed the caption to state it's not yet available, so the UI doesn't imply a live capability that doesn't exist.

CI re-verified green after both fixes (typecheck + full suite + push).


ms2sato:
@coderabbitai review

PR #1063, opened 2026-07-13
embedded-agent: thread AbortSignal through BuiltinTool.execute (Bash pgid kill) (closes #1052)

### LINKED ISSUE #1052: embedded-agent: thread AbortSignal through BuiltinTool.execute (cancel + FF-1b Bash prerequisite)
Non-blocking follow-up surfaced by architect audit of PR #1050 (FF-1a landed 2026-07-12).

## Context

PR #1050 shipped built-in tools (Read/Glob/Grep) via `CompositeToolExecutor` + `BuiltinTool.execute(args, ctx)`. There is currently no way to signal cancellation:

- Long-running `Glob` / `Grep` (regex backtracking on a big tree) cannot be interrupted from the outside
- Turn-level cancel / worker restart cannot interrupt an in-flight builtin call
- **FF-1b (#1043) Bash needs an abort signal from day one** — a shell command whose timeout fires must be killable, and process-group SIGTERM+SIGKILL requires a signal source

## Scope

- Extend `BuiltinTool.execute` signature to accept an `AbortSignal` (via `BuiltinToolContext` or as a fourth param)
- Wire through `CompositeToolExecutor.callTool`
- Integrate with `AgentLoop` cancel path (turn abort emits abort, restart emits abort)
- Implementations honor the signal:
  - `Read`: check between line chunks (small win)
  - `Glob` / `Grep`: check between traversal steps / matches
  - Bash (FF-1b): kill the child process group on signal

## Acceptance criteria

- Polarity test: a builtin whose execution is long-running, called with an abort signal that fires mid-execution → resolves early with `{ok: false, result: 'aborted'}` (no throw, ToolExecutor never-throws contract preserved)
- Existing Read/Glob/Grep tests still pass (signal is optional in the type; unpassed = never fires)
- Integration test: turn cancel during a builtin call causes the tool-result to indicate abort

## Timing recommendation

Land BEFORE or ALONGSIDE FF-1b (#1043) — Bash needs this signal to implement the process-group kill correctly.

## Cross-reference

- Umbrella: #1004
- Preceding: PR #1050 (FF-1a)
- Blocks / co-requires: #1043 (FF-1b Bash)

ms2sato:
## Summary

Implements Issue #1052 (AbortSignal threading through `BuiltinTool.execute`, including the architect's follow-up scope from the PR #1057 audit: Bash process-group kill on abort, UTF-8 chunk-boundary fix, and a `[Killed by signal]` marker).

- `packages/embedded-agent/src/tools/types.ts` — `BuiltinTool.execute` gains an optional third parameter, `signal?: AbortSignal`. Deliberately kept as a separate call-scoped parameter rather than folded into `BuiltinToolContext` (which stays a fixed per-session confinement root — see its own doc comment and the composite-executor test asserting ctx has exactly `{ locationPath }`).
- `packages/embedded-agent/src/tools/composite-executor.ts` — `callTool` now forwards the `signal` it already receives (the loop's per-turn `AbortController.signal`) into `builtin.execute(args, ctx, signal)`, closing the one dispatch site that was silently dropping it.
- `packages/embedded-agent/src/tools/bash.ts` — the main change:
  - `runBash` accepts `signal?: AbortSignal` in `RunBashOptions`. An `'abort'` listener triggers the same SIGTERM → 2s grace → SIGKILL process-group kill sequence already used on timeout (factored into a shared local `killProcessGroup()` closure so both paths call identical logic). Handles the already-aborted-before-call race by checking `signal.aborted` synchronously right after attaching the listener. Removes the listener in `settle()` to avoid leaking it. Final settlement still happens from `child.on('close', ...)`, never directly from the abort listener, so `exitCode` reflects reality.
  - `RunBashResult` gains `aborted: boolean`; `ok` becomes `!timedOut && !aborted`.
  - stdout/stderr decoding switched from per-chunk `Buffer#toString('utf-8')` to a persistent `node:string_decoder` `StringDecoder` per stream, flushed once via `.end()` at settle time (before truncation) — fixes multi-byte UTF-8 characters split across `data`-event chunk boundaries producing `<0xEF><0xBF><0xBD>` replacement characters.
  - `formatBashResult` gains an `aborted` branch (`[Command aborted and its process group was terminated.]`) and a `[Killed by signal]` marker for `exitCode === null` external-signal deaths (e.g. OOM/manual `kill -9`), ordered most-specific-first: `aborted` → `timedOut` → `exitCode === null` → `exitCode !== 0`.
  - `bashTool.execute` passes `signal` through to `runBash`.
- `packages/embedded-agent/src/tools/read.ts` — checks `signal?.aborted` right before starting the single `.text()` read (a "small win", per the issue — it can't interrupt an in-flight read, but avoids racing an already-aborted turn's result into the conversation).
- `packages/embedded-agent/src/tools/glob.ts` / `grep.ts` — check `signal?.aborted` between traversal/matching steps (`Glob.scan()`'s `for await` loop; `Grep`'s file-collection scan loop and its per-file matching loop) and return `{ ok: false, result: 'aborted' }` (the issue's literal AC shape) the moment it's observed set. Never throws.
- `docs/design/embedded-agent-worker.md` — `### Bash (FF-1b)` subsection gains a paragraph documenting the abort-signal contract, the source of the signal (loop → composite-executor → builtin), the process-group-kill reuse, the per-tool check points, the `StringDecoder` fix, and the `[Killed by signal]` marker.

**Verified before starting:** `packages/embedded-agent/src/agent-loop.ts` already creates a per-turn `AbortController` and already passes `abort.signal` into `executor.callTool(...)`; `loop.cancel()` already calls `this.currentAbort?.abort()` on the `cancel` command and via `gracefulExit()` on `shutdown`/stdin-EOF. Server-side, `EmbeddedAgentWorkerService.deactivate()` (`packages/server/src/services/embedded-agent-worker-service.ts:474`) already sends `{ type: 'shutdown' }` before killing the subprocess as part of worker restart. Both were re-read before this PR started; **no changes were made to `agent-loop.ts`, `main.ts`, or `packages/server/`** — `git diff --stat -- packages/server` against `main` is empty.

## Signature-shape change pre-estimate (`pre-pr-completeness.md` Q8)

`grep -rn "\.execute(" packages/embedded-agent/src` → 29 call sites total: 1 production (`composite-executor.ts`'s dispatch, updated) + 28 test call sites across `bash.test.ts` (5), `read.test.ts` (6), `glob.test.ts` (5), `grep.test.ts` (12). The new parameter is optional, so all 28 pre-existing 2-arg test call sites remain valid unchanged; only the 5 new/modified test cases pass a signal explicitly.

## Test plan

- [x] `bun run typecheck` — exit 0, all 4 packages.
- [x] `bun run test` (full suite) — green except the same pre-existing, unrelated flake documented in PR #1057 (see below), which was independently re-verified standalone in this PR.
- [x] `packages/embedded-agent/src/tools/__tests__/` (8 files, 61 tests) — includes: Bash abort-mid-execution polarity test (real `sleep 300`, `AbortController.abort()`, asserts `{ok:false, aborted:true, timedOut:false}` resolves within ~2.5s worst case, and a `process.kill(pid, 0)` liveness probe confirms the process group is actually dead — same technique as the existing timeout-kill test); an already-aborted-before-call race test; a UTF-8 chunk-boundary reassembly test (two `printf` calls emitting the two bytes of `é` with a `sleep` between them, defeating OS pipe coalescing, asserting the reassembled string has no `�`); a `composite-executor` test asserting the exact `signal` object is forwarded to `builtin.execute`'s third arg; and an already-aborted test per `read`/`glob`/`grep` asserting `{ok:false, result:'aborted'}` without throwing. All 8 pre-existing Read/Glob/Grep/Bash test assertions are unmodified except one strengthened assertion (the existing `kill -9 $$` test also now asserts the new `[Killed by signal]` marker is present, in addition to its original `not.toContain('Exit code: null')` assertion — nothing was weakened).
- [x] `bun run check:lang` — exit 0 (109 files scanned, clean).
- [x] `node .claude/skills/orchestrator/preflight-check.js` (run post-commit per its diff-scope caveat) — all 5 modified production files covered by sibling tests, no rule/skill duplication, no source-comment blame-shift. (`types.ts` is exempt per `test-trigger.md`'s bare-`types.ts` convention.)
- [ ] CodeRabbit — local CLI could not authenticate in this headless delegated worktree (`automatic_login_failed`, same constraint documented in PR #1057); relying on GitHub-side bot review post-push.
- [x] Multi-user smoke (`scripts/smoke/check-embedded-agent-bash-env.ts`) — ran in degenerate mode (`bun scripts/smoke/check-embedded-agent-bash-env.ts "$(whoami)"`, current-user target, no second-user elevation available in this sandbox): **6/6 assertions passed**. This PR does not change env-handling logic, only adds abort-listening on top of the same spawn/kill mechanism, so this is a sanity re-check rather than new-behavior verification.
- No UI changes in this PR.

## Pre-existing flake: `resolve-targets.test.ts`

Same flake as documented in PR #1057: `packages/server/src/services/inbound/__tests__/resolve-targets.test.ts` fails only inside the full `packages/server` suite run (`GitError: not a git repository`, exit 128) but passes standalone, re-verified independently in this PR:

```
$ cd packages/server && bun test src/services/inbound/__tests__/resolve-targets.test.ts
 9 pass
 0 fail
Ran 9 tests across 1 file. [90.00ms]
```

This PR makes zero changes to `packages/server` (confirmed via `git diff --stat -- packages/server` against `main`, empty output).

Closes #1052

## Summary by CodeRabbit

* **New Features**
  * Added cancellation support for built-in tools during active operations.
  * Bash commands now stop promptly when canceled and clearly report whether they were aborted or terminated by a signal.
  * File reading, searching, and matching operations now stop safely when cancellation occurs.
  * Improved handling of streamed UTF-8 command output to prevent corrupted characters.

* **Documentation**
  * Documented cancellation behavior for embedded-agent tools, including shutdown scenarios.


ms2sato:
@coderabbitai review

Issue #1077, opened 2026-07-13
CreateWorktreeForm draft-restore effect races with StrictMode double-invoke (dev-only visibility)


ms2sato:
## Context

Discovered during Browser QA for #1062 (worktree form localStorage embeddedAgentId validation). Not introduced by that PR — reproduces identically on a field that predates it (`sessionTitle`).

## Symptom

`CreateWorktreeForm`'s draft-restore-on-mount effect (`packages/client/src/components/worktrees/CreateWorktreeForm.tsx`) intermittently fails to restore any localStorage draft field when running under `bun run dev` (which wraps the app in `<StrictMode>` per `packages/client/src/main.tsx`).

## Root cause

React 18 StrictMode double-invokes effects in development builds only (mount → run effects → simulate unmount by running all cleanups → run effects again). The component has two separate effects:

1. The draft-restore effect (runs `reset(...)` / `setValue(...)` to hydrate form state from `localStorage`).
2. A save-debounce effect that subscribes to `watch()` and, on cleanup (real or StrictMode-simulated unmount), unconditionally writes `pickDirtyValues(getValues())` back to `localStorage`.

`pickDirtyValues` reads dirtiness from `dirtyFieldsRef.current`, a plain `useRef` that is only kept in sync with `formState.dirtyFields` during render (`dirtyFieldsRef.current = dirtyFields;` in the component body). StrictMode's simulated unmount runs the save-effect's cleanup *before* a re-render has occurred to reflect the dirty state produced by the draft-restore effect's `setValue`/`reset` calls. The cleanup therefore reads a stale, empty `dirtyFieldsRef.current` and writes `{}` back to `localStorage`, clobbering the draft. The effects then run a second time (StrictMode's second invocation), but by then the draft in `localStorage` is already gone.

Reproduced for both:
- `embeddedAgentId` (a `setValue`-only, never-`register()`-bound field)
- `sessionTitle` (a normal `register()`-bound field) — confirming this is not specific to the `agentId`/`embeddedAgentId` restore mechanism, but a general race in the draft-persistence effect pair.

## Confirmed NOT a production issue

React's production runtime does not double-invoke effects regardless of the `<StrictMode>` wrapper. Verified via an actual `vite build` + `NODE_ENV=production bun dist/index.js` run (fresh port, fresh DB): both the stale-ID-fallback and valid-ID-restore scenarios worked reliably and repeatably. Real end users on the production build are unaffected.

## Impact

[...]

PR #1078, opened 2026-07-13
fix: validate embeddedAgentId against registry + fix stale draft restore (#1062)

### LINKED ISSUE #1062: worktree form: validate localStorage draft embeddedAgentId existence on hydration (parity with terminalValueExists fallback)
Follow-up from PR #1058 audit (architect follow-up #3).

## Context

The worktree creation form persists user selections to localStorage as a draft. Terminal-agent selections have a `terminalValueExists` fallback: on hydration, if the stored `agentId` no longer refers to an existing agent (deleted between sessions), the form falls back to the repository's `default_agent_id` (or the first available terminal agent).

**Embedded-agent selections have no equivalent existence check.** A stored `embeddedAgentId` that no longer exists will hydrate the form with the stale ID; the user submits, and the server rejects with 400 (because #1058 added route-level pre-validation for `POST /api/worktrees`). The user experience is:

1. User opened the form previously with embedded agent "X" selected
2. Between sessions, admin deleted embedded agent "X"
3. User re-opens the form; embedded selector shows blank / broken state (or the ID is stored but no option matches)
4. Submit → 400 error, user has to re-select

## Scope

- On form hydration (probably in `CreateWorktreeForm.tsx` or the `WorktreeAgentSelector` component), check whether the stored `embeddedAgentId` matches an entry in the current `useEmbeddedAgents` list
- If not, silently fall back to unset (no selection) or to the terminal-agent default — mirroring `terminalValueExists`
- Preserve the intent: don't discard the draft's other fields, just fix the invalid ID

## Acceptance criteria

- Unit test: hydrate the form with a stored `embeddedAgentId` that does not exist in the current fetch → the form shows no error, falls back cleanly
- Regression test: hydrate with a valid `embeddedAgentId` → the form pre-populates the embedded agent selection as expected
- Polarity: remove the existence check → the invalid-ID hydration reproduces the broken state; restore → clean

## Priority

Low. The server 400 fails fast (no data corruption), and the affected UI window is narrow (embedded agent list is stable across most session boundaries). Fixing this improves polish; not fixing it is a mild user annoyance.

## Cross-reference

- Umbrella: #1004
- Preceding: [PR #1058](https://github.com/ms2sato/agent-console/pull/1058)
- Sibling: existing `terminalValueExists` fallback (search codebase for the exact function name)

ms2sato:
## Summary

Fixes #1062: the worktree creation form's `embeddedAgentId` had no existence-check fallback against the current embedded-agent registry, unlike the terminal `agentId` (which already had `useResolvedAgentId`). A stale `embeddedAgentId` referencing a deleted embedded agent could survive localStorage draft hydration and reach the submit request, which the server's route-level pre-validation (added in #1058) rejects with a 400.

## Scope: two fixes, both needed (transparent disclosure)

**1. Existence-check parity (the originally-scoped fix)**

- New `useResolvedEmbeddedAgentId` hook in `AgentSelector.tsx`, mirroring `useResolvedAgentId`. Key difference: embedded selection is optional, so an unknown/stale value falls back to `undefined` (unset), not the first embedded agent.
- `WorktreeAgentSelector`'s internal `selectedValue` computation gets an `embeddedValueExists` check, paralleling the existing `terminalValueExists` check.
- `CreateWorktreeForm.tsx` uses the resolved value (not the raw watched value) in `buildRequest` and as the selector prop.

**2. Root-cause fix discovered during TDD polarity verification (scope expansion, approved by the Orchestrator)**

While verifying TDD polarity for fix #1, both a `WorktreeAgentSelector`-level test and the `CreateWorktreeForm` stale-draft regression test showed no real polarity (passed identically with/without the fix). Investigation traced this to react-hook-form 7.68.0's `reset()` implementation:

```js
V = d.shouldUnregister ? (t.keepDefaultValues ? clone(h) : {}) : clone(i)
```

`V` is the internal `_formValues` read by `watch`/`getValues`. `h` is the original `useForm()` constructor `defaultValues`. When `shouldUnregister: true` (set on this form) and `keepDefaultValues: true` (passed by the draft-restore effect), `V` becomes `clone(h)` — **completely discarding whatever was passed into `reset(...)`** for any field that was never `register()`-bound. `agentId` and `embeddedAgentId` are exactly such fields (driven only by `watch`/`setValue` for the custom `WorktreeAgentSelector`, never through `register()`).

**Consequence:** the draft-restore effect's `reset({ ...getValues(), ...draft }, { keepDefaultValues: true })` never actually restored `agentId` or `embeddedAgentId` from localStorage — for any value, valid or stale. **This affected the terminal `agentId` side too, not just embedded** — restoring a previously-selected terminal agent from a draft silently never worked either.

**Fix:** explicit `setValue('agentId', ...)` / `setValue('embeddedAgentId', ...)` calls right after `reset()` in the draft-restore effect, since `setValue` writes through a path not subject to the discard above.

## Layered TDD polarity verification (paste, not summary)

- Stash existence-check only, keep `setValue` fix: stale-draft regression test **FAILS** (`embeddedAgentId: "nonexistent-embedded"` reaches the submitted request) — proves the existence-check layer is necessary, not redundant.
- Stash `setValue` fix only, keep existence-check: new valid-`agentId` restore test **FAILS** (falls back to `defaultAgentId` instead of the drafted `agentId`) — proves the `setValue` fix is what makes restoration work at all.
- Full baseline (both reverted): `AgentSelector.test.tsx` fails to load (missing export); `CreateWorktreeForm` valid-agentId test fails; stale-draft test passes trivially (no restore happens at all, not neutralization).
- Full fix restored: `AgentSelector.test.tsx` 22/22 pass; `CreateWorktreeForm.test.tsx` 37/37 pass.

## Test results

```
bun run typecheck: exit 0 (all packages clean)
[client] 1839 pass, 0 fail, 3821 expect() calls, 130 files
[shared] 508 pass, 0 fail
[integration] 58 pass, 0 fail
[embedded-agent] 170 pass, 0 fail
[server] 3380 pass, 1 skip, 1 fail — packages/server/src/services/inbound/__tests__/resolve-targets.test.ts
  ("fatal: not a git repository"), reproduces identically on origin/main, passes in isolation,
  fails only in the full 150-file server suite. Pre-existing test-fixture-state issue; zero
  packages/server files touched by this diff.
```

## Browser QA (real end-to-end, both dev build and production build)

Drove the actual `QuickWorktreeDialog` (the component that passes `draftKey`) via Chrome DevTools MCP against a real running server, with a genuinely-registered embedded agent present in the dropdown (true-path, not "no embedded agents available" masking the bug):

- **Stale ID**: seeded an invalid `embeddedAgentId` draft → form correctly falls back to the terminal default, with the real embedded agent still visible/selectable in the dropdown. Submitted → `POST /api/worktrees` returned **202** with `{"agentId":"claude-code-builtin", ...}` and no `embeddedAgentId` key — confirms the exact 400-causing bug no longer reproduces through the real shipping path.
- **Valid ID** (bonus fix): seeded a real embedded agent's ID as the draft value → form correctly restores it as selected.

Screenshots (dropdown open, showing the real embedded agent option alongside the correctly-resolved selection) are attached via a follow-up comment on this PR (screenshot upload convention per `browser-qa` skill).

### A finding during QA, investigated and resolved as non-blocking (see #1077)

While testing the valid-ID restore under `bun run dev` (StrictMode-wrapped dev build), it intermittently failed to restore. Traced to React 18 StrictMode's dev-only double-invoke-effects behavior racing the draft-restore effect against a second effect's "save dirty fields on unmount" cleanup (a stale `dirtyFieldsRef` clobbers the just-restored draft between the two invocations). Confirmed this is **pre-existing and not specific to this PR** — reproduced identically for `sessionTitle`, an unrelated `register()`-bound field. Confirmed via an actual `vite build` + `NODE_ENV=production` run that **production is unaffected** (React's production runtime doesn't double-invoke effects). Filed as a separate, non-blocking follow-up: #1077.

## Test plan

- [x] Unit tests: existence-check hook (5 cases) + `WorktreeAgentSelector` fallback + `CreateWorktreeForm` stale-draft regression + valid-`agentId` restore regression
- [x] TDD polarity verified per-layer (see above)
- [x] `bun run typecheck && bun run test` full suite paste
- [x] Browser QA: stale-ID fallback + valid-ID restore, both via dev build and production build, real POST payload confirmed
- [x] `node .claude/skills/orchestrator/preflight-check.js` clean (coverage, integration-test advisory noted below, language check, blame-shift check)
- [ ] CodeRabbit 3-layer verdict (pending post-push)

**Integration test advisory:** `preflight-check.js` flagged both changed files as UI components with no integration-test changes. This fix is entirely client-side (no new field added to a shared wire type — `embeddedAgentId` already existed on `CreateWorktreeRequest`), so it doesn't match `pre-pr-completeness.md` Q10's specific trigger (derived field crossing the wire), but flagging per the advisory for reviewer awareness.

Closes #1062

### DIFF
--- a/packages/client/src/components/AgentSelector.tsx
+++ b/packages/client/src/components/AgentSelector.tsx
@@ -93,6 +93,19 @@ export function useResolvedAgentId(
   return valueExists ? value : sortedAgents[0]?.id;
 }
 
+/**
+ * Hook that resolves the effective embedded agent ID with fallback logic.
+ * Unlike useResolvedAgentId, embedded agent selection is optional: an unknown/stale
+ * value falls back to undefined (unset) rather than the first embedded agent.
+ */
+export function useResolvedEmbeddedAgentId(value: string | undefined): string | undefined {
+  const { embeddedAgents, isLoading } = useEmbeddedAgents();
+
+  if (isLoading) return value;
+  const valueExists = value != null && embeddedAgents.some((a) => a.id === value);
+  return valueExists ? value : undefined;
+}
+
 export function useAgents() {
   const { data, isLoading, error, refetch } = useQuery({
     queryKey: agentKeys.all(),
@@ -146,7 +159,9 @@ export function WorktreeAgentSelector({
   const isLoading = agentsLoading || embeddedLoading;
 
   const terminalValueExists = agentId != null && sortedAgents.some((a) => a.id === agentId);
-  const selectedValue = embeddedAgentId
+  const embeddedValueExists =
+    embeddedAgentId != null && embeddedAgents.some((a) => a.id === embeddedAgentId);
+  const selectedValue = embeddedValueExists
     ? `embedded:${embeddedAgentId}`
     : terminalValueExists
       ? `terminal:${agentId}`
--- a/packages/client/src/components/__tests__/AgentSelector.test.tsx
+++ b/packages/client/src/components/__tests__/AgentSelector.test.tsx
@@ -2,7 +2,13 @@ import { describe, it, expect, mock, beforeEach, afterEach, afterAll } from 'bun
 import { render, screen, waitFor, cleanup, renderHook } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { AgentSelector, useResolvedAgentId, WorktreeAgentSelector } from '../AgentSelector';
+import {
+  AgentSelector,
+  useResolvedAgentId,
+  useResolvedEmbeddedAgentId,
+  WorktreeAgentSelector,
+} from '../AgentSelector';
+import { useEmbeddedAgents } from '../../hooks/useEmbeddedAgents';
 
 // Save original fetch and set up mock
 const originalFetch = globalThis.fetch;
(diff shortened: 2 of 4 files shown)


ms2sato:
## QA Screenshots (2026-07-14 07:37)

### stale embeddedagentid falls back to terminal
stale embeddedagentid falls back to terminal

<details>

stale embeddedagentid falls back to terminal
</details>

### valid embeddedagentid restores correctly
valid embeddedagentid restores correctly

<details>

valid embeddedagentid restores correctly
</details>

ms2sato:
@coderabbitai review

ms2sato:
## Note on CodeRabbit

Both review paths were unavailable for this PR's review window:
- **GitHub-side bot**: rate-limited at PR open ("Review limit reached... Next review available in 6 minutes"). Retriggered via `@coderabbitai review` after the window cleared, but CodeRabbit's incremental-review policy treated the rate-limited event as "already reviewed" and declined to actually re-review ("This command is applicable only when automatic reviews are paused").
- **Local CLI**: `coderabbit review --agent --base main` failed with an authentication error (`Automatic login timed out`) — the CLI requires interactive browser auth, which is not available in this headless delegated worktree.

No CodeRabbit feedback obtainable from either source. Per `.claude/skills/coderabbit-ops/troubleshooting.md`'s simultaneous-unavailability disposition, this follows existing PR Merge Authority rather than blocking indefinitely — deferring to the Orchestrator for that decision given this PR is a bug-fix (logic change) requiring owner-channel approval per project convention.

PR #1090, opened 2026-07-14
fix: signal entire PTY process tree on managed worker kill (closes #1087)

### LINKED ISSUE #1087: worktree removal leaks child claude PTY processes — memory exhaustion on long-running servers
## Symptom

On a long-running agent-console server (owner observation: PID 2943372, 2 days 19 hours uptime), the OS reported memory exhaustion. Manual `kill` of accumulated stray `claude` processes was required to recover.

Each removed worktree left behind at least one child `claude` process that was NOT terminated by the `remove_worktree` MCP tool (or the equivalent worktree/session cleanup path). Over time, these orphaned processes accumulated to full memory exhaustion.

## Impact

- **Severity**: high — resource exhaustion is a hard-fail mode that eventually kills the server (or forces manual intervention)
- **Reach**: every worktree removal cycle on a long-running server; scales with delegate-heavy workflows (Sprint 2026-07-11/12 delegate cascade exercised this pattern intensively)
- **Silent accumulation**: no logs surface the leak until the server is memory-starved

## Root cause hypothesis

`mcp__agent-console__remove_worktree`'s description states "kills PTY processes, removes the worktree via git, and deletes the session". Suspect that the PTY termination path sends `SIGTERM` to the shell process only, and does not propagate through the process tree to child `claude` processes (Claude Code CLI spawned inside the PTY).

Likely fix directions (needs code review to confirm):
1. Assign a POSIX process group to each spawned PTY (`setsid` or equivalent), then kill by pgid at teardown (`kill(-pgid, SIGTERM)` and later `SIGKILL`).
2. Alternatively, walk `/proc/*/ppid` (or `pgrep -P <pid>` recursively) to enumerate all descendants before killing.

Same-class prior art in this codebase: builtin `Bash` tool already uses `spawn(..., { detached: true })` + pgid kill for its subprocess (per `docs/glossary.md` entry for Built-in tools / Bash). The pattern exists — it just isn't applied to worker PTYs.

## Reproduction

1. Start agent-console server (`bun run dev` or equivalent).
2. Note the current claude process count on the host: `pgrep -c claude`.
3. Delegate to a worktree that starts an agent worker (e.g., via `delegate_to_worktree` with a Claude Code agent).
4. Wait for the delegate to complete + self-remove its worktree (or explicitly call `remove_worktree`).
5. Re-check `pgrep -c claude` — a leftover claude process remains.
6. Repeat several cycles; each cycle adds one persistent process, and each process holds ~100+ MB.

## Related architectural invariant

Directional analog of I-8 "Shared-Resource Artifact Lifetime" in `.claude/skills/architectural-invariants/SKILL.md`, but the resource class is different: I-8 covers artifacts (files, symlinks) whose lifetime must span consumer lifetimes; this bug is about **spawned-process lifetime that must not survive its spawn context**.

Consider whether this warrants a new I-9 "Spawned-Process Lifetime Follows Spawn Context" entry, or an I-8 extension. Discussion for the retro / brewing skill.

## Sprint context (contributed observations)

Sprint 2026-07-11/12 exercised delegate-heavy patterns:
- Wave 5-3 (dogfood 4-fix bundle) — 1 delegate
- Wave 5-4 (AGENTS.md loader) — 1 delegate + 1 dogfood delegate
- Slots 2 & 3 (parallel dispatch) — 2 delegates
- Owner-goal 3-parallel finish delegates (UI polish + persistence + UI editor) — 3 delegates
- Post-merge dogfood verifications — 1 delegate

Total ~9 delegates during Sprint 2026-07-11/12 on the same long-running server. Every self-cleanup after PR merge likely leaked one claude process (verified by owner-driven kill).

## Recommended priority

**High.** The leak is progressive and eventually fatal; delegate-heavy sprints (which are becoming more common with the 3-worktree parallelism policy) accelerate it. Fix before the next comparable sprint.

## Related

- `mcp__agent-console__remove_worktree` (implementation TBD-located in `packages/server/src/`)
- `docs/glossary.md` "Built-in tools (fast-follow)" Bash entry — cites the pgid-kill pattern already used elsewhere
- `.claude/skills/architectural-invariants/SKILL.md` I-8 — adjacent invariant
- Sprint 2026-07-11/12 delegate cascade session (owner observation of PID 2943372 memory exhaustion)

ms2sato:
## Summary

- `killWorker()` previously called `pty.kill()`, which signals only the PTY's root pid. That root pid is a login shell (see `sentinel-spawn-command.ts`), not the agent binary — the agent command is typed into the shell like a human user (`worker.pty.write(pendingCommand + '\r')`), so job control puts it in its own process group. Killing only the shell never reaches the agent process (or its own children); once the shell exits, the agent gets reparented to init and leaks indefinitely.
- Adds `packages/server/src/lib/process-tree.ts`, a generic utility that reads the OS process table (`ps -Ao pid,ppid`, portable across Linux/macOS, `procps` already installed in the production Docker image) and walks the ppid graph to find all transitive descendants of a root pid.
- `killWorker()` now snapshots the PTY's descendant pids **before** killing the root (avoids a reparent race), sends `SIGTERM` to the root and all descendants, and escalates to `SIGKILL` for any descendant still alive after the existing PTY-exit timeout window (`PTY_EXIT_TIMEOUT_MS`, reused rather than adding a second timer).
- `listDescendantPidsImpl` / `signalPidsImpl` are added as constructor DI seams on `WorkerManager` (mirroring the existing `runAsUserImpl` pattern), defaulting to the real implementations.

## Root cause

Same-class prior art already exists in this codebase: the built-in `Bash` tool (`packages/embedded-agent/src/tools/bash.ts`) spawns with `detached: true` and kills via the negative pid (process group) — but that pattern only works when the caller directly spawns the leaf command. Here, the agent command is typed into an already-running **interactive** shell (job control ON), so we can't control its process group at spawn time; the fix instead walks the actual OS process tree at kill time.

## Test plan

- New `packages/server/src/lib/__tests__/process-tree.test.ts`: unit tests for the pure parser (`parseProcessTable`) and BFS (`collectDescendantPids`, including sibling-tree exclusion), plus a real-process reproduction test (via `setsid`) that:
  1. Reproduces the leak directly: signaling only the root pid leaves a same-session, different-process-group descendant alive.
  2. Confirms the fix: `listDescendantPids` + `signalPids` terminates that descendant.
  - TDD polarity verified: stubbing `collectDescendantPids` to always return `[]` fails the fix-polarity assertions while leaving the raw leak-reproduction assertion (independent of the fix) unaffected.
- Extended `packages/server/src/services/__tests__/worker-manager.test.ts`: asserts `killWorker` calls `listDescendantPidsImpl`/`signalPidsImpl` with the expected pids/signals (SIGTERM then SIGKILL), and that no signal call happens when there are no descendants. All 84 pre-existing tests in this file pass unmodified (mock PTYs use fake pids, so the real process-table walk is a harmless no-op in those tests).
- `bun run typecheck`: exit 0.
- `bun run test` (full suite): 1 pre-existing failure in `packages/server/src/services/__tests__/multi-user-mode.test.ts` (the elevation-skip direct-spawn path test around SSH_AUTH_SOCK forwarding), confirmed via baseline comparison (ran the identical test against `origin/main` HEAD `666cee8` in a throwaway worktree) — it fails identically there. This is a pre-existing environment leak (the dev sandbox's real `SSH_AUTH_SOCK` from 1Password leaking into `process.env`), unrelated to this change. All other tests pass (3422 pass / 1 skip in server, plus client/shared/embedded-agent/integration all green).
- `node .claude/skills/orchestrator/preflight-check.js`: all checks green (coverage, rule/skill duplication, language, blame-shift).
- CodeRabbit CLI could not authenticate in the headless delegated worktree (known limitation of this environment); relying on the GitHub-side CodeRabbit review once this PR is open.

## Follow-up (not in scope here)

The Issue suggests considering a new architectural invariant ("I-9: Spawned-Process Lifetime Follows Spawn Context") for `.claude/skills/architectural-invariants/SKILL.md`. Leaving that decision to the retro/brewing pass rather than bundling it into this bug fix.

Closes #1087

## Note on CodeRabbit
GitHub-side CodeRabbit bot hit its review-limit at PR open ("Next review available in: 55 minutes"). Local CodeRabbit CLI also could not run (auth timeout in this headless delegated worktree). After the stated window elapsed, `@coderabbitai review` was posted as a retrigger; the bot acknowledged ("Review triggered") but noted it does not re-review already-reviewed commits outside of paused auto-review, and no review/inline comments landed within the following ~30 minutes. Both CodeRabbit layers are unavailable for this PR at review time -- proceeding under the documented simultaneous-rate-limit disposition (coderabbit-ops troubleshooting.md); merge authority for this production-code change is left to the coordinating session.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worker shutdown to terminate descendant processes, including detached child processes.
  * Added escalation from `SIGTERM` to `SIGKILL` for processes that do not exit promptly.
  * Made process-tree handling resilient to malformed process data, missing processes, and cyclic relationships.

* **Tests**
  * Added comprehensive coverage for process discovery, descendant signaling, and worker cleanup behavior.


ms2sato:
@coderabbitai review

PR #1112, opened 2026-07-15
fix: reject empty-string instructions entries in embedded-agent init command

### LINKED ISSUE #1080: embedded-agent: init command schema `instructions` lacks minLength(1) validation
## Context

Wave 5-4 ([PR #1076](https://github.com/ms2sato/agent-console/pull/1076)) added `instructions?: string[]` to `EmbeddedAgentDefinition`. The wire schema for the field uses `InstructionsListSchema` (with `minLength(1)` on each entry) in most places.

## The gap

Architect audit noted that the `init` command schema (the message sent from server to embedded-agent subprocess at spawn) uses a **plain string array** for `instructions` rather than the `InstructionsListSchema` shape used elsewhere. Result: an empty-string entry could pass through the init command boundary without validation.

## Impact

**Non-blocking**. Per AC A11 (missing/unreadable non-fatal), an invalid entry would be skipped + warn-logged, not fatal. Empty-string entries would also fall through the \`resolveConfinedPath\` skip path.

## Fix

Replace the plain string-array schema in the init command with \`InstructionsListSchema\` to make the wire-boundary validation symmetric with all other consumers.

## Priority

Low. Non-blocker + no observable incorrect behavior; hardening / consistency candidate.

## Related

- [PR #1076](https://github.com/ms2sato/agent-console/pull/1076) (introduced the field)
- Issue #1072 (parent)

ms2sato:
## Summary

- The `init` command schema (server -> embedded-agent subprocess spawn message) validated `instructions` as a plain `v.array(v.string())`, unlike every other consumer of `InstructionsListSchema` (`EmbeddedAgentDefinitionSchema`, `CreateEmbeddedAgentRequestSchema`, `UpdateEmbeddedAgentRequestSchema`), which all use `minLength(1)` per entry.
- Replaced it with `InstructionsListSchema` so the wire-boundary validation is symmetric and an empty-string entry is rejected at the schema level, matching every other consumer.

## Preflight integration-test-gap disposition (joint decision with Orchestrator)

Preflight detected the change to `packages/shared/src/schemas/embedded-agent.ts` and strongly recommends an integration test, but this PR skips it for this sprint for the following reasons:

1. **No new field added** — only swaps the init command's `instructions` field from `v.array(v.string())` to the existing `InstructionsListSchema` (with `minLength(1)`). `EmbeddedAgentDefinitionSchema` and `Create/UpdateEmbeddedAgentRequestSchema` already use this same schema, so its integration surface is already exercised.
2. **No new runtime behavior path introduced** — validation-tightening only, covered by the existing schema-consumption paths.
3. **Unit test precisely verifies the boundary** — confirms an empty-string entry in the init command is rejected at the valibot schema-parse level.
4. Per `workflow.md`'s joint-decision protocol (delegate proposal + Orchestrator approval), this is skipped with the preflight warning left in place.

Q10 (Schema-Type Parallel Maintenance) is scoped to newly-added derived fields crossing the wire; this change does not add a field, so it is out of Q10's scope.

## Test plan

- [x] TDD: added a failing test (`rejects an init command with an empty-string entry in instructions`) confirmed to fail against pre-fix code, then pass after the fix
- [x] `bun run test` — shared (519), client (1900), embedded-agent (190), integration (63) all pass. Server: 3428 pass / 1 pre-existing unrelated fail (`MultiUserMode > Issue #918`, SSH_AUTH_SOCK env leak in test environment, unrelated package/file)
- [x] `bun run typecheck` — clean across all packages
- [x] `bun run check:lang` — clean
- [x] Preflight coverage check — production file covered by sibling test

Closes #1080

## Note on CodeRabbit
Both local CodeRabbit CLI (browser auth timeout in headless worktree) and GitHub-side bot (posted "Review failed" error, then a `@coderabbitai review` re-trigger returned "Review finished" with no actual findings/review body — the incremental-review system treated the failed attempt as already-reviewed) were unavailable for this PR's review window. No CodeRabbit feedback obtainable from either source. Merging under existing [PR Merge Authority](https://github.com/ms2sato/agent-console/blob/main/.claude/skills/orchestrator/SKILL.md#pr-merge-authority).


ms2sato:
@coderabbitai review

Issue #1150, opened 2026-07-16
test(flaky): stabilize resolve-targets.test.ts (observed intermittent failures in delegated sessions)


ms2sato:
## Symptom

`packages/server/src/services/inbound/__tests__/resolve-targets.test.ts` was observed failing intermittently across **3 separate delegated agent sessions** during Sprint 2026-07-16. Local re-runs on the Orchestrator host typically pass. The failure surfaces as delegate-side CI red or local `bun run test` red inside a fresh worktree, forcing the delegate to retry.

## Reproduction (approximate)

- Environment: fresh worktree (post `bun install`), running `bun run test` in a delegated session.
- Frequency: intermittent — not reliably reproducible on demand.
- Signal set: no shared `mock.module` registrations in this file itself (unlike the wider pattern tracked in [#791](https://github.com/ms2sato/agent-console/issues/791)); the flakiness cause is not yet identified.

## Candidate root cause hypotheses (unconfirmed)

1. **Ordering-sensitive `mock(() => ...)`**: several test cases wrap `getOrgRepoFromPath` with a per-case `mock(() => Promise.resolve(...))`. If a subsequent test relies on call-count state (`swallows GitError and continues processing remaining sessions` uses `callCount++` inside the mock), state leakage from an earlier test in the same process could shift the branch.
2. **`new Date().toISOString()` timestamp non-determinism**: `createEvent()` builds events with `Date.now()`-based timestamps. If any assertion path branches on timestamp equality/ordering, it could flip.
3. **Repository lookup helper (`buildPersistedRepository`)** may share static fixture state across tests in the process.
4. **Sibling file interaction**: `diff-worker-handler.test.ts` in the same directory uses `mock.module` — process-global registration order may affect this file's imports.

## Acceptance Criteria

- Reproduce the intermittent failure deterministically (e.g. by running a targeted sequence multiple times, or by identifying the interfering sibling test).
- Once reproduced, fix at the root: prefer removing shared state / ordering dependence rather than adding retries.
- After the fix, `bun run test` should pass 20/20 consecutive runs on both local and a fresh delegated worktree.
- Add a regression note in the test file if the fix targets a subtle ordering assumption.

## Why

Repeated delegate-side failures burn round-trips and mask genuine test regressions. The flakiness was observed 3 times in Sprint 2026-07-16 across different delegates, so it is not a one-off — it is a systematic property of the test file that needs stabilization before it becomes a routine ignore.

## References

[...]

PR #1173, opened 2026-07-17
feat(preflight): accept .ts sibling for pure-logic .tsx source (#1049)

### LINKED ISSUE #1049: preflight: accept .ts sibling test for .tsx source when it is pure-logic-only
Follow-up surfaced during Sprint 2026-07-11/12 PR [#1048](https://github.com/ms2sato/agent-console/pull/1048) (MessagePanel D 案 reuse).

## Symptom

preflight-check.js flagged `packages/client/src/components/sessions/SessionPage.tsx` as "missing tests" even though **two sibling test files already exist**:

- `packages/client/src/components/sessions/__tests__/SessionPage.test.ts` (`.ts` extension, JSX-free pure-logic test — exercises `workerRestart` module integration)
- `packages/client/src/components/sessions/__tests__/SessionPage.keyboard.test.tsx` (`.tsx` extension, keyboard-specific integration test)

## Root cause

`check-utils.js`'s sibling detection expects `<basename>.test.<same-source-ext>` strict match. For a `SessionPage.tsx` source:

- **Expected** by preflight: `__tests__/SessionPage.test.tsx`
- **Not accepted**: `SessionPage.test.ts` (extension mismatch) or `SessionPage.keyboard.test.tsx` (basename suffix `.keyboard`)

This is a **legitimate blind spot**: a React component may have a `.tsx` source but a JSX-free logic test that lives naturally as `.ts` (no runtime need for the JSX-enabling extension). Forcing `.tsx` extension on such tests is convention friction without functional benefit.

The current work-around applied in PR #1048 was to rename `SessionPage.test.ts` → `SessionPage.test.tsx` so preflight would accept it. This is harmless but the check should not require the workaround.

## Proposed fix

Extend the sibling detection logic in `check-utils.js` to:

1. **Accept `.ts` sibling test for `.tsx` source** when the file exists and is a valid test (contains at least one `describe(` / `test(` / `it(`). This mirrors the actual bun/vitest runtime behavior — both `.ts` and `.tsx` tests are collected identically.
2. **Consider whether to accept variant suffixes** like `SessionPage.keyboard.test.tsx` or `SessionPage.integration.test.tsx` as sibling tests. This is a broader design decision — currently `.keyboard.test.tsx` is not the canonical sibling shape, and forcing an explicit sibling has value (clear "primary" test file). Discuss in the PR.

## Not in scope

- Test framework changes
- Modifying the `__tests__/` directory convention

## Change surface

Low — `.claude/skills/orchestrator/check-utils.js` (sibling detection helper) + `.claude/rules/test-trigger.md` mirror update (per the mirror-drift check) + tests for the extended pattern.

## Acceptance criteria

- Preflight recognises `SessionPage.test.ts` (or equivalent `.ts` sibling for `.tsx` source) as valid sibling test coverage
- `check-mirror-drift.js` passes after the `test-trigger.md` mirror update
- Existing preflight-check.js unit tests continue to pass; a new unit test covers the extension case
- The workaround applied in PR #1048 (rename `.ts` → `.tsx`) can be reverted in a follow-up cleanup PR (optional, low-priority)

## Cross-reference

- Surfaced by: PR #1048 (embedded-agent: reuse MessagePanel in EmbeddedAgentWorkerView)
- Related: umbrella #1004
- Related file: `.claude/rules/test-trigger.md` (mirror maintenance)

ms2sato:
## Summary

Issue #1049 diagnosed the root cause as `check-utils.js`'s sibling-test detection requiring a strict `<basename>.test.<same-source-ext>` match, rejecting a `.ts` sibling test for a `.tsx` source. **This diagnosis does not match the current code**: `findTestFiles()`'s `hasTest` check compares basenames only and never inspects extension, so a `.tsx` source's `.ts` sibling test is already recognized as coverage once that test file is part of the PR's diff.

The actual friction PR #1048 hit (documented in its own rename-commit message, `2cdfafe`) was different: preflight only credits coverage from files present in the PR diff, not files that merely exist on disk. That is the intended behavior of the "sibling test must be touched in the same PR" policy (`.claude/rules/pre-pr-completeness.md` Q3, from PR #764) and is **kept unchanged** in this PR — relaxing it would be a separate, more consequential policy change out of this Issue's scope. This was discussed with the Orchestrator before implementation; see the investigation note below.

## What this PR does (Q3 policy preserved)

1. **Locks in the already-correct extension-agnostic matching** with new positive/negative unit tests in `check-utils.test.js` (`.tsx` source + `.ts` sibling in `__tests__/`, same-dir sibling, no-sibling gap, cross-directory non-match, and the real `SessionPage.tsx`/`SessionPage.test.ts` pairing that motivated the Issue).
2. **Improves the "Missing Tests" message**: `findTestFiles()` now also returns `alternateTestPath` (e.g. `.../Bar.test.ts` alongside the primary `.../Bar.test.tsx` suggestion) for `.tsx` sources, and `preflight-check.js` surfaces it as "(or `.../Bar.test.ts` if JSX-free)" — so a future author facing this gap isn't misled into thinking only `.tsx` is acceptable, and doesn't need PR #1048's rename workaround to understand why.
3. **Updates `.claude/rules/test-trigger.md`** to document that JSX-free `.tsx`-source tests may live as `.ts`. Verified with `check-mirror-drift.js` (only the "File Pattern" column is drift-checked; the "Expected Test Location" prose column edited here is not).
4. **Reverts PR #1048's `SessionPage.test.tsx` → `.ts` workaround rename** back to `.ts` (no content change — content has no JSX, confirmed by grep). Safe in this PR because `SessionPage.tsx` itself is not modified here, so the coverage-check trigger doesn't apply to it in this diff.

## Investigation note (sent to Orchestrator before implementing)

Verified via a direct `findTestFiles()` call that `hasTest: true` already holds when a `.tsx` source's `.ts` sibling test is included in the diff, and `hasTest: false` when the sibling exists only on disk but isn't in the diff — matching the rename-commit's own stated reason, not an extension mismatch. Orchestrator approved: keep Q3 policy, land the 4 items above as a Sprint-scoped clean fix, and correct the Issue's root-cause claim in this PR body (this section).

## Test plan

- [x] `bun test ./.claude/skills/orchestrator/__tests__/` — 232 pass, 0 fail (new tests included)
- [x] `bun run test:scripts` — 434 pass, 0 fail
- [x] `node .claude/skills/orchestrator/preflight-check.js` — clean, exit 0
- [x] `node .claude/skills/orchestrator/check-mirror-drift.js` — in sync, exit 0
- [x] `bun run check:lang` — clean (111 files scanned)
- [x] `bun run typecheck` — exit 0
- [x] `bun run test` (full suite, run twice) — exit 1 both times, **solely** due to `packages/server/src/services/inbound/__tests__/resolve-targets.test.ts`, a pre-existing flaky test tracked in #1150 (being fixed by a parallel delegate this sprint). Confirmed unrelated to this PR: `packages/server/` is untouched by this diff, and the test passes 3/3 when run in isolation (`bun test src/services/inbound/__tests__/resolve-targets.test.ts`). All other packages (client 1949, shared 527, integration 70, embedded-agent 217) passed 0-fail both runs.

Closes #1049


ms2sato:
## Note on CodeRabbit

Local CodeRabbit CLI cannot authenticate in this headless delegated worktree (`coderabbit review --agent` attempts a browser OAuth flow and times out: "Automatic login timed out. Use the printed fallback URL to finish authentication."). Relying on the GitHub-side CodeRabbit bot review for the review layer per the documented fallback in `coderabbit-ops` troubleshooting.

PR #1186, opened 2026-07-18
fix: read live dirty state to avoid StrictMode draft-restore race

### LINKED ISSUE #1077: CreateWorktreeForm draft-restore effect races with StrictMode double-invoke (dev-only visibility)

ms2sato:
## Summary

`CreateWorktreeForm`'s draft-restore-on-mount effect intermittently lost the just-restored localStorage draft under React 18 `<StrictMode>` (dev only). Fixed by reading react-hook-form's live internal dirty state instead of a render-synced ref.

Closes #1077

## Root cause

`CreateWorktreeForm` has two effects:

1. **Draft-restore effect** — on mount, reads the localStorage draft and calls `reset()` / `setValue()` to hydrate the form.
2. **Save-debounce effect** — subscribes to `watch()`, debounce-saves dirty fields to localStorage on change, and on **cleanup** (real or StrictMode-simulated unmount) unconditionally writes the current dirty fields back to localStorage.

`pickDirtyValues` filtered fields using `dirtyFieldsRef.current`, a `useRef` kept in sync with `formState.dirtyFields` **only during render**.

React 18 StrictMode (dev only, via `main.tsx`'s `<StrictMode>` wrap) double-invokes effects synchronously right after mount: run effects → run all cleanups (simulated unmount) → run effects again, **with no React render in between**. So when the save effect's cleanup ran during the simulated unmount, `dirtyFieldsRef.current` still held its stale pre-restore value (empty), even though the restore effect's `reset()`/`setValue()` had already synchronously mutated react-hook-form's *internal* dirty state. The cleanup wrote `pickDirtyValues(getValues())` — which read the stale ref and produced `{}` — clobbering the just-restored draft in localStorage.

This reproduced for both `register()`-bound fields (`sessionTitle`) and `setValue`-only fields (`embeddedAgentId`) — a general race in the draft-persistence effect pair, not specific to one field. Confirmed NOT a production issue: React's production runtime never double-invokes effects, so real users are unaffected.

## Fix strategy

Stop relying on the render-synced `dirtyFieldsRef`. `pickDirtyValues` now reads react-hook-form's **live internal** dirty state directly via `control._formState.dirtyFields` (typed as `Control<TFieldValues>._formState: FormState<TFieldValues>`, no `as any` needed). This internal state is mutated synchronously by `reset()`/`setValue()` regardless of whether a React render/commit has happened, so it is never stale — in StrictMode or out of it, in dev or in prod. This works identically in both environments because it removes the actual race rather than special-casing StrictMode.

`dirtyFieldsRef` and its render-body sync line were deleted; `formState`'s destructured `dirtyFields` was also removed (no longer used).

## Test

Added `describe('StrictMode double-invoke draft-restore race (Issue #1077)')` in `CreateWorktreeForm.test.tsx`, using a `renderCreateWorktreeFormStrict` helper (mirrors the existing `EmbeddedAgentWorkerView.test.tsx` `renderViewStrict` convention) that wraps the component tree in `<StrictMode>`, since a non-StrictMode render cannot reproduce this race.

The regression-guarding assertion checks `localStorage.getItem(draftKey)` **after mount**, not just the rendered field value — the bug's actual effect is that the in-memory form value ends up looking fine (RHF's internal state isn't reverted) while the on-disk localStorage draft is silently wiped to `{}`, so a *subsequent* page reload loses the draft. An assertion on the rendered value would not polarity-flip; the localStorage assertion does.

**Deviation from a register()-bound field:** the reproduction field is `embeddedAgentId` (restored via `setValue(..., { shouldDirty: true })`) rather than `sessionTitle` (restored via `reset()`). Investigation found restoring a `register()`-bound field via this form's `reset(merged, { keepDefaultValues: true })` does not actually land in react-hook-form's internal `_formValues` when the form also has `shouldUnregister: true` — a separate, pre-existing RHF interaction reproducible with or without this fix and without StrictMode at all. This makes register()-bound fields unusable as a regression guard for *this* bug (their localStorage value never becomes the restored value either way). This is documented in the test file as a code comment and independently reproduced live in Browser QA below (see the "Note on a separate finding" section). It is out of scope for #1077; flagging as a candidate follow-up.

### Polarity check

1. With test + fix: **PASS**.
2. Reverted only the production fix (kept the new test) — re-ran: **FAIL**:
   ```
   expect(parsed.embeddedAgentId).toBe('embedded-1');
   Expected: "embedded-1"
   Received: undefined
   ```
   (i.e. the saved draft was `{}` — the exact clobbering behavior this PR fixes.)
3. Restored the fix — re-ran: **PASS** again.

The assertion that flips is `expect(parsed.embeddedAgentId).toBe('embedded-1')`.

## Verification

- `bun run typecheck`: exit 0, all packages clean.
- `bun run test` (full monorepo): 6469 pass / 1 skip / 1 fail. The sole failure (`packages/server/src/services/__tests__/multi-user-mode.test.ts:1107`, an elevation-skip (direct) path test) is a pre-existing environment-isolation artifact — my shell's real `SSH_AUTH_SOCK` leaks into the test process env, unrelated to this diff (zero changes to `packages/server/`). Confirmed by re-running with `SSH_AUTH_SOCK` unset: all 37 tests in that file pass.
- `packages/client` test suite alone (includes `CreateWorktreeForm.test.tsx`): 2064 pass / 0 fail.
- `node .claude/skills/orchestrator/preflight-check.js`: clean — coverage OK (sibling test modified), no rule/skill duplication, no language-policy violations, no source-comment blame-shift violations.
- Duplication grep (`dirtyFieldsRef|_formState.dirtyFields` across `packages/client/src`): only 2 hits, both in `CreateWorktreeForm.tsx` — no sibling component needs the same fix.

## Browser QA (StrictMode dev build)

Verified live under `bun run dev` (StrictMode-wrapped) via the "Create worktree" dialog (`QuickWorktreeDialog`, the only draftKey-using entry point), which mounts `CreateWorktreeForm` fresh with an existing localStorage draft on each dialog open.

**Before fix** (production code reverted to pre-fix state, draft pre-seeded with `embeddedAgentId`, page reloaded, dialog opened — genuine fresh StrictMode mount): Agent selector shows the default **"Claude Code (built-in)"**, not the restored `deepseek-v4-flash` — draft was clobbered to `{}` in localStorage.

**After fix** (same steps, fix restored): Agent selector correctly shows **"deepseek-v4-flash"**, restored from the draft — localStorage still holds `{"embeddedAgentId":"..."}`.

| Before fix (draft clobbered) | After fix (draft restored) |
|---|---|
| Agent shows default "Claude Code (built-in)" | Agent shows restored "deepseek-v4-flash" |

(Screenshots posted as a follow-up PR comment via `upload-qa-screenshots.sh`.)

**Note on a separate finding:** while setting up this QA, I independently reproduced — via manual browser interaction, not just the unit test — that a `register()`-bound field (`sessionTitle`) does NOT restore from a localStorage draft at all, regardless of this fix, regardless of StrictMode. This corroborates the pre-existing RHF interaction noted in the Test section above. Not fixed in this PR (out of scope for #1077); flagging as a candidate follow-up issue if the maintainers agree it's worth fixing.

## Test plan

- [x] `bun run typecheck` passes
- [x] `bun run test` passes (excluding pre-existing, unrelated `SSH_AUTH_SOCK` env-leak failure)
- [x] `preflight-check.js` clean
- [x] New regression test added, polarity-checked
- [x] Browser QA under StrictMode dev build, before/after screenshots
- [x] CodeRabbit CLI review addressed

## Note on CodeRabbit

Local CodeRabbit CLI is not authenticated in this headless delegated worktree session (`coderabbit auth status` reports signed out; interactive OAuth login is not available here). Relying on the GitHub-side CodeRabbit bot review per the documented rate-limit/unavailable-CLI fallback. Will confirm `reviewDecision` is APPROVED (or all inline comments addressed) before requesting merge.

### DIFF
--- a/packages/client/src/components/worktrees/CreateWorktreeForm.tsx
+++ b/packages/client/src/components/worktrees/CreateWorktreeForm.tsx
@@ -72,7 +72,8 @@ export function CreateWorktreeForm({
     watch,
     setValue,
     reset,
-    formState: { errors, dirtyFields },
+    control,
+    formState: { errors },
   } = useForm<CreateWorktreeFormData>({
     resolver: valibotResolver(CreateWorktreeFormSchema),
     defaultValues: {
@@ -93,14 +94,17 @@ export function CreateWorktreeForm({
   // Tracks whether draft was cleared (e.g., after successful submit) to skip unmount save
   const draftClearedRef = useRef(false);
 
-  // Ref to access dirtyFields inside effects without re-triggering them
-  const dirtyFieldsRef = useRef(dirtyFields);
-  dirtyFieldsRef.current = dirtyFields;
-
   /** Pick only user-modified fields from form values, so props-derived defaults
-   *  (e.g., agentId from server) are not persisted and can update between sessions. */
+   *  (e.g., agentId from server) are not persisted and can update between sessions.
+   *  Reads react-hook-form's live internal dirty state (control._formState.dirtyFields)
+   *  rather than a render-synced ref: reset()/setValue() mutate this internal state
+   *  synchronously, without waiting for a React render/commit. A render-synced ref
+   *  would go stale under React 18 StrictMode's dev-only double-invoke (mount ->
+   *  simulated unmount -> mount again with no render in between), causing the
+   *  simulated-unmount cleanup to read an empty ref and clobber a just-restored
+   *  draft with `{}`. */
   const pickDirtyValues = useCallback((values: Record<string, unknown>): Record<string, unknown> => {
-    const dirty = dirtyFieldsRef.current;
+    const dirty = control._formState.dirtyFields;
     const result: Record<string, unknown> = {};
     for (const key of Object.keys(values)) {
       if (dirty[key as keyof typeof dirty]) {
@@ -108,7 +112,8 @@ export function CreateWorktreeForm({
       }
     }
     return result;
-  }, []);
+    // control is a stable ref from react-hook-form
+  }, [control]);
 
   // Restore draft on mount (skipped when prefillValues is provided)
   useEffect(() => {
--- a/packages/client/src/components/worktrees/__tests__/CreateWorktreeForm.test.tsx
+++ b/packages/client/src/components/worktrees/__tests__/CreateWorktreeForm.test.tsx
@@ -1,4 +1,5 @@
 import { describe, it, expect, mock, beforeEach, afterEach, afterAll } from 'bun:test';
+import { StrictMode } from 'react';
 import { render, screen, waitFor, act, cleanup } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
@@ -69,6 +70,38 @@ function renderCreateWorktreeForm(props: Partial<React.ComponentProps<typeof Cre
       <TestWrapper>
         <CreateWorktreeForm {...mergedProps} />
       </TestWrapper>
+    ),
+    props: mergedProps,
+  };
+}
+
+/**
+ * Same as `renderCreateWorktreeForm`, but wrapped in `React.StrictMode`, matching
+ * production (`main.tsx` wraps the entire app in `<StrictMode>`). This distinction
+ * matters for the draft-restore race regression test below (Issue #1077):
+ * StrictMode's dev-mode double-invoke of effects (mount -> simulated unmount ->
+ * mount again, with no React render in between) is what actually reproduces the
+ * real-browser bug where a stale render-synced dirtyFields ref caused the
+ * simulated-unmount cleanup to clobber a just-restored draft with `{}`.
+ * `renderCreateWorktreeForm` (no StrictMode) cannot reproduce that failure mode.
+ */
+function renderCreateWorktreeFormStrict(props: Partial<React.ComponentProps<typeof CreateWorktreeForm>> = {}) {
+  const defaultProps = {
+    repositoryId: 'repo-1',
+    defaultBranch: 'main',
+    onSubmit: mock(() => Promise.resolve()),
+    onCancel: mock(() => {}),
+  };
+
+  const mergedProps = { ...defaultProps, ...props };
+
+  return {
+    ...render(
+      <StrictMode>
+        <TestWrapper>
+          <CreateWorktreeForm {...mergedProps} />
+        </TestWrapper>
+      </StrictMode>
     ),
     props: mergedProps,
   };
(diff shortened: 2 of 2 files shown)


ms2sato:
## QA Screenshots (2026-07-18 11:33)

### 04 strictmode fresh mount draft restored
04 strictmode fresh mount draft restored

<details>

04 strictmode fresh mount draft restored
</details>

### 05 before fix draft clobbered
05 before fix draft clobbered

<details>

05 before fix draft clobbered
</details>

### 06 after fix draft restored
06 after fix draft restored

<details>

06 after fix draft restored
</details>

PR #1193, opened 2026-07-18
fix: restore register()-bound draft fields via setValue under shouldUnregister:true (#1187)

### LINKED ISSUE #1187: bug: CreateWorktreeForm sessionTitle draft-restore fails under `shouldUnregister: true` + `reset()` combination
Discovered during Browser QA for #1077 (StrictMode double-invoke fix) as an independent, pre-existing bug.

## Symptom

For `register()`-bound fields in `CreateWorktreeForm` (specifically `sessionTitle`), a saved localStorage draft is **not** restored on remount, even after the #1077 fix. Under `bun run dev` (StrictMode), reloading the page after saving a draft leaves the `sessionTitle` field empty rather than restoring the saved value.

## Scope note (not #1077)

This is **not caused by** and **not resolved by** #1077. #1077 fixes the `dirtyFieldsRef` stale-read that was clobbering the draft to `{}`. The `sessionTitle` bug reproduces independently:

- With #1077 fix applied AND draft preserved as full object in localStorage → `sessionTitle` still empty on remount
- The same PR's `embeddedAgentId` field (setValue-based, non-`register()`) restores correctly

## Suspected root cause

Interaction between:
- `register()`-bound fields (which enter RHF's fields registry on mount)
- `shouldUnregister: true` in the `useForm` config (fields drop out of the registry on unmount)
- `reset(mergedDraft, { keepDefaultValues: true })` on the initial draft-restore effect (writes to the registry, but the field may not be registered yet at that moment)

When the field's `register()` runs (on the input's ref callback) after `reset()` fires, the value written to the registry gets overwritten by the field's initial default. #1077's fix does not touch this ordering.

## Reproduction

1. Open `CreateWorktreeForm` under `bun run dev` (StrictMode)
2. Type a value into `sessionTitle`
3. Wait for debounced save (draft in `localStorage[<form-key>]`)
4. Reload the page
5. Observed: `sessionTitle` field is empty
6. Expected: `sessionTitle` field shows the saved value

## Evidence (from #1077 PR #1186)

- Browser QA screenshots in [PR #1186](https://github.com/ms2sato/agent-console/pull/1186) body demonstrate `embeddedAgentId` restores but `sessionTitle` does not
- Independent frontend-specialist unit test explored the same order-of-operations issue

## Fix candidates (not decided)

- Trigger `setValue(name, draftValue)` for `register()`-bound fields inside a `useEffect` that runs after mount (post-register order)
- Set `shouldUnregister: false` if the register/unregister churn is not needed
- Explicit `field.onChange(draftValue)` after `register()` completes

## Related

- [PR #1186](https://github.com/ms2sato/agent-console/pull/1186) (Issue #1077, StrictMode fix) — orthogonal, closes different root cause
- Sprint 2026-07-18 architect audit for PR #1186 explicitly recommended filing this Issue rather than leaving it as PR body / test comment (rot risk)

## Priority

Medium — user-visible draft loss for a `register()`-bound field, but scoped to one form and workaround exists (retype). Not blocking for #1077 merge.

ms2sato:
Closes #1187

## Root cause

`CreateWorktreeForm` restores a saved localStorage draft on mount via:

```ts
reset({ ...getValues(), ...draft }, { keepDefaultValues: true });
```

The form is configured with `shouldUnregister: true`. For fields wired via `{...register('fieldName')}` (e.g. `sessionTitle`, `initialPrompt`, `customBranch`, `baseBranch`, `branchNameMode`), `reset(merged, { keepDefaultValues: true })` does **not** reliably populate react-hook-form's internal `_formValues` for those fields when `shouldUnregister: true` is set. The restored value is silently dropped and the field falls back to its original default.

This is reproducible **independent of React StrictMode** — a pre-existing comment in `CreateWorktreeForm.test.tsx` (added while investigating PR #1186 / Issue #1077) already documented this exact gap and explained why that regression test uses `embeddedAgentId` (restored via explicit `setValue`, not `reset()`) rather than `sessionTitle` as its guard field, precisely because `sessionTitle` never became the restored value either way and would not polarity-flip.

The Issue's "ref callback runs after `reset()`" ordering theory was a plausible but essentially unverified hypothesis. The empirical finding (confirmed by a new failing test against unfixed code, and by the pre-existing test-file comment) is that the gap is an interaction between `register()`-bound fields, `reset(..., { keepDefaultValues: true })`, and `shouldUnregister: true` — not a StrictMode timing race.

## Fix strategy

Extended the mount-restore effect so that **every key present in the draft object** is restored via explicit `setValue(key, value, { shouldDirty: true })` — the same mechanism already proven correct for `agentId` / `embeddedAgentId`. The `reset()` call and its two special-cased `if ('agentId' in draft)` / `if ('embeddedAgentId' in draft)` branches were removed since the generic `setValue` loop now covers all draft keys uniformly, including those two.

```ts
for (const key of Object.keys(draft) as Array<keyof CreateWorktreeFormData>) {
  setValue(key, draft[key] as never, { shouldDirty: true });
}
```

### Rejected alternatives (from the Issue's fix candidates)

- **`shouldUnregister: false`** — rejected because it has broader implications (field registry no longer churns across conditional renders, e.g. the `customBranch` / `baseBranch` fields that only render in certain `branchNameMode` states), and the `setValue`-based fix is sufficient without touching that behavior.
- **`field.onChange(draftValue)` after `register()` completes** — rejected because it requires per-field wiring at each JSX callsite (more surface area, easier to miss a field) versus the single generic loop in the existing mount effect.

## Test

Added `describe('register()-bound field restore (Issue #1187)')` in `CreateWorktreeForm.test.tsx`: sets `sessionTitle` in a localStorage draft, renders the form (non-StrictMode, since the pre-existing comment confirms this bug reproduces without StrictMode), waits for agents to load, and asserts the rendered `sessionTitle` input's value equals the draft value.

**Polarity check:** `git diff` the production fix into a patch, reverted only `CreateWorktreeForm.tsx` to the pre-fix state while keeping the new test, and ran the suite:
- **Without the fix:** 39 pass / 1 fail — the new assertion `expect(titleInput.value).toBe('Restored Title')` fails (input value stays empty, the field's default).
- **With the fix restored:** 40 pass / 0 fail — the assertion passes.

The `#1077` fix (`control._formState.dirtyFields` live read in `pickDirtyValues`) is unchanged and its regression test (`StrictMode double-invoke draft-restore race (Issue #1077)`, `embeddedAgentId`-based) still passes.

## Verification

- `bun run typecheck`: exit 0
- `bun run test` (full monorepo suite): all packages pass. One pre-existing, unrelated failure was observed and root-caused to a local dev-environment `SSH_AUTH_SOCK` leak into `process.env` (a 1Password agent socket path) affecting an unrelated `MultiUserMode` test in `packages/server`; re-running with `env -u SSH_AUTH_SOCK` gives a fully clean run (2065+573+72+3528+257+2065... all packages 0 fail, exit 0). Two other transient failures (`terminal-store-paging.test.ts`, `terminal-store.test.ts`) were observed once under full-suite parallel load and confirmed as pre-existing flakes unrelated to this change — both pass in isolation and did not recur on a repeat full-suite run.
- `node .claude/skills/orchestrator/preflight-check.js`: exit 0 (coverage check covered, blame-shift check clean, language check clean).
- CodeRabbit CLI is installed locally but requires interactive browser auth, unavailable in this headless worktree; the CLI review step was skipped. GitHub-side CodeRabbit review will run on this PR as usual.

## Browser QA (StrictMode, `bun run dev`)

Verified end-to-end in a real browser against the running dev server (React `StrictMode` is always on in `main.tsx`, dev or prod), using the global "Create worktree" dialog (`QuickWorktreeDialog`, which passes `draftKey`):

1. **Fresh mount (no draft)** — `Title (optional)` field empty, as expected.
2. **Before fix** — temporarily checked out the pre-fix version of `CreateWorktreeForm.tsx` only (via `git checkout <parent-commit> -- <file>`, uncommitted, reverted immediately after; no commit history was altered) with a `sessionTitle` draft already saved in `localStorage`. Reloaded and reopened the dialog: `Title` field renders **empty** despite the saved draft — reproduces the reported bug.
3. **After fix** — restored the fixed file (`git checkout HEAD -- <file>`), reseeded the same draft, reloaded, reopened the dialog: `Title` field renders **"QA Draft Restore Test 1187"**, matching the saved draft.

Screenshots posted in https://github.com/ms2sato/agent-console/pull/1193#issuecomment-5010018010 (fresh mount / before-fix bug / after-fix restored, in that order).

Working tree was confirmed clean (`git status --porcelain` empty, `git diff HEAD` empty) after each revert step, before and after the temporary before/after comparison.

## Note on CodeRabbit

**Correction (per architect verdict, Sprint 2026-07-18)**: GitHub-side CodeRabbit bot did complete a review of this PR — walkthrough posted, and a manual `@coderabbitai review` re-trigger returned "Review finished" (indicating no un-reviewed commits remain, not a rate-limit block). Inline actionable comments = 0. `reviewDecision` remains empty per this repo's status-check-only CodeRabbit pattern (matches PR #1183, #1191, #1194 precedent). Local CodeRabbit CLI required interactive browser auth unavailable in this headless worktree; the GitHub-side review is the authoritative signal here.

Merging under existing [PR Merge Authority](https://github.com/ms2sato/agent-console/blob/main/.claude/skills/orchestrator/SKILL.md#pr-merge-authority) with architect audit + Orchestrator gate satisfied.

### DIFF
--- a/packages/client/src/components/worktrees/CreateWorktreeForm.tsx
+++ b/packages/client/src/components/worktrees/CreateWorktreeForm.tsx
@@ -127,8 +127,8 @@ export function CreateWorktreeForm({
         // shouldUnregister: true, reset(merged, { keepDefaultValues: true })
         // does not reliably populate react-hook-form's internal _formValues
         // for register()-bound fields (e.g. sessionTitle) -- reproducible
-        // independent of React StrictMode (Issue #1187). setValue is the
-        // mechanism already proven correct here for the non-register()-bound
+        // independent of React StrictMode. setValue is the mechanism already
+        // proven correct here for the non-register()-bound
         // agentId/embeddedAgentId fields, so it is used uniformly for every
         // draft key.
         for (const key of Object.keys(draft) as Array<keyof CreateWorktreeFormData>) {
--- a/packages/client/src/components/worktrees/CreateWorktreeForm.tsx
+++ b/packages/client/src/components/worktrees/CreateWorktreeForm.tsx
@@ -71,7 +71,6 @@ export function CreateWorktreeForm({
     setError,
     watch,
     setValue,
-    reset,
     control,
     formState: { errors },
   } = useForm<CreateWorktreeFormData>({
@@ -124,23 +123,23 @@ export function CreateWorktreeForm({
       const saved = localStorage.getItem(draftKey);
       if (saved) {
         const draft = JSON.parse(saved) as Partial<CreateWorktreeFormData>;
-        // getValues() returns initial defaults (form just mounted).
-        // The draft overwrites them with previously saved user input.
-        reset({ ...getValues(), ...draft }, { keepDefaultValues: true });
-        // agentId/embeddedAgentId are never register()-bound (custom select, driven by
-        // watch/setValue), so reset()'s keepDefaultValues silently discards them; restore explicitly.
-        if ('agentId' in draft) {
-          setValue('agentId', draft.agentId, { shouldDirty: true });
-        }
-        if ('embeddedAgentId' in draft) {
-          setValue('embeddedAgentId', draft.embeddedAgentId, { shouldDirty: true });
+        // Restore every saved key via setValue rather than reset(): with
+        // shouldUnregister: true, reset(merged, { keepDefaultValues: true })
+        // does not reliably populate react-hook-form's internal _formValues
+        // for register()-bound fields (e.g. sessionTitle) -- reproducible
+        // independent of React StrictMode (Issue #1187). setValue is the
+        // mechanism already proven correct here for the non-register()-bound
+        // agentId/embeddedAgentId fields, so it is used uniformly for every
+        // draft key.
+        for (const key of Object.keys(draft) as Array<keyof CreateWorktreeFormData>) {
+          setValue(key, draft[key] as never, { shouldDirty: true });
         }
       }
     } catch {
       // Ignore corrupted drafts
     }
-    // reset, getValues and setValue are stable refs from react-hook-form
-  }, [draftKey, reset, getValues, setValue, prefillValues]);
+    // setValue is a stable ref from react-hook-form
+  }, [draftKey, setValue, prefillValues]);
 
 
 
--- a/packages/client/src/components/worktrees/__tests__/CreateWorktreeForm.test.tsx
+++ b/packages/client/src/components/worktrees/__tests__/CreateWorktreeForm.test.tsx
@@ -900,6 +900,36 @@ describe('CreateWorktreeForm', () => {
       expect(localStorage.getItem('undefined')).toBeNull();
       expect(localStorage.getItem('null')).toBeNull();
     });
+
+    describe('register()-bound field restore (Issue #1187)', () => {
+      const draftKey = 'register-bound-restore-draft-key';
+
+      beforeEach(() => {
+        localStorage.removeItem(draftKey);
+      });
+
+      afterEach(() => {
+        localStorage.removeItem(draftKey);
+      });
+
+      it('should restore a register()-bound sessionTitle value from localStorage on mount', async () => {
+        localStorage.setItem(draftKey, JSON.stringify({ sessionTitle: 'Restored Title' }));
+
+        renderCreateWorktreeForm({ draftKey });
+
+        // Wait for agents to load
+        await waitFor(() => {
+          expect(screen.getByText('Claude Code (built-in)')).toBeTruthy();
+        });
+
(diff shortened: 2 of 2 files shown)


ms2sato:
## QA Screenshots (2026-07-18 14:10)

### 1 fresh mount
1 fresh mount

<details>

1 fresh mount
</details>

### 2 before fix bug
2 before fix bug

<details>

2 before fix bug
</details>

### 3 after fix restored
3 after fix restored

<details>

3 after fix restored
</details>

ms2sato:
@coderabbitai review

Issue #1162, opened 2026-07-17
embedded-agent preview sanitizer: known mXSS depth gap (cure53 adoption-agency shape) — concrete target for tokenizer/allowlist redesign


ms2sato:
## Context

Filed as a follow-up during [#1106](https://github.com/ms2sato/agent-console/issues/1106) implementation, per architect direction (session 84a3a530, 2026-07-17). Records a concrete Chromium mXSS vector that survives the current `sanitizePreviewFragment` and re-parse cycle, so the deferred tokenizer / allowlist redesign has a concrete example to solve.

## Vector

```html
<form><math><mtext></form><form><mglyph><style></math><img src=1 onerror=alert(1)>
```

Classic cure53 adoption-agency-triggering shape. Empirically verified in Chromium via Chrome DevTools MCP (during #1106 corpus construction):

- **First DOMParser pass** (sanitize): `<img onerror>` is trapped as inert TEXT inside a `<style>` RAWTEXT element due to the double-`<form>` triggering the adoption-agency algorithm's foreign-content-breakout quirk. `<style>` inside `<mglyph>` re-enters HTML parsing rules, but `<style>`'s own content model is RAWTEXT, so the sanitizer's `querySelectorAll` never sees a real `<img>` element to strip `onerror` from.
- **Re-parse** (iframe blob: URL load): the same string, when parsed by the browser as the iframe's content, produces a REAL `<img>` element with a REAL `onerror` attribute.
- Confirmed to fire in a sandbox-DISABLED diagnostic iframe (`scriptExecutedWithoutSandbox: true`).

**Narrower variants tested and did NOT reproduce**:
- Plain `<math><mglyph><style>...`
- `<math><malignmark><style>...`
- `<svg><foreignObject><style>...`
- `<svg><title><style>...`

The gap is specific to this exact adoption-agency-triggering shape.

## Not a direct vulnerability (per architect risk model)

Production preview iframe uses `sandbox=""` (no `allow-scripts` token) + meta CSP. Both are engine-independent hard barriers enforced by the browser itself:

- `sandbox=""` → opaque origin, cross-origin-blocked
- meta CSP → script execution disabled

The mXSS gap is **depth erosion** (sanitizer layer 3 of the 3-layer defense), not a direct vulnerability. The sandbox+CSP layers 1 and 2 fully contain the mutated `<img onerror>`. Confirmed empirically: in the production sandbox configuration, this vector does not execute.

## Why record it

[...]

Issue #1106, opened 2026-07-14
embedded-agent preview: Safari/Firefox defense-in-depth sanitizer (v1.1)


ms2sato:
## Motivation

[PR #1105](https://github.com/ms2sato/agent-console/pull/1105) landed the v1 static sandbox iframe preview per [Issue #1097](https://github.com/ms2sato/agent-console/issues/1097), using `DOMParser.parseFromString(html, 'text/html')` for pre-sandbox sanitize (A6 architect finalized). CodeRabbit raised a Major Security finding based on the MDN caveat "DOMParser inert document may still fetch `<img>`/`<iframe>` subresources".

Empirical verification via Chrome DevTools MCP (3 probes + control experiment against the shipped `sanitizePreviewFragment`) proved that **current Chromium does NOT trigger subresource fetches** from DOMParser-parsed documents, and the codebase's usage never inserts parsed nodes into a live document (only serialized string → blob URL → sandbox iframe with strict CSP). CR bot withdrew the finding for current scope.

**However**, Chrome DevTools MCP cannot verify Safari or Firefox. The MDN caveat may reflect their engine behavior. This Issue tracks the follow-up defense-in-depth for cross-browser posture.

## Scope (v1.1)

- Verify empirically whether Safari and Firefox trigger DOMParser subresource fetches for `<img>` / `<link>` / `<iframe>` / SVG `<image>` / `<video>` / `<audio>` (mirror the PR #1105 verification methodology)
- If fetch behavior differs:
  - Design a pre-parse tokenizer / allowlist that removes `src` / `href` / event handler attributes **before** `parseFromString`
  - Architect A6 revision required (A6 explicitly rejected regex, requires design update)
  - Owner sign-off required (security posture change parity with A4-A6)
- If no fetch (matches Chromium): document as "verified consistent across Chromium/Safari/Firefox", close as no-action-needed

## Non-goals

- Interactive JS execution mode (separate v2 Issue)
- CDN allowlist (separate v2 Issue)

## Related

- [PR #1105](https://github.com/ms2sato/agent-console/pull/1105): v1 landing + empirical Chromium verification
- CodeRabbit thread (withdrawn Major finding, `preview-sandbox.ts:62`)
- MDN `DOMParser.parseFromString` caveat: https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString

## Priority

Low. Current Chromium is safe; cross-browser is a robustness improvement, not a hot bug.

[...]

PR #1194, opened 2026-07-18
test: real-Chromium runner for preview sanitizer mXSS corpus (Issue #1162)

ms2sato:
## Summary

Related: #1162 (this is infra, not the tokenizer/allowlist fix itself)

Adds a real-Chromium regression runner for the embedded-agent preview sanitizer's mXSS corpus (`scripts/run-preview-sandbox-browser-check.mjs`, Playwright-driven). `bun:test` runs on happy-dom, which does not reproduce Chromium's HTML5 foreign-content/adoption-agency parsing (confirmed empirically in PR #1164), so the mXSS corpus's real-browser behavior could previously only be checked via one-off manual Chrome DevTools MCP sessions. This script automates that verification and is a prerequisite for the eventual tokenizer/allowlist redesign scoped to #1106/#1162 (per architect direction, session `84a3a530`, 2026-07-17 consult #2, recorded on #1162).

## What the runner checks

For every corpus vector (5 `NEUTRALIZED_VECTORS` + the known-gap vector):

1. **Production containment (blocking).** Builds the exact document `PreviewPanel.tsx` builds (`sanitizePreviewFragment` -> `buildPreviewDocument`) and loads it into a real `<iframe>` using `PreviewPanel.tsx`'s own current `sandbox` attribute value, read from its source at runtime (not hardcoded) -- a future weakening of that attribute changes what the runner exercises too. Asserts no script execution becomes observable (dialog + console signal). Must hold for every vector, including the known gap, because containment is defense-in-depth (verified during development: even a locally-weakened `sandbox` attribute alone did not let the known-gap vector fire, because `buildPreviewDocument`'s CSP is a second, independent containment layer -- exactly what PR #1164's body documented).
2. **Re-parse neutralization (blocking for the 5 neutralized vectors, informational for the known gap).** Mirrors `preview-sandbox.test.ts`'s `assertFullyNeutralizedAcrossReparse`, but using real Chromium's `DOMParser` instead of happy-dom's -- this is the concrete value this script adds beyond the existing bun:test coverage. For the known-gap vector this is expected to currently find a survivor (reproduces the documented parser-quirk gap under real Chromium); this is logged as an informational note, not a failure, so a future sanitizer fix doesn't have to touch this script to land.
3. **Harness self-test.** Runs once before the corpus: loads a raw (unsanitized) payload into a deliberately permissive iframe and confirms the dialog-detection mechanism actually observes execution. If this fails, the run aborts with exit 2 instead of reporting misleadingly-green corpus results.

## Design questions (per the architect's AC)

- **Runner shape**: script only. I originally also added an opt-in CI workflow (`workflow_dispatch` + label-gated `pull_request`, relying on the Google Chrome preinstalled on `ubuntu-latest` runners so it adds no download cost), but the push credential available in this delegated worktree lacks the GitHub OAuth `workflow` scope, so GitHub rejected the push of any file under `.github/workflows/`. The script is the required deliverable (AC (a) says "script or optional CI job") and is unaffected. The workflow YAML is straightforward to re-add later by anyone with workflow-scope push access -- happy to paste it here or open a tiny follow-up PR if useful.
- **Corpus source**: extracted to a new shared fixture, `packages/client/src/lib/__fixtures__/preview-sandbox-corpus.ts` (the actual sanitizer lives at `packages/client/src/lib/preview-sandbox.ts`, not the `components/preview-sandbox/` path referenced in the Issue -- I read the real location instead). `packages/client/src/lib/__tests__/preview-sandbox.test.ts` was refactored to import `NEUTRALIZED_VECTORS` / `KNOWN_GAP_VECTOR` from this fixture instead of inlining vector strings, so the two suites cannot drift apart. No new corpus vectors were added; existing bun:test assertions are unchanged.
- **Assertion shape**: JS dialog events (Playwright `page.on('dialog', ...)`), since every corpus vector's payload literally calls `alert(N)`. Supplemented by capturing Chromium's own console diagnostic ("Blocked script execution ... sandboxed") as corroborating evidence in the runner's INFO output.
- **Non-goal respected**: the runner never attempts to make a vector fire under raw/unsandboxed HTML as its main assertion -- containment is always checked against `PreviewPanel.tsx`'s actual production composition. The one raw/unsandboxed load in the script is the harness self-test, which exists only to prove the detection mechanism itself works, and is clearly separated from the corpus assertions in both code and output.

## Zero production code change

```
$ git diff --stat 7b71448 HEAD
 bun.lock                                                            |   3 +
 package.json                                                        |   2 +
 packages/client/src/lib/__fixtures__/preview-sandbox-corpus.ts       |  82 +++
 packages/client/src/lib/__tests__/preview-sandbox.test.ts            |  37 +-
 scripts/run-preview-sandbox-browser-check.mjs                        | 389 +++
 5 files changed, 490 insertions(+), 23 deletions(-)
```

`packages/client/src/lib/preview-sandbox.ts` and `packages/client/src/components/workers/PreviewPanel.tsx` are untouched. The production sanitizer is bundled read-only into a browser-runnable module via `Bun.build` at runner start (never written to). `PreviewPanel.tsx`'s `sandbox` attribute is read via a regex against its source text (never written to).

## New dev dependency

`playwright-core@^1.61.1` (root `devDependencies`) -- the driver for real Chromium via CDP. Deliberately `-core`, not `playwright`, so no bundled-browser download is required; the script resolves an existing Chrome/Chromium install (env override, then common system paths, then this repo's documented aarch64 dev-machine snap-chromium path -- see the `chrome-mcp-aarch64-setup` skill for why the snap launcher wrapper itself can't be used directly).

## Runner output (real run on this aarch64 dev host, `/snap/chromium/current/usr/lib/chromium-browser/chrome`)

```
==> using Chromium executable: /snap/chromium/current/usr/lib/chromium-browser/chrome
==> production iframe sandbox attribute (read from PreviewPanel.tsx): ""

==> harness self-test (confirms dialog-detection actually observes execution)
  OK    self-test: permissive sandbox correctly observed execution

==> production containment check (iframe sandbox=""), 6 vectors
  OK    contained: on* attribute on a MathML element (namespace-agnostic attribute removal)
  OK    contained: <script> element nested inside MathML (namespace-agnostic tag removal)
  OK    contained: on* attribute directly on <noscript>
  OK    contained: <script> element whose content uses an HTML-comment decoy
  OK    contained: javascript: URL on xlink:href nested inside <svg><a>, case-varied
  OK    contained: cure53 adoption-agency shape (Issue #1162, known gap, containment-only)
  INFO    browser confirmed the block: error: Blocked script execution in 'blob:null/...' because the document's frame is sandboxed and the 'allow-scripts' permission is not set.

==> re-parse neutralization check (real Chromium DOMParser), 5 vectors
  OK    neutralized: on* attribute on a MathML element (namespace-agnostic attribute removal)
  OK    neutralized: <script> element nested inside MathML (namespace-agnostic tag removal)
  OK    neutralized: on* attribute directly on <noscript>
  OK    neutralized: <script> element whose content uses an HTML-comment decoy
  OK    neutralized: javascript: URL on xlink:href nested inside <svg><a>, case-varied

==> re-parse check on known gap vector (Issue #1162, informational -- not a blocking gate)
  INFO  known gap reproduced under real Chromium: cure53 adoption-agency shape (Issue #1162, known gap, containment-only) -- sanitize -> re-parse still yields a live script/on* survivor, matching Issue #1162

PASSED: 11 assertion(s) passed (2 informational note(s), see INFO lines above)
```

Exit code: 0. All corpus vectors -- including the known #1162 gap -- did NOT fire under the real production `<iframe sandbox="">` composition. The known gap's re-parse survivor IS reproduced (informational), matching the documented current state.

Also negative-tested during development: temporarily weakened `PreviewPanel.tsx`'s `sandbox` attribute locally (not committed) to confirm the runner's detection is real, not a tautology -- documented in the script's comments, including the finding that `buildPreviewDocument`'s CSP is a second independent containment layer that still blocks execution even when only the sandbox attribute is weakened (consistent with PR #1164's two-layer containment finding).

## Verification

- `bun run typecheck` -- `TYPECHECK_EXIT: 0`
- `bun run test` (full suite) -- `TEST_EXIT: 1`, but the only failure is the pre-existing, unrelated `packages/server/src/services/inbound/__tests__/resolve-targets.test.ts` flake ("not a git repository"), verified via baseline `git stash` comparison to reproduce identically on unmodified `origin/main` (same flake PR #1164 documented). `packages/client` (2064 pass, 0 fail), `packages/shared`, `packages/integration`, `packages/embedded-agent` all clean.
- `node .claude/skills/orchestrator/preflight-check.js` -- clean (no coverage gaps, no rule/skill duplication, language check clean, comment blame-shift check clean after paraphrasing the fixture's Issue references).
- `bun run lint` -- clean (dependency-cruiser, madge, knip all pass; `playwright-core` resolves as used via the runner script entry point).
- Runner invocation -- see output above, exit 0.

## Test plan

- [x] `bun scripts/run-preview-sandbox-browser-check.mjs` -- all 6 corpus vectors contained under production sandbox composition; 5/5 neutralized vectors confirmed under real Chromium's DOMParser; known #1162 gap reproduced as informational evidence, not a blocking failure
- [x] `packages/client/src/lib/__tests__/preview-sandbox.test.ts` -- 20 pass, 0 fail (refactored to consume the shared fixture; assertions unchanged)
- [x] `bun run typecheck` -- exit 0
- [x] `bun run test` -- only the pre-existing, unrelated `resolve-targets.test.ts` flake (baseline-verified)
- [x] `bun run lint` -- clean
- [x] `node .claude/skills/orchestrator/preflight-check.js` -- clean

## Merge policy

Test-infra, zero production code change -- per Issue #1162's owner-directed merge policy, this is in the Orchestrator-merge-without-owner-approval bucket, gated on architect review (session `84a3a530`).

## Note on CodeRabbit

Both local CodeRabbit CLI (timed out after 5 min, no output -- consistent with the known headless-auth pattern in delegated worktrees) and the GitHub-side bot (rate limited at PR open; a manual `@coderabbitai review` re-trigger after the stated wait window returned "Review finished ... does not re-review already reviewed commits" with no actual review submitted) were unavailable at this PR's review window. No CodeRabbit feedback obtainable from either source. Merging under existing [PR Merge Authority](https://github.com/ms2sato/agent-console/blob/main/.claude/skills/orchestrator/SKILL.md#pr-merge-authority) -- per this task's explicit merge policy (test-infra, zero production code change), architect review (session `84a3a530`) is the gate, not CodeRabbit.


ms2sato:
@coderabbitai review

Issue #1198, opened 2026-07-18
bun-pty provider PTY master fd leak on natural exit (same signature as #1196, non-default provider)


ms2sato:
Surfaced by the #1196 sibling audit (Sprint 2026-07-18). Same leak signature as #1196 but the fix mechanism is non-trivial and requires new design; split per architect verdict to keep #1196's critical fix PR scope-clean.

## Reach

**Non-default provider.** \`PTY_PROVIDER\` default is \`bun-terminal\` (\`packages/server/src/lib/server-config.ts:135\`); this Issue only affects environments that explicitly set \`PTY_PROVIDER=bun-pty\`. Operators of such environments should be aware that #1196's fix does NOT close this leak for their provider.

## Probe evidence

\`node_modules/.bun/bun-pty@0.4.2/.../dist/index.js\` \`Terminal._startReadLoop()\` on natural-exit branch:

\`\`\`js
} else if (n === -2) {
  const exitCode = lib.symbols.bun_pty_get_exit_code(this.handle);
  this._onExit.fire({ exitCode });
  break;               // <-- bun_pty_close(handle) is NEVER called
}
\`\`\`

Direct measurement (delegate probe in #1196 fix session):

- Natural exit only (5 iterations, no explicit \`pty.kill()\`): ptmx master fd count grows 3, 6, 9, 12, 15 (linear) — same signature as #1196.
- Adding \`pty.kill()\` after natural exit: growth stops (native \`bun_pty_close\` fires from the kill path).

## Why the naive fix is BANNED

**Do NOT ship a \"call \`pty.kill()\` from natural-exit onExit handler\" 1-line fix.** It introduces two concrete regressions:

1. **PID reuse signaling risk.** \`bun_pty_kill(handle)\` sends a signal to the pid recorded in the handle. If the process has already exited AND the OS has reused the pid for an unrelated process, the signal reaches the wrong target. The Rust-side handle→pid mapping's liveness check has NOT been read; without confirmed liveness, this vector is real.
2. **Double-fire of \`_onExit\`.** \`kill()\` sets an internal \`_closing\` flag that suppresses \`_onExit\`, but that flag is set inside \`kill()\` only — NOT at natural-exit detection. Calling \`kill()\` from the natural-exit onExit handler fires \`_onExit\` a second time. Consumers (\`worker-manager\`) will see a spurious duplicate exit event.

[...]

Issue #1197, opened 2026-07-18
[umbrella] multi-user orphan cleanup hardening (isProcessAlive EPERM / elevation-aware kill / SESSION_ID tree cleanup)


ms2sato:
Umbrella for 3 structural defects surfaced by the Sprint 2026-07-18 PTY leak investigation (parent Issue #1196). All confirmed via architect (session 84a3a530) direct code read on the dogfood host. **NOT bundled into #1196's critical fix PR** — architect verdict: scope hygiene, split to prevent event-hardening PR bloat.

## Defect 1: `killOrphanWorkers` uses raw `process.kill`

`SessionInitializationService.killOrphanWorkers` calls `processKill(worker.pid, SIGTERM/SIGKILL)`. In multi-user mode, `worker.pid` refers to a **root-owned elevation wrapper** — `agentconsole` cannot signal it, EPERM. Failure is logged but processing continues → orphan silently survives.

Fix: cross-user kill via `runAsUser`-based helper (elevation-helpers.md pattern). Candidate primitive: `killAsUser(pid, signal, username)` in `packages/server/src/services/privilege-elevation.ts`. Strict-thin-wrapper contract per elevation-helpers.md.

## Defect 2: `isProcessAlive` misreports EPERM as "dead"

`isProcessAlive` implementation: `process.kill(pid, 0)` in try, catch returns false. Does not distinguish EPERM (exists, no permission) from ESRCH (does not exist).

- `kill(pid, 0)` with EPERM means "process exists but signaling not permitted"
- Multi-user: root-owned wrapper → signal 0 returns EPERM → live orphan misreported as dead → `killOrphanWorkers` silently skips (no error log)
- **Silent — no error log surfaces this class of skip**

Fix: `EPERM = alive`, only `ESRCH = false`. 1-line + polarity test. Audit all callers for "assume dead" assumptions before landing.

## Defect 3: Orphan cleanup single-pid vs. live kill tree-aware asymmetry

Normal `killWorker` takes descendant-pid snapshot and processes tree. Restart-time `killOrphanWorkers` only sees persisted `worker.pid` (dead wrapper). `shutdownAppContext` doesn't explicitly kill PTYs — relies on kernel master-fd close → SIGHUP. **SIGHUP only reaches foreground process group**; background/detached descendants (chrome-devtools MCP, `npm exec chrome`, agent-launched `bun run dev` / `npm start` in terminal worker) survive.

[...]

Issue #828, opened 2026-06-22
Stage 3: retire bun-pty + #823 shim/lib-copy once bun-terminal default has soaked


ms2sato:
## Goal

Retire the `bun-pty` dependency and every workaround built around its native shared library — **stage 3 of the 3-stage rollout** committed in PR #826. Run only after the stage-2 default flip (separate Issue) has soaked in production for at least one release without regression.

## Why

`bun-pty` is a native module that loads `librust_pty.{so,dylib}` at startup. That single property has caused two distinct production-deploy bugs already:

- **#820 / #823**: the bundled `dist/index.js` could not resolve the shared library, requiring the `src/shim.ts` entrypoint + a build-time lib copy into `dist/rust-pty/target/release/`. Operators on Model A and Model B both had to coexist with the workaround.
- The systemd unit templates carried a `BUN_PTY_LIB=...` environment fallback for any deployment shape that didn't match the shim's discovery path.

`Bun.Terminal` (adopted in #826, default-flipped in stage 2) is a built-in API with no native shared library, so all of that bundling friction disappears once we stop shipping `bun-pty`.

## Pre-conditions (must hold before this Issue starts)

- [ ] The stage-2 default flip (separate Issue) has been merged.
- [ ] At least one release cycle has elapsed with `bun-terminal` as the default and no regression reports (color, resize, signal, env, exit, multi-user identity).
- [ ] No external operator is known to be pinning `PTY_PROVIDER=bun-pty` for a use case we haven't addressed in `bun-terminal` (poll if uncertain).

## Scope

- `package.json` — remove `bun-pty` from `dependencies`.
- `packages/server/src/lib/pty-provider.ts` — remove `bunPtyProvider`, the `bunPtyOptions` shape, and the `PTY_PROVIDER` switch (always use `bunTerminalProvider`). Keep the `PtyProvider` interface (still useful for DI in tests).
- `packages/server/src/lib/server-config.ts` — remove `PTY_PROVIDER` from the config (no longer meaningful).
- `packages/server/build.ts` — remove the `NATIVE_DEPENDENCIES` entry for `bun-pty`, the `dist/rust-pty/target/release/` lib copy step, and the `dist/index.js` → `dist/server.js` shim renaming (the bundler can emit `dist/index.js` directly again since there is no per-platform lib to discover).
- `packages/server/src/shim.ts` and `packages/server/src/lib/bun-pty-shim.ts` (+ their tests) — delete. Remove their `knip.json` entry points.
[...]

Issue #827, opened 2026-06-22
Stage 2: flip PTY_PROVIDER default from bun-pty to bun-terminal (rollout after #826)


ms2sato:
## Goal

Flip the default `PTY_PROVIDER` from `bun-pty` to `bun-terminal`, completing **stage 2 of the 3-stage rollout** committed in PR #826 (which adopted `bunTerminalProvider` behind the `PTY_PROVIDER` flag).

## Why

PR #826 (Closes #824) shipped `bunTerminalProvider` as a coexisting second PTY backend, with the original color regression confirmed fixed by both byte-level probe and owner visual E2E (Claude Code orange / "Opus 4.7 with high effort" / message round-trip — all visually identical to the `bun-pty` baseline). Default was left at `bun-pty` so adoption is intentionally staged.

Stage 2 is to **switch the default** and run the full surface under `bunTerminalProvider` for a window before retiring `bun-pty` entirely (stage 3, separate Issue).

## Scope

- `packages/server/src/lib/server-config.ts` — change `PTY_PROVIDER` default from `'bun-pty'` to `'bun-terminal'`. The escape hatch (operator setting `PTY_PROVIDER=bun-pty` explicitly) **must remain** for stage 2; do not remove the `bun-pty` arm.
- Update the relevant `__tests__/server-config.test.ts` default-value test.
- No production-code behavior change other than default selection. `bunPtyProvider` stays compiled and selectable.
- Update `docs/multi-user-setup-guide.md` (or the right doc surface) to mention `PTY_PROVIDER` and the new default, with a note that `bun-pty` is still selectable for one release as a rollback escape hatch.

## Out of scope (stage 3 — separate Issue)

- Removing the `bun-pty` dependency from `package.json`.
- Removing `bunPtyProvider` and the related DI arm.
- Removing the #823 shim entrypoint (`src/shim.ts`, dist lib copy in `build.ts`).
- Removing `BUN_PTY_LIB` from any systemd templates.

These all become safe once stage 2 has run in production for one release without regression.

## Verification

- Default behavior of every PTY consumer (`worker-manager` spawn paths, both agent + terminal, both single-user and multi-user modes) now runs through `bunTerminalProvider`.
- Owner re-runs the visual confirm one more time on the default path (no `PTY_PROVIDER` env set) to catch any subtle difference from the explicit-flag path.
- Full test suite green; `bun run dev` smoke green; Claude / lazygit / htop / git diff colors render normally.

## Cross-reference

[...]

PR #1199, opened 2026-07-18
fix: release Bun.Terminal master fd on PTY exit to prevent ptmx leak

### LINKED ISSUE #1196: PTY master fd (/dev/ptmx) leak in bun-terminal provider — memory/PTY exhaustion on long-running servers
## Symptom

On a long-running agent-console server (14 days uptime), the host progressively degrades until the system is essentially unusable:

- Load average: **93.11 / 82.91 / 63.67** (7-core host)
- Memory: **22.9G / 23.4G** used (RAM effectively exhausted)
- Swap: **7.93G / 8.00G** used (212 KB free)
- Multiple processes stuck in the `D` (uninterruptible I/O wait) state due to swap thrashing
- `who` reported 59 residual `pts/*` sessions all owned by the `agentconsole` user, opened at the same wall-clock timestamp days earlier

Manual restart of the server process (parent `bun run start`) recovered the system.

## Impact

- **Severity: high.** Progressive resource exhaustion; hard failure mode that eventually requires manual restart. Silent accumulation — nothing in the application logs surfaces the leak until the host is memory-starved.
- **Reach:** every long-running deployment. Scales with how many PTY workers are spawned over the server's lifetime.
- Related but distinct from #1087 (which fixed *child claude process* orphaning on worktree removal). This bug is about **`/dev/ptmx` master file descriptors accumulating in the parent `bun` process itself**, independent of whether descendants are cleaned up.

## Smoking gun

Before restart, the running server process was inspected directly:

| Metric | Value |
|---|---|
| Total open fds in the `bun` process (PID 484856) | 242 |
| Of which `/dev/ptmx` (PTY master) | **168** |
| Of which `/dev/pts/*` (PTY slave) | 0 |
| System-wide allocated PTYs (`/proc/sys/kernel/pty/nr`) | 116 |
| Active user sessions (`who \| wc -l`) | 59 |

Interpretation:
- The `bun` process is holding 168 master fds while only ~59 user sessions are logically active.
- No `/dev/pts/*` slaves are held in the parent (expected — slaves live in the child). The leak is entirely on the master side.
- The 168 vs 116 gap is consistent with some ptmx fds pointing at PTYs whose partner has already been released — the master ref is what keeps them from being reclaimed.

After restarting the server:
- `/proc/sys/kernel/pty/nr` dropped from **116 → 38** immediately.
- Load average began recovering (1-min value fell from 93 to 28 within ~1 minute of the cleanup).

The 78-PTY drop across the restart, with no user activity change, is the direct evidence that the parent process was the holder.

## Environment

- OS: Ubuntu (host, 14 days uptime).
- Runtime: `bun` (see reproduction step 1 for exact version — please attach when filing details).
- Service: `AgentConsole` (`bun dist/index.js`), started as a long-running server process.
- **PTY implementation: `PTY_PROVIDER=bun-terminal` (the default per `packages/server/src/lib/server-config.ts:135`).** Uses `Bun.spawn({ terminal: ... })` — no external PTY module (`bun-pty`, `node-pty`) involved.
- Deployment: single-user mode, no privilege elevation involved in the affected code path.

## Suspected location

`packages/server/src/lib/pty-provider.ts` — the `BunTerminalPtyAdapter` class and the surrounding `bunTerminalProvider.spawn(...)` factory.

Observations from reading the current code:

1. **`kill()` sends a signal but does not close the terminal handle** (`pty-provider.ts:196-200`):

   ```ts
   kill(signal?: string): void {
     this.subprocess.kill((signal ?? 'SIGTERM') as NodeJS.Signals);
   }
   ```

   `Bun.Subprocess.kill` sends `SIGTERM`; it does not close `subprocess.terminal` (the Bun.Terminal handle that owns the ptmx fd). If the ptmx fd is not automatically closed when the child exits, this alone would leak on every kill.

2. **`subprocess.exited.then(...)` fires `onExit` but does not release the terminal** (`pty-provider.ts:119-122`):

   ```ts
   void this.subprocess.exited.then((exitCode) => {
     this.fireExit(exitCode);
   });
   ```

   The adapter never explicitly disposes / closes `this.terminal` after the child exits. The consumer's disposable (`onData`/`onExit` return values) only nulls out listener slots (`pty-provider.ts:142-175`); it does not release the terminal handle.

3. **Consumer teardown (`packages/server/src/services/worker-manager.ts`) drops the reference but does not close the master.** `killWorker` (`worker-manager.ts:1001-1106`) is well-structured — it awaits `pty.onExit`, kills descendants, and finally calls `detachPty(worker)` which sets `worker.pty = null`. But `detachPty` (`worker-manager.ts:964-966`) is a one-liner:

   ```ts
   detachPty(worker: InternalPtyWorker): void {
     worker.pty = null;
   }
   ```

   No fd close, no terminal disposal. This relies on GC + Bun runtime finalization to reclaim the ptmx fd. If a stray reference (a still-live callback closure, an unfired listener, or an internal Bun handle) keeps the adapter reachable, the fd persists.

4. **Unmanaged-exit path also relies on the same GC assumption** (`worker-manager.ts:701-707`):

   ```ts
   const onExitDisposable = pty.onExit(({ exitCode, signal }) => {
     // ... logging ...
     this.detachPty(worker);
     // ...
   });
   ```

   Same pattern — set to null and hope.

The consistent hypothesis: **Bun's `Subprocess.terminal` does not release its underlying `/dev/ptmx` master fd on child exit unless something explicitly closes the Terminal handle**, and the adapter has no code path that closes it.

## Reproduction

1. **Capture the exact bun version** on the affected host:
   ```bash
   bun --version
   ```

2. **Baseline PTY count** immediately after starting the server:
   ```bash
   cat /proc/sys/kernel/pty/nr
   ```

3. **Load the server** with any workflow that spawns and disposes PTY workers (open sessions, add workers, close them; or run agent delegate workflows that exit naturally).

4. **Sample the ptmx-fd count in the `bun` process** every minute (requires root to read another user's `/proc/<pid>/fd`):
   ```bash
   BUNPID=$(pgrep -f 'bun dist/index.js' | head -1)
   for i in {1..30}; do
     ptmx=$(ls -l /proc/$BUNPID/fd/ 2>/dev/null | grep -c ptmx)
     pty=$(cat /proc/sys/kernel/pty/nr)
     echo "$(date +%T) ptmx=$ptmx pty_total=$pty"
     sleep 60
   done
   ```

   Expected (buggy) behavior: `ptmx` climbs monotonically and stays above the count of currently-active sessions. Expected (fixed) behavior: `ptmx` returns to a stable baseline within seconds of each worker exit.

## Investigation checklist

- [ ] Confirm the bun version on the affected host and cross-check against upstream bun issues for known `Bun.spawn({ terminal })` fd-leak reports (`https://github.com/oven-sh/bun/issues`).
- [ ] Add an explicit terminal-close step to `BunTerminalPtyAdapter`. Candidates:
  - Whatever method `Bun.Terminal` exposes for disposal (needs verification against the current `@types/bun`; the terminal handle likely has a `close()` or similar).
  - If no such API exists, verify whether closing the ReadableStream sides (`subprocess.stdout` / `subprocess.stderr`) is the intended release path.
- [ ] Ensure `kill()` waits for `subprocess.exited` before dropping references, so any Bun-side cleanup that keys off exit completion actually runs.
- [ ] Extend the existing `pty-provider.test.ts` with a leak-shape test: spawn N adapters, kill each, force GC, assert `/proc/self/fd` ptmx count returns to baseline.
- [ ] Consider adding a long-running smoke test (`scripts/smoke/`) that spawns 100 short-lived PTYs in a loop and verifies the ptmx count stays bounded, matching the `os-environment-coupling.md` "real-machine smoke test" discipline.
- [ ] Once fixed, run the reproduction step 4 monitor on the dogfood host and confirm ptmx stays bounded.

## Recommended priority

**High.** The failure mode is progressive and eventually forces manual restart of a shared server. Users on long-running deployments cannot avoid it without periodic restarts (a workable stopgap, but not a fix).

## Related

- #1087 — sibling problem (descendant claude processes leaked on worktree removal); already fixed. This bug is about the **parent-side master fd**, not the child processes, so #1087's fix (`listDescendantPidsImpl` + `signalPids`) does not address it.
- `packages/server/src/lib/pty-provider.ts` — suspect implementation.
- `packages/server/src/services/worker-manager.ts:1001-1106` — consumer teardown, which is careful but relies on GC for terminal reclamation.

ms2sato:
## Summary

`BunTerminalPtyAdapter` (`packages/server/src/lib/pty-provider.ts`, the default `PTY_PROVIDER=bun-terminal` path) held a `Bun.Terminal` handle that owns the PTY master fd (`/dev/ptmx`) and never released it. A long-running server accumulates leaked ptmx fds (168+ observed in production) until it needs a manual restart.

Fix: an idempotent `dispose()` on the adapter, wired into two lifetime endpoints, backed by real-fd regression tests and a post-deploy smoke test.

Closes #1196

## Phase 0 probe result (real Bun 1.3.10 execution, not typings)

- Without any explicit close: each spawn+exit cycle leaks **3** ptmx-fds that never return (confirmed growing linearly over 5 iterations: 3, 6, 9, 12, 15 — reproduces the issue's accumulation shape). The multiplicity of 3 fds per single `Bun.Terminal` spawn is not fully root-caused (likely internal dup()s for read/write/event-loop registration; `/proc/self/fd` readlink cannot distinguish a dup from a fresh open) — recorded here so a future reader isn't confused by the ×3 factor in the regression test's expected deltas.
- `subprocess.terminal.close()` called after `subprocess.exited` resolves releases all 3 fds, confirmed flat across repeated cycles, for both the natural-exit path and the killed-process path.
- `close()` on a **still-running** process also works, does not throw, and additionally delivers `SIGHUP` to the child (child exits within ~2ms) — see "kill-timeout give-up path" below.
- `close()` is idempotent (safe to call twice); `terminal.closed` flips to `true`.

**Important correction on the GC claim.** An earlier draft of the regression test (and this PR's own investigation notes) stated GC "cannot" reclaim the fd. That's not quite right: Bun's native `Terminal` binding appears to have a GC finalizer that *can* release the fd once the JS wrapper is unreachable — confirmed by a standalone probe with `Bun.gc(true)`. The actual production problem is that `InternalPtyWorker.pty` stays *reachable* (referenced via session/worker maps) for the life of the worker, so that finalizer never gets a chance to fire until `detachPty` nulls the reference — by which point 168+ fds have already accumulated. Relying on incidental GC is unsound regardless of whether GC *can* theoretically reclaim it; only a deterministic `dispose()` call fixes the production symptom. Docs/comments in this PR are worded to reflect this precisely.

## Design

- `BunTerminalPtyAdapter.dispose()`: idempotent (`disposed` guard), calls `this.terminal.close()` in a try/catch (logs a warning on failure, never throws — matches `backend.md` "Resource Cleanup").
- Called unconditionally at the top of `fireExit()`, **before** the existing early-return checks (`fireExit` currently returns early when no `onExit` listener is attached yet, which would otherwise skip disposal on that path).
- `PtyInstance` type widened to `IPty & { dispose?(): void }` — optional so `bunPtyProvider`'s native `Terminal` (no such method) still satisfies the type structurally. This also makes the bun-pty/bun-terminal asymmetry explicit at the type level (see "Sibling audit" below).
- `WorkerManager.detachPty` now calls `worker.pty?.dispose?.()` before nulling, as a backstop.

### `detachPty` call-site enumeration (2 production call sites, verified via grep)

| Site | Trigger | Exit confirmed? | Notes |
|---|---|---|---|
| `worker-manager.ts:707` (inside `pty.onExit(...)` in `setupWorkerEventHandlers`) | Unexpected/natural process exit (only fires when NOT preceded by `killWorker`, since `killWorker` disposes these listeners before calling `pty.kill()`) | Yes — inside the exit-observed callback | `dispose()` here is redundant with the adapter's own `fireExit`-triggered dispose (already ran by construction), but harmless — idempotent. |
| `worker-manager.ts:1106` (end of `killWorker`, after the `exitOrTimeout` race) | Managed kill | **Mixed.** If the race resolved via real exit: redundant/idempotent, same as above. If it resolved via the `PTY_EXIT_TIMEOUT_MS` timeout (process didn't exit in time — code already logs a warning and "proceeds anyway"): **not** confirmed-exited. | This is the one path where `dispose()` is the *first* cleanup call and forcibly closes a still-running process's terminal. Per the Phase 0 probe, this is safe (doesn't throw) and has the beneficial side effect of delivering `SIGHUP` to the still-running process — likely terminating it. Strictly an improvement over current behavior (leaked fd + a process that may or may not ever die on its own). |

## Sibling audit: bun-pty provider (non-default, `PTY_PROVIDER=bun-pty`)

Read `node_modules/.../bun-pty/dist/index.js`'s `Terminal._startReadLoop()`: on natural exit (`n === -2` from the read loop), it fires `onExit` but never calls `bun_pty_close(handle)` — only the explicit `kill()` path does. Confirmed via direct execution: natural-exit-only cycles leak the same 3× shape (3, 6, 9, 12, 15 across 5 iterations); calling `.kill()` again after natural exit stops further growth.

**This PR does not fix bun-pty.** Initially filed as a separate Issue (#1198), but the owner pointed out — and I confirmed by reading it — that **Issue #828 ("Stage 3: retire bun-pty...")** already tracks full removal of `bunPtyProvider` and the `PTY_PROVIDER` switch (Stage 2's default-flip to `bun-terminal` is already merged/closed via #827; #828 is gated on a soak-period pre-condition). Once #828 lands, the leaky bun-pty code path is deleted rather than patched, so a standalone fix isn't worth carrying. #1198 was closed with the probe data preserved as a comment, cross-linked to #828, in case #828's pre-conditions don't hold and bun-pty needs to stick around longer than planned.

## Doc drift fix (touched-file, same PR per architect ruling)

`pty-provider.ts`'s doc comments claimed `bunPtyProvider` is the default "for backward compatibility." That's stale — `serverConfig.PTY_PROVIDER` (`server-config.ts:135`) actually defaults to `'bun-terminal'` (flipped in #827). Corrected both doc comments in this PR since they're in a file already being touched.

## Tests

- `pty-provider.test.ts` — 3 new unit tests using the existing mocked-`Bun.spawn` harness: `dispose()` called exactly once on `subprocess.exited`; manual `dispose()` + later exit still calls close exactly once; double-`dispose()` calls close exactly once.
- `pty-provider-fd-leak.test.ts` (new) — real-fd regression test, Linux-gated (`describe.skipIf`), retains strong references to every spawned adapter for the assertion's lifetime (see the GC-masking note below), covers both the natural-exit path and the `kill()` path, N=10 spawns each, asserts ptmx fd count returns to baseline.
- `worker-manager.test.ts` — asserts `detachPty` calls `dispose()` on the pty when present.
- `scripts/smoke/check-pty-fd-leak.ts` (new) — 100 real spawn/kill cycles through production `bunTerminalProvider`, asserts both `/proc/self/fd` ptmx count and `/proc/sys/kernel/pty/nr` (kernel-wide counter) stay flat. Documented in `docs/multi-user-setup-guide.md` Post-deploy Verification and `.claude/rules/test-trigger.md`.

### Polarity verification (mandatory per `workflow.md`)

Verified independently by me (not just the implementing delegate) by disabling `this.dispose()` in `fireExit()` and re-running `pty-provider-fd-leak.test.ts`:

```
Expected: 0
Received: 30
✗ natural exit: ptmx fd count returns to baseline after N short-lived spawns

Expected: 30
Received: 60
✗ kill() path: ptmx fd count returns to baseline after N kill()-then-exit cycles

 0 pass
 2 fail
```

Restored the fix, re-ran: `2 pass, 0 fail`. Fix confirmed to be load-bearing for both tests, not incidental.

### GC-masking discovery (important test-design correction, found during implementation)

The implementing delegate's first draft of the regression test/smoke didn't retain strong references to spawned adapters. Because `Bun.Terminal` appears to have a GC finalizer, unretained loop-scoped `pty` variables became GC-eligible mid-test and were incidentally finalized — masking the leak (~90% of runs converged to baseline even with the fix disabled). Fixed by retaining every spawned adapter in an array for the assertion's lifetime, mirroring how `InternalPtyWorker.pty` holds a live reference in production until `detachPty` runs — this makes `dispose()` the only possible release path in the test. See the file header of `pty-provider-fd-leak.test.ts` and `scripts/smoke/check-pty-fd-leak.ts` for the full account.

## Verification

- `bun run typecheck` — clean, all packages.
- `bun run test` — full suite green (paste below).
- `bun run check:pty-fd-leak` — smoke passes (paste below).
- `bun run check:lang` — clean, 111 files.
- `bun scripts/check-source-comment-blame-shift.mjs` — 0 violations.
- `node .claude/skills/orchestrator/preflight-check.js` — all clean (coverage, rule/skill duplication, language, blame-shift).

```
$ bun run test (last lines)
@agent-console/server test: 3534 pass, 1 skip (expected memfs-active skip for pty-provider-fd-leak.test.ts under the full suite — run it alone to see the real assertions), 0 fail
@agent-console/client test: 2065 pass, 0 fail
@agent-console/shared test: 573 pass, 0 fail
@agent-console/integration test: 72 pass, 0 fail
@agent-console/embedded-agent test: 257 pass, 0 fail
scripts/hooks tests: 457 pass, 0 fail
TEST_EXIT: 0
```

```
$ bun run check:pty-fd-leak
==> ran 100 spawn/kill cycles via bunTerminalProvider
==> ptmx fd count (this process, /proc/self/fd)
  OK    ptmx fd count non-increasing (before=0, after=0)
==> kernel pty counter (/proc/sys/kernel/pty/nr)
  OK    kernel pty counter non-increasing (before=39, after=39)
PASSED: 2 assertion(s) passed
```

## Reversibility

Single revert restores prior (leaky) behavior. Changes are scoped to `pty-provider.ts` + `worker-manager.ts`'s `detachPty` and their tests/docs — no schema, no wire-protocol, no other service touched.

## Out of scope

- bun-pty native provider fix — superseded by planned full removal, Issue #828 (see "Sibling audit" above).
- The 3 multi-user orphan-cleanup defects (`isProcessAlive` EPERM / elevation-aware kill / `SESSION_ID` env tree cleanup) — tracked separately as Issue #1197, intentionally not bundled here.

## Post-merge

Will measure dogfood ptmx count post-deploy and record it in Issue #1196 as recovery evidence, per the architect's AC.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PTY resource cleanup during normal exits and forced detach operations, reducing the risk of file-descriptor leaks.
  * Ensured PTY disposal is safe to call multiple times.

* **Documentation**
  * Added guidance for running a post-deployment PTY resource smoke check.
  * Clarified the default and legacy PTY provider options.

* **Tests**
  * Added automated coverage for PTY disposal, detach cleanup, and repeated spawn/terminate scenarios.
  * Added a `check:pty-fd-leak` command for Linux environments.

### DIFF
--- a/.claude/rules/test-trigger.md
+++ b/.claude/rules/test-trigger.md
@@ -40,6 +40,10 @@ When modifying production files matching these patterns, corresponding test file
 
 PRs touching `packages/client/src/lib/preview-sandbox.ts`, `packages/client/src/lib/__fixtures__/preview-sandbox-corpus.ts`, or `packages/client/src/components/workers/PreviewPanel.tsx` must run `bun run check:preview-sandbox-browser` locally before pushing. This runs `scripts/run-preview-sandbox-browser-check.mjs`, which re-verifies the mXSS regression corpus against a real Chromium browser — `bun:test`'s happy-dom environment does not reproduce Chromium's HTML5 parsing edge cases (see `.claude/rules/os-environment-coupling.md`). This check is a real-browser regression gate, not a sibling-test requirement, so it is not part of the `preflight-check.js` coverage patterns above.
 
+## Additional Verification: PTY Master FD Leak Check
+
+PRs touching `packages/server/src/lib/pty-provider.ts` or `packages/server/src/services/worker-manager.ts`'s `detachPty` must run `bun run check:pty-fd-leak` locally before pushing. This runs `scripts/smoke/check-pty-fd-leak.ts`, which drives 100 real spawn/kill cycles through the production `bunTerminalProvider` and asserts that the process's ptmx-fd count (`/proc/self/fd`) and the kernel-wide allocated-pty counter (`/proc/sys/kernel/pty/nr`) stay flat — confirming `BunTerminalPtyAdapter.dispose()` actually releases the `Bun.Terminal` master-fd handle deterministically, rather than relying on the object becoming unreachable and incidentally GC-finalized (unsound in production, where `InternalPtyWorker.pty` stays reachable via session/worker maps for the life of the worker) (see Issue #1196). This check is a real-fd regression gate, not a sibling-test requirement, so it is not part of the `preflight-check.js` coverage patterns above.
+
 ## Before Creating a PR
 
 Run the coverage check to verify all production files have corresponding tests:
--- a/docs/multi-user-setup-guide.md
+++ b/docs/multi-user-setup-guide.md
@@ -1108,6 +1108,45 @@ Passing the current process user as `<target-user>` runs in a degenerate
 same-user mode where `spawnAsUser` bypasses elevation — this still exercises
 the full Bash tool-call round trip except the actual OS-boundary crossing.
 
+### PTY master fd leak check
+
+Run after every deploy that touches `packages/server/src/lib/pty-provider.ts`
+or `packages/server/src/services/worker-manager.ts`'s `detachPty` (unlike the
+other smokes above, this one is not privilege-elevation-specific -- it
+verifies a resource-lifecycle contract of the default `bunTerminalProvider`):
+
+```bash
+bun run check:pty-fd-leak
+```
+
+What it verifies: `bunTerminalProvider` wraps `Bun.spawn({ terminal: ... })`,
+whose returned `Bun.Terminal` handle owns the PTY master fd (`/dev/ptmx`).
+Bun's native binding appears to release the fd via a GC finalizer once the
+`Bun.Terminal` wrapper becomes unreachable, but production `InternalPtyWorker`
+objects stay reachable (referenced via session/worker maps) for the life of
+the worker, so incidental GC is not a reliable release path -- only an
+explicit `Bun.Terminal.close()` call deterministically releases it.
+`BunTerminalPtyAdapter.dispose()` performs that release and is wired into
+both the adapter's own
+`subprocess.exited`-triggered exit path and `WorkerManager.detachPty` (a
+backstop for the case where a killed PTY's exit was never confirmed within
+the kill timeout). The smoke runs 100 spawn/kill cycles through the
+production `bunTerminalProvider` and asserts that both the process's own
+ptmx-fd count (`/proc/self/fd`) and the kernel-wide allocated-pty counter
+(`/proc/sys/kernel/pty/nr`) are flat (non-increasing) across the run,
+confirming the master fd is actually released rather than merely
+believed-released. See Issue #1196.
+
+Exit codes: `0` all assertions passed, `1` an assertion failed (a leak was
+observed), `2` bad usage / cannot run (non-Linux; the check depends on
+`/proc`).
+
+This smoke is scoped to `bunTerminalProvider` (the default `PTY_PROVIDER`).
+The legacy native `bunPtyProvider` (`PTY_PROVIDER=bun-pty`) has a
+similar-shaped leak on natural process exit, but rather than a standalone
+fix, `bun-pty` itself is slated for full removal (Issue #828), which deletes
+the leaky code path entirely -- out of scope here.
+
 ## Local Multi-User Dev Mode
 
 `scripts/dev-multiuser.sh` (also runnable as `bun run dev:multiuser`) starts
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "test:coverage": "bun run --filter '*' test:coverage",
     "check:lang": "bun scripts/check-public-artifacts-language.mjs",
     "check:preview-sandbox-browser": "bun scripts/run-preview-sandbox-browser-check.mjs",
+    "check:pty-fd-leak": "bun scripts/smoke/check-pty-fd-leak.ts",
     "hooks:install": "bun scripts/install-hooks.mjs",
     "preinstall": "bun scripts/check-bun-version.mjs",
     "postinstall": "bun scripts/install-hooks.mjs || echo 'hooks:install skipped (likely no git or non-interactive env)'",
--- a/packages/server/src/__tests__/utils/memfs-detection.ts
+++ b/packages/server/src/__tests__/utils/memfs-detection.ts
@@ -1,0 +1,21 @@
+/**
+ * Detect whether `fs`/`node:fs`/`fs/promises` have been swapped for memfs by
+ * an earlier test file's `mock.module` call (see `mock-fs-helper.ts`).
+ * `mock.module` is process-global in bun:test (see `.claude/rules/testing.md`
+ * "Module-Level Mocking" anti-pattern), so any test file that runs real-fs
+ * probes (kernel-level PTY/fd checks, mode/ownership checks, etc.) needs to
+ * detect this and skip gracefully rather than fail against the in-memory
+ * simulation.
+ *
+ * Sentinel: `/proc` exists on every Linux real fs but is not populated in the
+ * memfs volume.
+ */
+export async function isMemfsActive(): Promise<boolean> {
+  try {
+    const fsp = await import('fs/promises');
+    await fsp.lstat('/proc');
+    return false;
+  } catch {
+    return true;
+  }
+}
--- a/packages/server/src/__tests__/utils/mock-pty.ts
+++ b/packages/server/src/__tests__/utils/mock-pty.ts
@@ -20,6 +20,8 @@ export class MockPty {
   private dataCallback: ((data: string) => void) | null = null;
   private exitCallback: ((event: { exitCode: number; signal?: number }) => void) | null = null;
   killed = false;
+  /** Set true when dispose() is called. Mirrors PtyInstance's optional dispose(). */
+  disposed = false;
   writtenData: string[] = [];
   currentCols = 120;
   currentRows = 30;
@@ -76,6 +78,11 @@ export class MockPty {
         }
       });
     }
+  }
+
+  /** Mirrors PtyInstance's optional dispose() so detachPty wiring is assertable. */
+  dispose() {
+    this.disposed = true;
   }
 
   // Test helpers - simulate PTY events
--- a/packages/server/src/lib/__tests__/pty-provider-fd-leak.test.ts
+++ b/packages/server/src/lib/__tests__/pty-provider-fd-leak.test.ts
@@ -1,0 +1,144 @@
+/**
+ * Real-fd regression test for the Bun.Terminal master-fd leak (Issue #1196).
+ *
+ * `bunTerminalProvider` wraps `Bun.spawn({ terminal: ... })`, which allocates
+ * a PTY master fd (`/dev/ptmx`) via `Bun.Terminal`. `BunTerminalPtyAdapter`
+ * did not release it on exit; `dispose()` (wired into the adapter's
+ * `fireExit()`) is the fix under test here.
+ *
+ * This file imports `bunTerminalProvider` directly (production code, no
+ * reimplementation) and spawns REAL processes -- it does not use the mocked
+ * `Bun.spawn` harness in `pty-provider.test.ts`, and must not run inside that
+ * mock's `beforeEach`/`afterEach` (kept in a separate file to guarantee
+ * that).
+ *
+ * Linux-only: relies on `/proc/self/fd` + `/dev/ptmx`, unavailable on darwin.
+ *
+ * IMPORTANT -- strong references are retained deliberately. Bun's native
+ * `Bun.Terminal` binding appears to release the fd via a GC finalizer once
+ * the JS wrapper object becomes unreachable, independent of this file's
+ * explicit `dispose()` fix (confirmed via a standalone probe: without
+ * retained references, the fd count converged to baseline even with the fix
+ * temporarily disabled, because the loop-scoped `pty` variables became
+ * GC-eligible and were finalized mid-test). Retaining every spawned adapter
+ * in `retained` for the duration of the assertion keeps them reachable, so
+ * the ONLY path that can release the fd is the explicit `dispose()` call --
+ * exactly mirroring the production shape (`InternalPtyWorker.pty` holds a
+ * live reference until `detachPty` runs). Without this, the test would
+ * pass/fail based on incidental GC timing rather than the fix.
+ *
+ * A short bounded poll (`waitForPtmxCount`) is layered on top of the
+ * retained-references technique for the kill() path specifically: killing
+ * via signal has an observed small asynchronous tail between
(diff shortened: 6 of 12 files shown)


ms2sato:
## Follow-up commit: R1/R2/R3 fixes + one additional finding

Pushed `865b041` addressing the architect's direct-audit R1/R2/R3:

- **R1 (required):** `pty-provider-fd-leak.test.ts` and `check-pty-fd-leak.ts` now count ptmx fds via a spawned child process reading `/proc/<pid>/fd` (`Bun.spawnSync`, unaffected by `mock.module('node:fs')`) instead of in-process `node:fs` calls. The memfs-detection skip path is removed entirely -- the assertions now run unconditionally under the full suite. Verified: `bun test src/ --test-name-pattern "ptmx fd leak regression"` (all 161 files loaded) shows `2 pass`, not a silent skip.
- **R2 (micro):** fixed the smoke script's opening paragraph's stale "never released by GC" claim to match the corrected understanding already stated in its own IMPORTANT paragraph and in `docs/multi-user-setup-guide.md`.
- **R3:** added a concurrent-PTY-activity hint to the kernel pty-counter assertion's failure detail.

**Additional finding surfaced while re-verifying polarity after R1:** fixing R1 exposed a second, narrower issue. In the full 161-file suite process, incidental GC of *other* test files' real PTY spawns could mask a disabled fix specifically on the kill()-path test (confirmed 3x: with `dispose()` disabled, full-suite runs showed 1 fail (natural-exit, reliable) / 1 pass (kill-path, false) -- while the same test failed correctly when run in isolation). This is a cross-test variant of the GC-masking problem the `retained` array already guards against within a single test. Fixed by forcing `Bun.gc(true)` right before each test's baseline capture, flushing unrelated garbage before measuring. Re-verified 3x in both full-suite and isolated contexts: 2 fail with the fix disabled, 2 pass with it restored, consistently.

All verification re-run clean: `bun run typecheck`, `bun run test` (full suite, exit 0), `bun run check:pty-fd-leak` (smoke, 2/2 pass), `bun run check:lang`, `bun scripts/check-source-comment-blame-shift.mjs`.

ms2sato:
## Note on CodeRabbit

GitHub-side CodeRabbit bot hit a per-developer review-limit on the R1/R2/R3 follow-up push (`865b041`): "Review limit reached ... Next review available in: 21 minutes." The prior push (`50ce5c5`) was reviewed cleanly (no actionable comments, pre-merge checks 4 passed / 1 non-actionable docstring-coverage warning). Local CLI is also unusable in this sandbox (browser-auth timeout, not rate-limit). Per `coderabbit-ops` rate-limit fallback disposition, proceeding with architect direct-audit as the primary review gate for this PR rather than waiting/retriggering.

PR #1207, opened 2026-07-18
fix: add killAsUser primitive and correct isProcessAlive cross-user detection (#1197 Part A)

ms2sato:
## Summary

Part A of Issue #1197 (multi-user orphan-worker cleanup). Fixes two coupled defects that together mean a server restart can never actually terminate an orphan worker PID spawned as a different OS user in multi-user mode:

- **Defect 2 (`isProcessAlive` EPERM misdetection):** `process.kill(pid, 0)` raises `EPERM` when the PID exists but is owned by a different OS user — exactly the cross-user orphan case. The old code treated *any* thrown error, including `EPERM`, as "process is dead," so cross-user orphans were silently judged dead and skipped forever.
- **Defect 1 (no elevated `kill` primitive):** even once correctly detected as alive, the server process cannot `process.kill()` a cross-user PID (`EPERM` again) — there was no elevated-kill primitive in `privilege-elevation.ts`.

These are bundled in one PR because defect 2 unmasking cross-user PIDs as "alive" has no effect without defect 1's elevated kill path — fixing either alone doesn't close the gap.

Defect 3 of #1197 (`SESSION_ID` env-marker sweep) is **out of scope** for this PR; it will land as a separate PR after a rebase on this one.

## Changes

- **`killAsUser(pid, signal, username, opts)`** — new strict-thin-wrapper primitive in `privilege-elevation.ts`, following the family's contract (`.claude/rules/elevation-helpers.md`): composes `kill -s <SIG> -- <pid>` at `cwd: '/'`, validates `pid` is a positive integer (throws otherwise — the injection-safety equivalent of `shellEscape` for a bare numeric argument), does not interpret the kill's exit code, and exposes a `runAsUserImpl` DI seam mirroring `rmRecursiveAsUser`.
- **`isProcessAlive`** (`process-utils.ts`) — now distinguishes `ESRCH` ("no such process" → dead) from every other outcome including `EPERM` (→ alive). See "Fail-direction rationale" below.
- **`SessionInitializationService.killOrphanWorkers`** — now `async`, takes an injected `resolveSpawnUsername` (+ optional `killAsUserImpl` DI seam). Resolves the session's spawn username once per session, gates elevation via the existing `shouldElevateForUser`, and uses `killAsUser` for the SIGTERM→SIGKILL sequence when elevated; the non-elevated branch keeps the original `processKill` call **verbatim** (caller-side pre-filter pattern per `elevation-helpers.md`).
- **`session-manager.ts`** — one-line wiring: `resolveSpawnUsername: (createdBy) => resolveSpawnUsername(createdBy, this.userRepository)`, identical in shape to the two pre-existing wirings for `WorkerLifecycleManager`/`EmbeddedAgentWorkerService`.
- **Smoke test** — `scripts/smoke/check-kill-as-user.ts`: spawns a real elevated `sleep` (via the privileged-spawn helper), kills it via `killAsUser`, confirms it's gone, and asserts a second unrelated `sleep` process is untouched (negative assertion). Runs in degenerate same-user mode when no second OS user is configured.
- **Docs** — new "Elevated orphan-worker kill check" subsection in `docs/multi-user-setup-guide.md`'s Post-deploy Verification.

## `isProcessAlive` caller audit

Full repo grep confirms **3 call sites**, all in `session-initialization-service.ts`. No call site depends on the old "EPERM ⇒ dead" behavior for correctness, so no architect consultation was needed per the AC's decision tree.

| Call site | What it checks | Category | Effect of this fix |
|---|---|---|---|
| `initializeSessions()` — `session.serverPid` check (deciding whether another live server owns the session) | The Agent Console **server process's own PID** | **Same-user confirmed** — the server process always runs as a single OS user/service account; it is never spawned per-session via `resolveSpawnUsername`. `EPERM` is not reachable here by design. | None in the intended design. (Theoretical hardening: if a PID were ever reused across a user boundary by pure OS coincidence, the new fail-safe-toward-"alive" default is still the correct direction — see rationale below.) |
| `cleanupOrphanProcesses()` — same `session.serverPid` check | Same as above | **Same-user confirmed** | Same as above |
| `killOrphanWorkers()` — `worker.pid` check (the actual per-worker kill decision) | A **worker's PTY process PID**, spawned via `resolveSpawnUsername(session.createdBy)`, which can be a different OS user than the server process in multi-user mode | **Cross-user possible — this is the fix's target** | This is the defect: previously a live cross-user worker PID read `EPERM` → treated as dead → orphan cleanup silently skipped it forever. Now it correctly reads as alive and reaches the (new) elevated kill path. |

No call site was found in the "EPERM=dead dependency" category (a caller that *needs* `EPERM` to mean "dead" for correct behavior), so this PR did not require an architect round-trip per the AC's decision tree.

## Fail-direction rationale (`isProcessAlive`'s asymmetry)

Only `ESRCH` ("no such process") means "dead" (`false`); every other outcome, including `EPERM` and any unexpected errno, is treated as "alive" (`true`). This is a deliberate fail-safe-toward-"alive" design:

- **If this is wrong and the process is actually dead:** the caller's next step is always "try to kill it," which is safe — killing an already-dead PID just fails harmlessly (tolerated by the existing try/catch/warn-and-continue structure in `killOrphanWorkers`).
- **If this is wrong the other way** (misreading a live cross-user process as "dead"): orphan cleanup would skip it forever — this is the actual bug being fixed, and it's the worse failure mode (a resource leak that never self-heals) versus the harmless no-op above.

## AC checklist

- [x] `killAsUser` primitive — strict-thin-wrapper contract, pid validation throw, cwd `/`, no ESRCH/EPERM interpretation, `runAsUserImpl` DI seam
- [x] `killOrphanWorkers` wiring — `shouldElevateForUser` gate, non-elevated branch verbatim, `killAsUserImpl` DI seam (pay-as-you-go per `elevation-helpers.md`)
- [x] `isProcessAlive` fix — ESRCH-only-dead, fail-safe-toward-alive documented
- [x] Unit tests: `killAsUser` (bypass / argv shape both signals / pid validation throws pre-spawn / timeoutMs / passthrough / default-branch smoke)
- [x] Unit tests: `isProcessAlive` (success / ESRCH / EPERM / unexpected-code flip, all polarity-verified against the pre-fix implementation)
- [x] Unit tests: `killOrphanWorkers` polarity (0/1/N orphans, EPERM-style alive-pid scenario, mixed dead/succeed/fallback set, unresolvable-`createdBy` fallback + warn condition, natural-death-race tolerance — no uncaught exception)
- [x] Smoke test `scripts/smoke/check-kill-as-user.ts` — real spawn/kill/verify + negative assertion + exit-code convention + degenerate-mode support
- [x] `docs/multi-user-setup-guide.md` Post-deploy Verification entry
- [x] Sibling coverage for the `session-manager.ts` wiring line, isolated from `SessionPauseResumeService`'s independent `resolveSpawnUsername` call via the `recoveryState: 'orphaned'` recipe (see CodeRabbit disposition #4 below)
- [x] R1 CodeRabbit findings addressed (4 fixed / 1 deferred to PR-B with rationale — see disposition table below)
- [x] `bun run typecheck` clean
- [x] `bun run test` full suite green (pasted below)
- [x] Smoke script run locally (degenerate mode; no second OS user configured in this sandbox — see output below)
- [x] `preflight-check.js` clean (all 3 production files covered)
- [x] `bun run check:lang` clean

## Verification

`bun run typecheck`: clean, zero errors (all 4 packages).

`bun run test` (full suite, post-R1-fix, `SSH_AUTH_SOCK` unset to avoid an ambient sandbox artifact unrelated to this diff — see Note below):
```
@agent-console/client   test: 2074 pass, 0 fail
@agent-console/shared   test:  592 pass, 0 fail
@agent-console/integration test: 73 pass, 0 fail
@agent-console/server   test: 3583 pass, 1 skip, 0 fail
@agent-console/embedded-agent test: 287 pass, 0 fail
scripts/ + hooks/ + skills tests: 457 pass, 0 fail
TEST_EXIT: 0
```

Smoke test (degenerate same-user mode — this sandbox has no second configured OS user for elevation; post-R1-fix, now with cleanup verification as its own assertion):
```
$ bun scripts/smoke/check-kill-as-user.ts "$(whoami)"
  WARN  target user equals the server-process user; spawnAsUser/killAsUser will bypass elevation (degenerate same-user mode)...
==> spawning tracked target sleep ... target sleep pid: 3099032
==> spawning tracked CONTROL sleep ... control sleep pid: 3099033
  OK    target pid is alive before killAsUser
  OK    control pid is alive before killAsUser
==> killAsUser(3099032, 'SIGTERM', ...)
  OK    killAsUser SIGTERM exitCode === 0
  OK    target pid is gone after killAsUser SIGTERM
  OK    control pid is still alive after killAsUser on the target pid
==> cleanup (best-effort SIGKILL of any surviving tracked pids, verified)
  OK    cleanup left no surviving process for pid=3099033
PASSED: 6 assertion(s) passed
SMOKE_EXIT: 0
```

Also manually verified the new timeout path (temporarily forcing a short deadline in a scratch copy, not committed): a stalled PID-read correctly produces `PROBE TIMEOUT: reader.read() timed out after 50ms` and exits with code 2, distinct from an assertion failure.

`node .claude/skills/orchestrator/preflight-check.js`: all 3 production files covered, no rule/skill duplication, language check clean, no blame-shift references.

**Note (pre-existing, unrelated):** with the ambient `SSH_AUTH_SOCK` set (this sandbox has a real agent socket in the environment), `multi-user-mode.test.ts`'s `Issue #918` elevation-skip test fails on an unrelated env-leak assertion. This file is untouched by this diff; it passes 37/37 standalone with `SSH_AUTH_SOCK` unset. Matches known ambient-leakage behavior for dogfood sandboxes.

### Accepted limitation (addressed by PR-B, see #1197)

`killAsUser` returns exit 0 on signal delivery, not on termination confirmation. A worker ignoring `SIGTERM` after this PR would be counted in `killedCount` (a log-only counter) and could remain live after `killOrphanWorkers` returns. This is byte-parity with the pre-existing `processKill` code path and is intentional scope for this PR (cross-user parity, not kill-contract hardening).

The tree-wide remedy — TERM → grace → KILL with liveness re-check at the same startup point — is structurally covered by PR-B's `SESSION_ID` env-marker sweep (see Issue #1197). PR-B's test matrix will include a `trap '' TERM` polarity case verifying `SIGTERM`-ignoring survivors are collected via `SIGKILL`.

## CodeRabbit round-1 review disposition

CodeRabbit's GitHub-side bot posted 4 inline findings after the rate-limit window cleared. Architect-reviewed disposition, 4 fixed in this PR / 1 deferred to PR-B:

| # | Severity | Finding | Disposition |
|---|---|---|---|
| 1a | 🟠 Major | `killAsUser` exit 0 confirms signal delivery, not termination; `killedCount` increments before confirming the worker actually died | **Deferred to PR-B** — pre-existing `processKill` contract, not a regression from this PR; see "Accepted limitation" above |
| 1b | 🟠 Major | `killOrphanWorkers`'s elevated `killAsUser` calls omitted `timeoutMs`, risking unbounded startup block on a stalled elevated-login-shell/NSS chain | **Fixed** — explicit `KILL_AS_USER_TIMEOUT_MS = 10_000` on both SIGTERM/SIGKILL calls; `result.timedOut` now treated as failure (escalates SIGTERM→SIGKILL, warn-and-continue on SIGKILL). 2 new tests: timeoutMs forwarding + timedOut-with-exitCode-0 polarity (proves the `timedOut` check is not redundant with the `exitCode !== 0` check) |
| 2 | 🟠 Major | Smoke script's `reader.read()` had no deadline; both smoke `killAsUser` calls also omitted `timeoutMs` — could hang the deploy instead of failing cleanly | **Fixed** — bounded PID-read with reader cancellation on expiry, explicit `timeoutMs` on both smoke `killAsUser` calls, any phase timeout now exits distinctly with code 2 (environment problem) via a dedicated `ProbeTimeoutError`. Verified locally: timeout path produces exit 2 with a clear message; normal path still passes 6/6 |
| 3 | 🟡 Minor | Smoke's cleanup loop only caught thrown exceptions; `killAsUser` resolves normally on non-zero exit / timeout, so most cleanup failures went undetected | **Fixed** — cleanup now inspects `result.timedOut`/`result.exitCode` explicitly, calls `waitUntilGone` after SIGKILL, and records a failure (not a silent success) if the tracked PID survives |
| 4 | 🟡 Minor | `session-manager.test.ts`'s wiring test could pass even if the new resolver ignores `this.userRepository`, because `SessionPauseResumeService`'s independent auto-resume path also calls `findById` for the same createdBy | **Fixed** — rewrote using CodeRabbit's suggested recipe: persist the session with `recoveryState: 'orphaned'` (preserved-but-never-auto-resumed by `initializeSessions()`, picked up only by the later `cleanupOrphanProcesses()` pass), which isolates the assertion to exactly one `resolveSpawnUsername` call. Verified via TDD polarity (breaks cleanly to `toEqual([])` when the wiring is deliberately broken, passes precisely `toEqual(['user-alice-uuid'])` otherwise) |

Note on #4's original transparency comment: my initial PR description characterized the confound as a "structural reality" that couldn't be isolated without disproportionate mocking. CodeRabbit's recipe (persist with `recoveryState: 'orphaned'`) proved that wrong — the isolation IS achievable via the existing repository API, just not via the `update()` field I originally reached for (see the incidental finding below). Revised accordingly.

**Incidental finding (out of scope, not fixed here — filed as #1210):** while implementing the #4 fix, discovered that `JsonSessionRepository.update()` (`packages/server/src/repositories/json-session-repository.ts`) does not support the `recoveryState` / `orphanedAt` / `orphanedReason` fields at all — it silently no-ops on them while still returning `true`. `SqliteSessionRepository.update()` handles them correctly. Since `session-initialization-service.ts`'s `detectOrphans()` calls `sessionRepository.update(id, { recoveryState: 'orphaned', ... })` for exactly this purpose, any deployment using the JSON repository backend would have orphan-marking silently fail to persist. Worked around in the test via `findById` + `save()` (a full-record replace, which IS supported) rather than `update()`. Follow-up filed as [Issue #1210](https://github.com/ms2sato/agent-console/issues/1210) — not fixed in this PR as it's unrelated to #1197's scope.

## Not in scope

- Defect 3 of #1197 (`SESSION_ID` env-marker sweep) — separate PR after rebasing on this one.
- Real cross-user elevation smoke verification (the actual OS-user-boundary crossing) — this sandbox has no second configured OS user; the smoke script itself is verified logically sound end-to-end in degenerate mode and should be re-run against a real multi-user deploy per the doc's Post-deploy Verification instructions.
- `JsonSessionRepository`'s `recoveryState`/`orphanedAt`/`orphanedReason` update gap (see incidental finding above, filed as [#1210](https://github.com/ms2sato/agent-console/issues/1210)) — unrelated to this PR's scope.

Related: #1197 (umbrella — not closed by this PR; other defects tracked separately)

## Note on CodeRabbit

The local CodeRabbit CLI timed out without returning a result during initial PR creation. The GitHub-side bot was rate-limited on PR open, but its review completed after the stated cooldown and posted the 4 inline findings addressed in the "CodeRabbit round-1 review disposition" section above.

**Round-2 (verifying the R1 fix commit `c37e6dc`) is unavailable from either source:**
- Local CLI: retried with a longer timeout specifically for round-2 verification — failed with `authentication_failed` / `automatic_login_failed` (not a rate-limit; interactive browser auth cannot complete in this headless worktree sandbox).
- GitHub bot: two manual `@coderabbitai review` retriggers (after the original rate-limit cooldown, and again after the R1 fix push) both returned "Review finished... CodeRabbit is an incremental review system and does not re-review already reviewed commits. This command is applicable only when automatic reviews are paused" — no new walkthrough, no new `reviewDecision`, no new inline comments were posted for commit `c37e6dc`. No further automatic incremental review arrived after 30+ minutes.

This matches the "both simultaneously unavailable" disposition in the `coderabbit-ops` skill's troubleshooting guide — CodeRabbit's independent verification of the round-1 fix is unobtainable from either channel for this PR. The round-1 findings that WERE obtained are addressed above with a documented disposition (4 fixed with new tests, 1 explicitly deferred to PR-B with rationale). Merge decision proceeds under existing [PR Merge Authority](https://github.com/ms2sato/agent-console/blob/main/.claude/skills/orchestrator/SKILL.md#pr-merge-authority) (Architect + Orchestrator dual-clean), not taken here.

## Summary by CodeRabbit

* **New Features**
  * Added multi-user support for safely stopping orphaned worker processes under the appropriate operating-system account.
  * Added fallback termination when graceful shutdown does not succeed.
  * Added a production smoke test for verifying targeted process termination without affecting unrelated processes.

* **Bug Fixes**
  * Process checks now treat permission-related errors as “process still alive” rather than incorrectly reporting termination.

* **Documentation**
  * Expanded the multi-user setup guide with post-deployment validation steps and expected outcomes.

### DIFF
--- a/docs/multi-user-setup-guide.md
+++ b/docs/multi-user-setup-guide.md
@@ -1108,6 +1108,53 @@ Passing the current process user as `<target-user>` runs in a degenerate
 same-user mode where `spawnAsUser` bypasses elevation — this still exercises
 the full Bash tool-call round trip except the actual OS-boundary crossing.
 
+### Elevated orphan-worker kill check
+
+Run after every deploy that touches
+`packages/server/src/services/privilege-elevation.ts`'s `killAsUser` or
+`SessionInitializationService.killOrphanWorkers` (Issue #1197 Part A). Prior
+to this helper, a server restart could not actually terminate an orphaned
+worker PID that was spawned as a different OS user in multi-user mode —
+`process.kill` raises `EPERM` for a cross-user PID, and (before the
+companion `isProcessAlive` fix) that `EPERM` was misread as "already dead",
+so the orphan was silently left running forever:
+
+```bash
+sudo -u agentconsole bun scripts/smoke/check-kill-as-user.ts <target-user>
+```
+
+What it verifies (against the **real** machine — real `sudo`, real target
+process):
+
+- `spawnAsUser` launches a real, long-lived `sleep` as `<target-user>`, and
+  the smoke captures that process's own PID (not the outer `sudo`/`sh`
+  wrapper PID `subprocess.pid` would report when elevated).
+- `killAsUser` sends a real elevated `SIGTERM` to that PID and the process
+  actually terminates (polled via `/proc/<pid>` existence, bounded to 5s).
+- Negative: a second, unrelated `sleep` process spawned as the same target
+  user survives the `killAsUser` call against the first one — proof the
+  helper signals exactly the targeted PID (`kill -s <SIG> -- <pid>`), not
+  something broader like a name-based `pkill`.
+
+This smoke does NOT exercise `killAsUser`'s SIGTERM → SIGKILL fallback
+orchestration (that lives in the caller,
+`SessionInitializationService.killOrphanWorkers`, and is covered by unit
+tests with an injected fake) or `isProcessAlive`'s ESRCH/EPERM distinction
+(covered by `packages/server/src/lib/__tests__/process-utils.test.ts`,
+which spies on `process.kill` directly and needs no second OS user).
+
+Exit codes: `0` all assertions passed, `1` an assertion failed (the system
+is wrong), `2` bad usage or the smoke could not run (missing target-user
+argument, spawn-launch failure).
+
+Passing the current process user as `<target-user>` runs in a degenerate
+same-user mode where `spawnAsUser` / `killAsUser` bypass elevation — this
+still exercises the full mechanism (spawn, PID capture, kill, negative
+assertion) except the actual `sudo` OS-user-boundary crossing. Unlike the
+smokes above, this degenerate mode does not require `AUTH_MODE=multi-user`
+to be meaningful — the elevation bypass is evaluated independently of it,
+based solely on whether `<target-user>` equals the server-process user.
+
 ### PTY master fd leak check
 
 Run after every deploy that touches `packages/server/src/lib/pty-provider.ts`
--- a/packages/server/src/lib/__tests__/process-utils.test.ts
+++ b/packages/server/src/lib/__tests__/process-utils.test.ts
@@ -1,0 +1,82 @@
+import { describe, it, expect, spyOn } from 'bun:test';
+
+/**
+ * `mock-process-helper.ts` (used by several service tests, e.g.
+ * `session-initialization-service.test.ts`) registers a process-global
+ * `mock.module()` override for this exact file's resolved absolute path
+ * (see `.claude/rules/testing.md` Anti-Pattern #2: "mock.module() is
+ * process-global in bun:test and pollutes all test files in the same
+ * process"). When the full suite runs in one `bun test` invocation, a
+ * plain `import { isProcessAlive } from '../process-utils.js'` here would
+ * resolve to that OTHER file's fake (`alivePids.has(pid)`), not the real
+ * implementation under test -- silently defeating every assertion below
+ * regardless of the `process.kill` spy.
+ *
+ * A dynamic `import()` with a distinguishing query string resolves to a
+ * fresh module specifier that `mock.module()`'s exact-path registration
+ * does not match, so it always loads the real, unmocked module. Verified
+ * empirically: this suite passes in isolation AND alongside
+ * `session-initialization-service.test.ts` in the same invocation.
+ */
+let loadCounter = 0;
+async function loadRealProcessUtils(): Promise<typeof import('../process-utils.js')> {
+  // Template literal (not a string literal) so TypeScript treats the
+  // specifier as dynamic and does not attempt static module resolution on
+  // the query-string suffix; the explicit return type annotation above
+  // still gives callers the real module's typed shape.
+  return import(`../process-utils.js?real-impl-under-test=${++loadCounter}`);
+}
+
+function makeErrno(code: string): NodeJS.ErrnoException {
+  const err = new Error(`mock process.kill error: ${code}`) as NodeJS.ErrnoException;
+  err.code = code;
+  return err;
+}
+
+describe('isProcessAlive', () => {
+  it('returns true when process.kill succeeds', async () => {
+    const { isProcessAlive } = await loadRealProcessUtils();
+    const spy = spyOn(process, 'kill').mockImplementation(() => true);
+    try {
+      expect(isProcessAlive(12345)).toBe(true);
+    } finally {
+      spy.mockRestore();
+    }
+  });
+
+  it('returns false when process.kill throws ESRCH (no such process)', async () => {
+    const { isProcessAlive } = await loadRealProcessUtils();
+    const spy = spyOn(process, 'kill').mockImplementation(() => {
+      throw makeErrno('ESRCH');
+    });
+    try {
+      expect(isProcessAlive(12345)).toBe(false);
+    } finally {
+      spy.mockRestore();
+    }
+  });
+
+  it('returns true when process.kill throws EPERM (process exists, owned by another user)', async () => {
+    const { isProcessAlive } = await loadRealProcessUtils();
+    const spy = spyOn(process, 'kill').mockImplementation(() => {
+      throw makeErrno('EPERM');
+    });
+    try {
+      expect(isProcessAlive(12345)).toBe(true);
+    } finally {
+      spy.mockRestore();
+    }
+  });
+
+  it('returns true (fail-safe default) when process.kill throws an unexpected error code', async () => {
+    const { isProcessAlive } = await loadRealProcessUtils();
+    const spy = spyOn(process, 'kill').mockImplementation(() => {
+      throw makeErrno('EINVAL');
+    });
+    try {
+      expect(isProcessAlive(12345)).toBe(true);
+    } finally {
+      spy.mockRestore();
+    }
+  });
+});
--- a/packages/server/src/lib/process-utils.ts
+++ b/packages/server/src/lib/process-utils.ts
@@ -4,6 +4,7 @@
  * These wrappers allow tests to mock process operations using mock.module()
  * without modifying global process object.
  */
+import { isErrnoException } from './type-guards.js';
 
 /**
  * Kill a process by PID.
@@ -14,12 +15,28 @@ export function processKill(pid: number, signal?: NodeJS.Signals | number): bool
 
 /**
  * Check if a process is alive (exists).
+ *
+ * Fail-safe-toward-"alive" asymmetry: `process.kill(pid, 0)` throws `EPERM`
+ * when the process exists but is owned by a different OS user -- exactly
+ * the case multi-user orphan-worker cleanup needs to detect (a worker's PID
+ * spawned via `resolveSpawnUsername` under a different OS user than the
+ * server process). Only `ESRCH` ("no such process") means the process is
+ * actually dead; every other outcome, including `EPERM` and any unexpected
+ * error, is treated as "alive". If this is wrong and the process is
+ * actually dead, the caller's next step is always "try to kill it", which
+ * is safe (killing an already-dead pid just fails harmlessly). If this is
+ * wrong the other way -- misreading a live cross-user process as "dead" --
+ * orphan cleanup would skip it forever, which is the bug this asymmetry
+ * fixes.
  */
 export function isProcessAlive(pid: number): boolean {
   try {
     process.kill(pid, 0);
     return true;
-  } catch {
-    return false;
+  } catch (err) {
+    if (isErrnoException(err) && err.code === 'ESRCH') {
+      return false;
+    }
+    return true;
   }
 }
--- a/packages/server/src/services/__tests__/privilege-elevation.test.ts
+++ b/packages/server/src/services/__tests__/privilege-elevation.test.ts
@@ -5,6 +5,7 @@ import {
   runAsUser,
   rmRecursiveAsUser,
   writeUserOwnedSecretFile,
+  killAsUser,
   spawnAsUser,
   buildSpawnArgs,
   buildInnerCommand,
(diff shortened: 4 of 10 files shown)


ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

ms2sato:
## Cross-user smoke output (architect required followup, dogfood host)

Executed by owner on the dogfood host against R1 branch (commit `c37e6dc`, worktree `wt-004-9csx`):

```
$ sudo -u agentconsole bash -c 'cd /var/lib/agent-console/repositories/ms2sato/agent-console/worktrees/wt-004-9csx && /home/agentconsole/.bun/bin/bun scripts/smoke/check-kill-as-user.ts ms2sato'

==> spawning tracked target sleep as 'ms2sato' (elevated: true)
  target sleep pid: 3326097
==> spawning tracked CONTROL sleep as 'ms2sato' (negative-assertion witness)
  control sleep pid: 3326104
==> sanity: both pids alive before killAsUser
  OK    target pid is alive before killAsUser
  OK    control pid is alive before killAsUser
==> killAsUser(3326097, 'SIGTERM', 'ms2sato')
[01:40:29.643] INFO: spawnAsUser spawned
    service: "service:privilege-elevation"
    username: "ms2sato"
    elevated: true
[01:40:29.664] INFO: spawnAsUser spawned
    service: "service:privilege-elevation"
    username: "ms2sato"
    elevated: true
  OK    killAsUser SIGTERM exitCode === 0
==> waiting for target pid to exit (bounded poll, up to 5s)
  OK    target pid is gone after killAsUser SIGTERM
==> negative assertion: control pid must still be alive (killAsUser targeted only the intended pid)
  OK    control pid is still alive after killAsUser on the target pid
==> cleanup (best-effort SIGKILL of any surviving tracked pids, verified)
[01:40:29.699] INFO: runAsUser completed
    service: "service:privilege-elevation"
    username: "ms2sato"
    elevated: true
    exitCode: 0
    timedOut: false
  OK    cleanup left no surviving process for pid=3326104
PASSED: 6 assertion(s) passed
[01:40:29.717] INFO: runAsUser completed
    service: "service:privilege-elevation"
    username: "ms2sato"
    elevated: true
    exitCode: 0
    timedOut: false
```

**Result: 6/6 assertions passed, cross-user elevation path (`elevated: true`) exercised end-to-end.** killAsUser SIGTERM correctly terminated only the intended target pid; the negative-assertion control pid remained alive; cleanup left no leftovers. Architect's os-environment-coupling pre-merge requirement satisfied.

Note: agentconsole's default PATH does not include bun; ran with the explicit full path `/home/agentconsole/.bun/bin/bun`. This is a host-config detail (not a script defect), operator-noted for future runs.

PR #1219, opened 2026-07-19
fix: sweep SESSION_ID-marked orphan processes tree-wide on restart (#1197 Part B)

### LINKED ISSUE #1197: [umbrella] multi-user orphan cleanup hardening (isProcessAlive EPERM / elevation-aware kill / SESSION_ID tree cleanup)

ms2sato:
## Summary

Issue #1197 Part B (follow-up to Part A, PR #1207 / merged 410dcd0): adds a tree-wide SESSION_ID env-marker sweep that catches orphan processes `SessionInitializationService.killOrphanWorkers` cannot see.

**Scope note (owner-approved)**: This PR incidentally includes a 1-line `.github/workflows/language-lint.yml` fix (adding a `bun install --frozen-lockfile` step). This is unrelated to the #1197 sweep logic — it addresses a CI regression PR #1220 introduced when `scripts/schema-source-normalize.mjs` started depending on the `typescript` package in a workflow that never installed devDeps in the first place (confirmed by reproducing the exact CI failure against a clean `origin/main` checkout with no dependency on this PR's own diff). Same-PR inclusion approved by the owner to avoid a separate infra-fix round-trip; see the "CI Language Lint infra fix" commit for the isolated change.

- `killOrphanWorkers` only kills processes tracked via a session's persisted `worker.pid` — the direct PTY wrapper process. Detached/background descendant processes (a worker's `bun run dev` child, an MCP subprocess, ...) are never in that PID set and leak forever across server restarts.
- Every worker process already carries `AGENT_CONSOLE_SESSION_ID=<sessionId>` in its environment (set at PTY spawn time in `user-mode.ts`, inherited by every descendant). New module `packages/server/src/services/orphan-process-sweeper.ts` scans `/proc/[pid]/environ` for that marker, tree-wide, and kills anything matching, regardless of whether it was ever tracked as a `worker.pid`.
- **Multi-user constraint**: `/proc/<pid>/environ` for another OS user's process is not readable by the server process (EACCES) — the scan (which requires reading `environ` to match) cannot be split into "scan as server, then kill as elevated user". A single POSIX `sh` script does scan + match + kill in one invocation, executed via `runAsUser` as the resolved session owner (same elevation pattern as `killAsUser` / `rmRecursiveAsUser`, i.e. an elevated login shell via the existing elevation helper).
- Wired into `SessionInitializationService.sweepSessionProcesses` (new static method), called unconditionally at both existing `killOrphanWorkers` call sites (`initializeSessions()`, `cleanupOrphanProcesses()`) — startup-only, best-effort (never throws, logs a warn on failure since `killOrphanWorkers` remains the primary, already-tested cleanup mechanism).

**Explicitly out of scope** (per Architect AC): `deleteSession`-time cleanup and `shutdownAppContext` are NOT wired to this sweep — both already know precisely which pids they're tearing down (no discovery problem to solve).

## `trap '' TERM` polarity case (Architect audit amend)

`packages/server/src/services/__tests__/orphan-process-sweeper.test.ts` includes a required real-process test (`escalates to SIGKILL for a marked process that ignores SIGTERM (trap TERM)`) that spawns a real `sh -c "trap '' TERM; while :; do :; done"` process with the marker set, runs the sweep script directly via `Bun.spawn` (same-user, no elevation needed to exercise the shell logic itself), asserts the process survives the TERM-only grace window, then is reaped by the SIGKILL escalation. This proves the escalation branch actually fires against a SIGTERM-immune target, not just that SIGTERM alone happens to work.

## Test coverage

- **Tier A (pure/fast)**: `buildSweepScript` script-shape assertions (shell-escaping/injection safety, TERM-before-KILL ordering, `SWEEP_PROC_ROOT` override, self-pid exclusion), `parseSweptCount` boundary matrix, `sweepOrphanProcesses` via injected `runAsUserImpl` fake (forwarding, non-throw on non-zero exit/timeout, empty-sessionId guard).
- **Tier B (real-process, same-user, no elevation)**: 0/1/N candidates, unmarked-process survives, different-session-id survives, `trap '' TERM` escalation (above), self-exclusion (script's own env legitimately carries the marker it searches for — a broken self-exclusion would self-SIGTERM and never print output), environ-unreadable/vanished entries don't crash the scan (via a `SWEEP_PROC_ROOT` override pointed at a scratch dir with ENOENT/EACCES fixtures alongside one real matching entry).
- `SessionInitializationService` gains `sweepSessionProcesses` unit tests: called once per session at both call sites with the resolved username, **0-sessions → 0 sweep-elevation calls** boundary, non-throw on sweep failure.
- Smoke: `scripts/smoke/check-orphan-sweep.ts` (real machine, real elevated login shell, real cross-user `/proc/<pid>/environ` read) — new marker-discovery smoke distinct from Part A's `check-kill-as-user.ts` (which only signals an already-known pid and never reads `environ`). Ran locally in degenerate same-user mode: 7/7 assertions passed, exit 0.

## Q10 (schema-type parallel maintenance) — N/A

No wire/shared-type change. This is a pure server-internal mechanism (no client-facing schema, no new shared type crossing the server/client boundary). Confirmed no `packages/shared/src/schemas/` or `packages/shared/src/types/` touched.

## Glossary trigger check

Walked `.claude/rules/glossary-maintenance.md`'s trigger list: no new design doc, no new type/interface crossing a wire boundary, no new DB field, no new API/MCP param, no Terminology-section edit. `orphan-process-sweeper.ts` is an internal helper extension within the already-documented elevation-helper family — same judgment as Part A's `killAsUser`, which also did not touch the glossary. Not a glossary trigger.

## Merge disposition note (2026-07-19)

**CR state**: local CLI hit the known headless-worktree auth failure (`automatic_login_failed` — a structural limitation of a delegated worktree session with no browser, not a rate-limit; same shape as PRs #1204/#1207/#1208/#1209/#1212 this sprint). The GitHub-side bot's initial walkthrough at 03:27 UTC was a rate-limit warning (next review in 44 min), and a manual `@coderabbitai review` retrigger at 03:33 UTC got the "Review finished... does not re-review already reviewed commits" quirk reply with no actual code review content — i.e. the only "walkthrough" for this commit is the rate-limit message itself (the rate-limit-only-walkthrough case, not the walkthrough-exists clean-equivalent case). inline comments: 0, reviewDecision: empty.

**Fallback disposition applied** (`coderabbit-ops` skill, headless-auth-unavailable + GitHub-bot-rate-limited-with-quirk): both CodeRabbit channels are unavailable for this commit. Substituting Architect independent audit + Orchestrator acceptance as the dual-clean gate. Architect audit verdict: **CLEAN-WITH-FOLLOWUPS** — no blocking findings; confirmed the single-elevated-invocation design, the strict-thin-wrapper contract, the PID-reuse mitigation, the required `trap '' TERM` escalation test, the startup-only wiring scope, and the smoke script's cross-user `/proc/<pid>/environ` read-permission coverage all match the AC. One non-blocking follow-up noted: `resolveSpawnUsername` is resolved independently by `killOrphanWorkers` and `sweepSessionProcesses` at the same call site (2 DB lookups instead of 1 per dead-server session at startup) — acceptable since it's a cold startup path, not a hot path, and the duplication is deliberately tested rather than accidental; worth a low-priority follow-up Issue if it's ever found to matter at scale.

All CI checks green (test, CodeQL x2, blame-shift lint, language-lint, preflight, structural metrics).

**Owner-run real multi-user smoke on the dogfood host remains a merge precondition** for this specific PR (per `.claude/rules/os-environment-coupling.md` pre-merge discipline, same as Part A) — requesting this explicitly; merging is deferred until that positive+negative result is confirmed.

## Update (2026-07-22): owner smoke found a real bug, fixed

The owner's real multi-user smoke on the dogfood host FAILED with `sh: 1: Syntax error: "do" unexpected`. Root cause: the elevated `sudo -u <user> ... -i sh -c <command>` path re-joins its trailing argv into a single line before the target login shell re-parses it, collapsing the sweep script's embedded newlines and breaking its `for ... do ... done` loop. This never surfaced in local/degenerate-mode testing because the non-elevated path execs directly (no re-join). Fix (Architect-directed): keep the elevated `command` fixed to the single-line `sh -s` and deliver the actual multi-line script over `stdin` instead — a channel immune to the argv re-join. New TDD-polarity unit test added; `.claude/rules/elevation-helpers.md` gained a new "Multi-line elevated commands" subsection documenting the general pattern. Re-requesting the owner smoke against this fix.

Separately, pushing this fix surfaced an unrelated main-side CI regression (`.github/workflows/language-lint.yml` never had a `bun install` step; PR #1220 added a script dependency on the `typescript` package that only that workflow's test suite exercises) — see the Scope note above for the 1-line fix included in this PR.

## CodeRabbit finding disposition (2026-07-22, commit 75b207cc)

CodeRabbit's GitHub-side bot successfully reviewed this commit (not rate-limited this time) and posted 1 Trivial/Nitpick finding: sequential per-session elevated `/proc` scans in `sweepSessionProcesses` could add latency after a crash with many dead sessions (one `sudo` spawn + full-`/proc` scan per session, serialized). **Deferring, not fixing** — same disposition class as the already-noted `resolveSpawnUsername` double-lookup: this is a startup-only, best-effort, cold path (not a hot path), and batching would add real complexity (multi-ID `grep -F -f`) for a scenario (many simultaneous dead sessions after a crash) that is itself already a rare, degraded-startup case where a few extra seconds of sequential sweeping is an acceptable trade-off against the correctness this sweep buys. Worth reconsidering only if this is ever observed to matter in practice.

All CI checks green (test, CodeQL x2, blame-shift lint, language-lint, preflight, structural metrics) as of commit `dd87528d`.

## Test plan

- [x] `bun run typecheck` — exit 0
- [x] `bun run test` (full monorepo suite) — all packages pass; server package's only failure is the pre-existing, environment-specific `Issue #918: elevation-skip (direct) path ... SSH_AUTH_SOCK` test (confirmed identical on unmodified `origin/main` via `git stash` baseline comparison — unrelated to this diff)
- [x] `node .claude/skills/orchestrator/preflight-check.js` — all checks green (coverage, rule/skill duplication, language, blame-shift)
- [x] `bun run check:lang` — exit 0
- [x] `bun scripts/smoke/check-orphan-sweep.ts "$(whoami)"` (degenerate same-user mode) — exit 0, 7/7 assertions
- [ ] Real multi-user smoke on dogfood host (owner-run, merge precondition — see above)

Closes #1197

## Summary by CodeRabbit

- **New Features**
  - Added startup and cleanup sweeps to find and terminate orphaned descendant processes associated with sessions.
  - Added safeguards to avoid affecting unrelated processes and to escalate termination when needed.
  - Added a post-deployment smoke test for cross-user orphan-process cleanup.

- **Documentation**
  - Documented elevated command execution requirements and the new orphan-process smoke test.

- **Tests**
  - Added comprehensive unit, integration, error-handling, and process-cleanup coverage.

### DIFF
--- a/.claude/rules/elevation-helpers.md
+++ b/.claude/rules/elevation-helpers.md
@@ -132,6 +132,26 @@ The reciprocal cost of preserving two branches is a small per-caller `if (should
 - Any caller currently doing `Bun.spawn(['sh','-c', cmd], { env: getCleanChildProcessEnv() })` (or analogous env-shape transformation) that is about to gain a multi-user elevation path. Check `getCleanChildProcessEnv` usage in `packages/server/src/`.
 - Any caller doing pre-spawn arg sanitization (path normalization, secret stripping, etc.) where the sanitization is not encoded in the spawn primitive.
 
+### Multi-line elevated commands: deliver via stdin, not argv
+
+When `runAsUser`'s elevated branch runs (`shouldElevateForUser` true), the argv it constructs is `['sudo', '-u', <user>, --preserve-env=..., '-i', 'sh', '-c', <innerCommand>]`. The target user's login shell (invoked via `-i` with trailing arguments) reconstructs its own command line by joining those trailing arguments with spaces before re-parsing them -- this collapses any literal newline characters embedded inside `innerCommand` into plain whitespace. A multi-statement shell script that relies on newlines as statement separators (e.g. a `for ... do ... done` loop written across multiple lines) breaks with a syntax error once every line lands on one collapsed line (observed on a real host: `sh: 1: Syntax error: "do" unexpected`). The non-elevated branch (`['sh', '-c', command]`, executed directly via `execve`, no intermediate login-shell re-join) does not have this problem -- the bug is specific to the elevated path.
+
+**Rule**: a `command` string passed to `runAsUser` / `spawnAsUser` must stay single-line whenever the caller cannot guarantee it will only ever run non-elevated. If the actual payload is a multi-line script, keep `command` fixed to the single-line `'sh -s'` (POSIX `sh`'s "read the script from stdin" invocation) and deliver the script's bytes via `opts.stdin` instead of embedding them in `command`. `opts.stdin` travels as a separate pipe, never touches argv, and is therefore immune to the elevated login shell's argv-join.
+
+```ts
+// Wrong: a multi-line command breaks once elevated (newlines collapse).
+await runAsUser({ username, command: multiLineScript, cwd: '/' });
+
+// Right: single-line command, script delivered over stdin.
+await runAsUser({ username, command: 'sh -s', stdin: multiLineScript, cwd: '/' });
+```
+
+This is safe for the non-elevated (or bypassed-elevation, e.g. `username` equals the server-process user) branch too -- `sh -c "sh -s"` with the script piped on stdin behaves identically to running the script directly, just through one extra layer of `sh`. No branching is needed between elevated and non-elevated cases; always use the `command: 'sh -s'` + `stdin: <script>` shape once the payload is multi-line.
+
+**Worked example**: `packages/server/src/services/orphan-process-sweeper.ts`'s `sweepOrphanProcesses` composes a multi-line POSIX `sh` script (`buildSweepScript`) to scan+match+kill orphan processes in one elevated invocation. The Issue #1197 Part B owner smoke on a real multi-user host caught the argv-join collapsing the script's `for` loop; the fix switched `command` from the raw multi-line script to `'sh -s'` with the script delivered via `stdin`, with no change to the script's own content.
+
+**Proactive audit candidates** (the bug shape, not current sites): any future caller that composes a multi-statement shell script (more than a single `cmd1 && cmd2` chain) and passes it directly as `command` to `runAsUser` / `spawnAsUser`. Single-line `command` strings (the overwhelming majority of existing call sites: `rm -rf --`, `kill -s`, `gh ...`, `git ...`) are unaffected and need no change.
+
 ## How this rule is expected to evolve
 
 As the helper family grows (`chmodAsUser`, `lstatAsUser`, etc. are plausible future additions when fs operations on cross-user paths come up), the naming pattern (`<verb>AsUser`) and the strict-thin-wrapper contract self-extend. This rule's body should not need to change; the glossary entry's enumeration of helpers should be updated in the same PR that adds the new primitive.
@@ -149,6 +169,7 @@ If a consumer surfaces a recurring need for semantic layering (e.g. "every calle
   - [PR #881](https://github.com/ms2sato/agent-console/pull/881) (Issues #869 / #870) — `lib/git.ts` consolidation; `gitExec`-routed helpers accept `requestUser`.
   - [PR #877](https://github.com/ms2sato/agent-console/pull/877) (Issue #876) — `delegate_to_worktree` resolves `parentSession.createdBy → osUsername`, hoist pattern.
   - [PR #880](https://github.com/ms2sato/agent-console/pull/880) (Issue #879) — `run_process` elevated via `spawnAsUser`.
+  - [PR #1219](https://github.com/ms2sato/agent-console/pull/1219) (Issue #1197 Part B) — `orphan-process-sweeper.ts`'s multi-line sweep script; discovered and fixed the argv-join-collapses-newlines bug this rule's "Multi-line elevated commands" subsection documents.
 - **Related Issues** (multi-PR convergence drivers):
   - [#871](https://github.com/ms2sato/agent-console/issues/871) — unregister-repository error surfacing (separate identity audit).
   - [#878](https://github.com/ms2sato/agent-console/issues/878) — MCP auth-boundary (caller-claimed `sessionId` verification); explicit out-of-scope for elevation PRs, addressed horizontally.
--- a/.github/workflows/language-lint.yml
+++ b/.github/workflows/language-lint.yml
@@ -39,6 +39,9 @@ jobs:
         with:
           bun-version: "1.3.5"
 
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
       - name: Run language check
         run: bun run check:lang
 
--- a/docs/multi-user-setup-guide.md
+++ b/docs/multi-user-setup-guide.md
@@ -1155,6 +1155,61 @@ smokes above, this degenerate mode does not require `AUTH_MODE=multi-user`
 to be meaningful — the elevation bypass is evaluated independently of it,
 based solely on whether `<target-user>` equals the server-process user.
 
+### SESSION_ID marker orphan-process sweep
+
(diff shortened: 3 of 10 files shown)


ms2sato:
@coderabbitai review

PR #1220, opened 2026-07-19
chore(schemas): normalize comment-only diffs out of SCHEMA_VERSION hash

### LINKED ISSUE #1190: chore(schemas): normalize comment-only diff in schema-version.gen.ts hash regen to avoid spurious wire-version bumps
Surfaced by architect review during PR #1188 (Issue #898 comment cleanup) audit.

## Observation

`packages/shared/src/schema-version.gen.ts` embeds a hash of the shared schemas source. Currently that hash is computed over the raw source, **including comments**. This means a comment-only PR that touches any `packages/shared/src/schemas/*.ts` bumps `SCHEMA_VERSION`, which propagates via `wire-version` handshake and triggers a one-time reload for every connected client with the previous bundle.

The reload behavior is deliberate for real schema changes (protecting against wire drift). For **comment-only** edits — CLAUDE.md drift cleanup, English-Japanese docstring adjustments, whitespace — the reload is unnecessary friction.

## Proposed fix

Normalize the input to the hash by stripping comments and blank/whitespace-only lines before hashing:

- Strip single-line comments (`// ...` up to newline; block `/* ... */`)
- Strip blank lines / trim trailing whitespace
- Preserve string literals (do NOT strip comment-like characters inside `\"...\"` / `'...'` / `\` ... \``)
- Hash the normalized token stream

Regenerate `schema-version.gen.ts` from the normalized input.

## Non-goals

- Do NOT strip semantic whitespace (JSON string contents, template literal expressions)
- Do NOT change what constitutes a \"real schema change\" — any code / type / literal change still bumps the version
- Do NOT retro-adjust the current hash: this is a forward improvement; existing wire compatibility is unaffected

## Test plan

- Unit test: normalizer with fixtures (pure-comment diff, pure-code diff, mixed diff, string literal preservation, template literal preservation)
- Integration: run codegen on `main` vs a comment-only patch — should produce identical `SCHEMA_VERSION`
- Integration: run codegen on `main` vs a real type change — should produce different `SCHEMA_VERSION`

## Priority

Low. Comment-only PRs on shared schemas are infrequent. The behavior is documented (benign one-time reload). This is a quality-of-life fix, not a correctness gap.

## Related

- PR #1188 (Issue #898) — surfaced this observation via `schema-version.gen.ts` 2-line hash regen from a comment-only sweep
- Architect audit note in Sprint 2026-07-18 memo (Orchestrator)

ms2sato:
## Why

`scripts/generate-schema-version.mjs` computed `SCHEMA_VERSION` as a sha256 over the raw bytes of every `.ts` file directly under `packages/shared/src/schemas/`. Because it hashed raw bytes, a comment-only edit (docstring fix, cleanup sweep — see PR #1188 / Issue #898) bumped `SCHEMA_VERSION`. That version propagates via the `wire-version` WebSocket handshake and forces a one-time reload for every connected client, despite zero wire-format impact. This PR normalizes the hash input so comment-only and whitespace-only edits stop bumping the version, while any real code/type/literal change still bumps it exactly as before.

## Approach: real TypeScript parser, not regex

`scripts/schema-source-normalize.mjs` (new) exports `normalizeSchemaSource(source)`, which parses each schema file with `ts.createSourceFile` — the TypeScript compiler's own parser — walks the AST down to its leaf tokens, drops comment trivia, and collapses whitespace/blank-line gaps between tokens to a single normalizing separator (`''`, `' '`, or `'\n'`). Every other token's text — including the full contents of string, template, and regex literals — is kept byte-for-byte verbatim.

**Regex-based comment stripping was intentionally not used.** A regex cannot reliably distinguish:
- a comment-like sequence (`//`, `/* */`) that appears *inside* a string or template literal, from an actual comment
- a regex literal's `/` delimiter from a division operator or the start of a comment

Both are real patterns in this codebase's schema files (e.g. `packages/shared/src/schemas/agent.ts` contains non-trivial regex literals with `+`, `*`, `[`, `]`). A regex stripper that gets either wrong produces a **false negative** — semantic content silently dropped from the hash input, meaning a real wire-format change would no longer bump `SCHEMA_VERSION`. That is a correctness bug (silent wire drift, no client reload), and is categorically worse than the friction this PR fixes. Only a real parser resolves these cases correctly by construction, because it tracks full grammatical context rather than pattern-matching characters.

One additional wrinkle found during implementation: `ts.createSourceFile` expands `/** ... */` JSDoc comments into their own AST subtree (tags become real `Identifier` nodes). A naive leaf-walk would leak JSDoc-derived text into the hash. Fixed by skipping recursion into any node in the `ts.SyntaxKind.FirstJSDocNode..LastJSDocNode` range — the surrounding tokens' trivia normalization already accounts for the JSDoc comment's byte range.

## Fail-closed fallback

If a schema file fails to parse (syntax error), `computeVersion()` catches the thrown error and falls back to hashing that file's **raw, unmodified bytes** instead of the normalized text. Worst case: an unnecessary version bump for a file that shouldn't even exist in a broken state. It never silently drops content — the asymmetry that matters here is that a missed semantic change (false negative) is far more dangerous than an extra bump (false positive), so every failure mode degrades toward "bump more," never "bump less."

## Polarity matrix — all 8 rows covered by tests

| # | Input diff shape | Required outcome | Verified by |
|---|---|---|---|
| 1 | Pure comment diff (line/block/trailing comment added, removed, or edited) | **Same** hash | `scripts/__tests__/schema-source-normalize.test.mjs` (3 cases) + `packages/shared/src/__tests__/schema-version.gen.test.ts` real-file case |
| 2 | Pure code diff (literal value or field name changes) | **Different** hash | `schema-source-normalize.test.mjs` (2 cases) + `scripts/__tests__/generate-schema-version.test.mjs` (temp-fixture end-to-end) |
| 3 | Mixed diff (comment change + code change together) | **Different** hash (code portion still registers) | `schema-source-normalize.test.mjs` |
| 4 | String literal containing comment-like characters (`"// not a comment"`, `'/* also not a comment */'`) | Preserved verbatim; changing only the literal flips the hash | `schema-source-normalize.test.mjs` (3 cases) |
| 5 | Template literal with an interpolated expression, both containing comment-like text | Preserved verbatim; changing literal text OR the expression flips the hash | `schema-source-normalize.test.mjs` (3 cases) |
| 6 | Whitespace-only diff (blank lines, indentation, inline spacing) | **Same** hash | `schema-source-normalize.test.mjs` (3 cases) |
| 7 | File fails to parse (malformed syntax) | Falls back to raw-byte hash for that file (fail-closed) | `schema-source-normalize.test.mjs` (throw contract) + `generate-schema-version.test.mjs` (fallback determinism) |
| 8 | Empty file | Deterministic, no crash | `schema-source-normalize.test.mjs` (2 cases) |

Also covered: a dedicated regex-literal section proving the parser correctly distinguishes a regex literal's `/` from division/comments, using a real complex pattern already present in `packages/shared/src/schemas/agent.ts`.

## One-time SCHEMA_VERSION bump (expected)

`SCHEMA_VERSION` changes from `b1b79bb76b53040d` to `ab98cef18cf6b3f3` in this PR. This is a single, deliberate bump from switching the hash algorithm itself (raw bytes → normalized tokens) on the current, unchanged schema files — not a regression. It happens exactly once, at landing; every comment-only PR after this one will no longer bump the version. Confirmed idempotent: repeated `--print` / `--check` / plain-regen runs all produce this same value.

## Existing test updated (not weakened)

`packages/shared/src/__tests__/schema-version.gen.test.ts` had a test that appended a bare comment to `embedded-agent.ts` and asserted the hash changed — which is now the opposite of the intended behavior. Updated it to perturb via a non-comment addition (`export const __schemaVersionSensitivityProbe = 1;`) instead, and added a complementary test proving a comment-only perturbation now leaves the hash unchanged.

## Test plan

- [x] `bun run typecheck` — 0 errors
- [x] `bun run test` — full suite green (see below)
- [x] `bun run check:lang` — exit 0 (no `docs/`/`.claude/`/`CLAUDE.md` touched)
- [x] `node .claude/skills/orchestrator/preflight-check.js` — exit 0; flags `packages/shared/src/schema-version.gen.ts` as a cross-package contract change and recommends integration test review (its generic heuristic for any `packages/shared/src/` diff). No wire-schema *shape* changed here — only the hash algorithm — and the existing `packages/integration/src/schema-version-boundary.test.ts` (unmodified, still passing) already covers the `SCHEMA_VERSION` wire-boundary contract. Scoped this PR's proof to the pure-function unit tests plus a real end-to-end CLI run against the actual schemas directory instead of adding a new integration test, per the Issue's test plan.
- [x] Manual idempotency check: `--print` x2, `--check` x2, plain regen — single version-string change only, no drift across repeated runs
- [x] Duplication check: grepped for existing `createScanner` / `createSourceFile` / `getLeadingCommentRanges` usage — none found. `scripts/check-source-comment-blame-shift.mjs` handles a different concern (Issue/PR reference detection in comments) via an unrelated hand-rolled state machine; not reused since it doesn't provide the parse-fidelity or fail-closed guarantees this change requires.

**Pre-existing unrelated flake noted, not caused by this PR.** One full-suite `bun run test` run showed 1 failure in `packages/client` (`EmbeddedAgentWorkerView > Context Handoff (Phase A) > blocks Send (morphs to Cancel) while a handoff is in flight`). This diff touches zero files under `packages/client`. The same test file passes 70/70 in isolation, and the suite as a whole passes with 0 failures on repeated full runs — this reads as an order-dependent flake between unrelated client test files, not a regression from this change.

## Note on CodeRabbit

Local CLI (`coderabbit review --agent --base main`) hit the headless-worktree browser-auth limitation (stuck at `awaiting_browser_auth` with no interactive browser available in this session) — a structural limitation, not a rate-limit. GitHub-side bot returned a genuine rate-limit reply ("Review limit reached... next review available in: 39 minutes") on the initial auto-review at PR open. Per the `coderabbit-ops` skill's headless-auth + bot-rate-limit disposition: this is a real countdown on the GitHub side (not a permanently-unavailable channel), so once the window elapses a manual `@coderabbitai review` retrigger (or the next auto-review) should produce a genuine walkthrough before merge — will confirm `reviewDecision` before merging rather than substituting a dual-clean audit, since the GitHub channel is expected to become available.

Closes #1190


ms2sato:
## Note on a benign edge case (non-blocking, for future readers)

The trivia-normalization separator (`''` / `' '` / `'\n'`) is a function of *whether the gap contains a newline*, not of comment presence/absence. This means a mid-line comment insertion that changes the surrounding whitespace shape — e.g. `a b` → `a // x\nb` — flips the separator from `' '` to `'\n'` even though the only "real" content added was a comment. In this rare case, a comment-only edit *can* still bump `SCHEMA_VERSION`.

This is intentional and safe under the design's failure-asymmetry: it's a false positive (unnecessary bump / extra client reload), never a false negative (a real change silently unhashed). The common case — a comment added/edited without changing surrounding line structure — stays stable, which is what this PR set out to fix.

Issue #1226, opened 2026-07-27
Add lint to gate new cross-file-imported mock.module() targets


ms2sato:
## Background

Issue #977 (closed by PR #1227) converted the last 5 active `mock.module()` cross-file poisoners in `packages/client` to DI seam / `spyOn` / real-module-plus-injected-value patterns, and added the prohibition to `.claude/rules/testing.md` Anti-Pattern #2: never `mock.module()` a target that any other test file imports for real.

That rule is currently enforced by review only. This Issue adds the mechanical lint that gates *new* violations in CI.

## Design ruling (architect, 2026-07-28) — supersedes the original client-only / import-graph framing

Two facts established by a repo-wide call-site scan (comment mentions excluded) change the original plan:

1. **The "expected baseline: 0" claim is true only for `packages/client`.** Repo-wide, ad-hoc `mock.module()` call sites still exist outside the sanctioned central mock registry — in `packages/server` (`__tests__/api.test.ts` ×2, `routes/__tests__/system.test.ts`, `services/__tests__/worktree-creation-service.test.ts`, `services/__tests__/worktree-deletion-service.test.ts`, `services/inbound/__tests__/diff-worker-handler.test.ts`) and `packages/integration` (`paste-focus-isolation.test.tsx` ×2). The integration case is a live latent instance of the exact #1225 class: it mocks `@agent-console/client/src/lib/api`, which sibling integration tests (e.g. `system-api-boundary.test.ts`) import for real in the same `bun:test` process. A client-only lint would gate the one package that is already clean while leaving the two packages where the #970/#976/#1225 incidents actually occurred unguarded. **Scope is therefore repo-wide** (`packages/*/src/**/*.{ts,tsx}`).
[...]

Issue #1225, opened 2026-07-27
CI: main red since 2026-07-23 — integration System API openPath fails with ChildProcess unref compat error (blocks all merges)


ms2sato:
## Summary

`main` has been **red since 2026-07-23** and nobody noticed for four days, because no push landed on `main` during the owner-gate wait on PR #1223. Every open PR is blocked: CI green is a merge precondition, and this failure reproduces on any branch cut from current `main`.

Two `@agent-console/integration` cases fail:

```
(fail) Client-Server Boundary: System API > openPath > should call the correct endpoint /api/system/open
(fail) Client-Server Boundary: System API > openPath > should handle file path (opens containing directory)
 2 fail
```

## Verified facts

| Fact | Evidence |
|---|---|
| `main` CI is red at `c5c5c795` | [run 29969428929](https://github.com/ms2sato/agent-console/actions/runs/29969428929) conclusion `failure` |
| It is **deterministic**, not flaky | Re-ran the same run 2026-07-27; identical two failures |
| It reproduces on a fresh branch | PR #1224 (`test/952-...`, test-only diff in `packages/server`) fails identically |
| It is **not** a server-package regression | Same runs: `@agent-console/server` 3652 pass / 0 fail / exit 0 |
| It does **not** reproduce locally | Same commit `c5c5c795`, bun **1.3.10**: `packages/integration` 73 pass / 0 fail / exit 0 |
| CI pins an older bun | `.github/workflows/ci.yml:29` — `bun-version: "1.3.5"` |

## Mechanism

The server returns **400** with the message:

```
ApiError: this.#handle.unref is not a function. (In 'this.#handle.unref()', 'this.#handle.unref' is undefined)
    at handleApiError (packages/client/src/lib/api.ts:78:9)
    at async openPath (packages/client/src/lib/api.ts:666:11)
    at async packages/integration/src/system-api-boundary.test.ts:147:13
```

`packages/server/src/routes/system.ts`'s `POST /open` handler calls the `open` npm package. `open@11.0.0/index.js` calls `subprocess.unref()` at lines 289 and 296. The `#handle` private field belongs to the runtime's `node:child_process` `ChildProcess` implementation, so `unref` being undefined is a runtime-compat failure, not a package-code failure. The handler's `catch` converts the thrown error into a `ValidationError`, which surfaces as HTTP 400 — which is why the test sees an `ApiError` rather than a crash.

## Open questions for the fix

[...]

Issue #970, opened 2026-07-03
test(client): order-dependent 'window is not defined' failures (48 tests) at main HEAD under bun >= 1.3.8 — masked by CI's bun 1.3.5


ms2sato:
## Symptom

Running the FULL client test suite at current main HEAD (`4dd093b`, after today's merges including #962 / #967) fails **48 tests** with `window is not defined` — the happy-dom global appears to be lost mid-run. Reported by the Issue #948 agent during pre-push verification of PR #969 (2026-07-03).

## Key facts (verified by the reporting agent)

- Reproduces under **bun 1.3.10 (host)** AND **bun 1.3.8 (inside the new Docker dev container — its first practical use)**
- **CI (bun 1.3.5) is green** — the failure is masked by CI's older bun
- **Every failing file passes in isolation**; small ordered subsets also pass → order-dependent teardown/pollution, not a per-file defect
- The reporting branch's client code is **byte-identical to origin/main** → main-side issue, unrelated to PR #969's diff
- Two quick culprit-bisect attempts (terminal-store, new terminal component tests) did NOT reproduce in small sets — a proper bisect over the full file order is needed

## Impact

- Local full-suite verification on any up-to-date bun fails 48 tests → every agent/developer running `bun run test` at main pays a false-failure triage cost (same failure-mode class as Issue #952's env flake: normalizes "N fail" as expected noise)
- CI green vs local red divergence undermines the "main is always GREEN" invariant in practice

## Suspects (unverified)

- Today's client-heavy merges #962 / #967 (labs terminal renderer track) landed test files that may unregister/close the happy-dom window in a shared teardown
- bun >= 1.3.8 test runner ordering/isolation change interacting with an existing latent teardown bug

## Suggested fix path

1. Bisect over the full client test file order at main HEAD with bun >= 1.3.8 (binary-search the file list; failing halves keep the pollution source)
2. Once the polluting file is identified, fix its teardown (restore/preserve the happy-dom global) rather than reordering
3. Consider bumping CI's bun to match the dev floor (>= 1.3.8) so this class is caught in CI — separate decision, note the supply-chain gate in package.json preinstall

## Acceptance Criteria

- [ ] Full client suite passes at main under bun 1.3.8 and 1.3.10 → paste both runs
- [ ] Root-cause file + mechanism named in the PR body (not just a reorder)
- [ ] Polluting teardown has a regression test or lint-able pattern if generalizable

## Related

[...]

PR #1227, opened 2026-07-27
test: eliminate remaining cross-file mock.module poisoning in client tests (closes #977)

### LINKED ISSUE #977: Eliminate remaining cross-file mock.module poisoning in client tests (bun readdir-order hazard, lesson from #976 CI)
## Origin

PR #976 CI failed with 21 `no ws` test failures that never reproduced locally. Root cause (proven with a deterministic synthetic mocker/victim pair, both bun 1.3.5 and 1.3.8): **bun:test's `mock.module` is process-global and poisons every test file that loads after the mocking file**, and bun executes test files in **directory readdir order** (not CLI-arg order), which differs between Linux CI and macOS — so the same suite is green on one OS and red on the other depending on which file happens to load first.

PR #976 removed the three active poisoners (`TerminalAdapter.test.tsx`, `MessagePanel.test.tsx`, `useMessageTemplates.test.ts`) by converting them to DI seams / fetch-level stubs. This issue tracks the REMAINING debt and the standing rule.

## Remaining `mock.module` usages (pre-existing, currently harmless by readdir accident)

`useAppWs`, `__root`, `capabilities`, `useCreateWorktree` test files still `mock.module` shared modules. Convert them to the same patterns (DI seam prop / fetch-level request router / real store + installMockWebSocket) so no test file can poison a neighbor.

## Rule candidate (add to test-standards / testing rules once converted)

**Never `mock.module` a module that any other test file imports for real.** Prefer, in order: (1) a DI seam on the component/module under test, (2) fetch-level request stubbing, (3) real module + injected fakes. If `mock.module` is unavoidable, the target must be consumed exclusively by that one test file — and note that bun gives NO isolation guarantee between files in one process.

## Also noted

CI runs bun 1.3.5, local dev 1.3.8 — drift didn't matter here (leak exists on both) but a pin/upgrade alignment is worth doing while touching CI config.

---

### Correction to the residual list

A `grep -rln 'mock\.module'` over `packages/client/src` returns 16 files, but that pattern matches **comments and string literals** as well. Filtering to actual call sites (`mock.module(`) at main `c5c5c795` gives **5 call sites across 4 target modules**:

| Call site | Mocked target |
|---|---|
| `src/hooks/__tests__/useCreateWorktree.test.ts:7` | `routes/__root` |
| `src/components/__tests__/SessionSettings.test.tsx:98` | `routes/__root` |
| `src/__tests__/routes/WorktreeRow.test.tsx:10` | `lib/capabilities` |
| `src/components/sessions/hooks/__tests__/useSessionPageState.test.ts:24` | `hooks/useAppWs` |
| `src/components/worktrees/__tests__/QuickWorktreeDialog.test.tsx:11` | `hooks/useCreateWorktree` |

This matches the residual list in the Issue body above — it had not drifted. Two notes that follow from the correction:

- **`src/components/terminal/TerminalAdapter.tsx` (production) is not in scope.** Its hit is a comment explaining why PR #976 moved to a DI seam. No production change, no follow-up Issue.
- **The three files PR #976 converted are still correctly converted.** Their hits are comments / props-based code, not re-introduced `mock.module` calls. This is not a regression.

Apply the same discipline during implementation: do not conflate a comment or string occurrence with a real call.

### 1. Scope — two phases, limited to *active* poisoners

**Phase 1 (classification, first task).** For each of the 5 call sites, determine whether **any other test file real-imports the mocked target module**. Classify each as:

- **active poisoner** — another test file imports the target for real, so this mock leaks into it
- **benign** — the target is consumed exclusively by this one test file

Publish the classification as a table in the PR body: target module / victim file(s) / verdict.

**Phase 2 (conversion).** Convert **only the active poisoners** to a DI seam, fetch-level request stub, or props-based injection (in that order of preference). Leave benign cases as-is and record why.

`routes/__root` is the expected principal class — it is the route root, so other route tests real-import it. If Phase 1 surfaces substantially more work than this shape implies, raise a PR-split proposal rather than pushing one oversized PR.

### 2. Rule wording — prohibit cross-file-imported targets, not `mock.module` outright

A blanket ban would kill legitimate self-consumed mocks. The rule is:

> Never `mock.module` a module that any other test file imports for real. `bun:test`'s `mock.module` is process-global and irreversible, so it poisons every test file loaded afterwards in the same process, and bun executes test files in readdir order (which differs across operating systems).

State the `routes/__root` mock as the prohibited example and a self-consumed mock as the permitted one.

### 3. Rule placement

- **`.claude/rules/testing.md`** — the prohibition and the cross-file criterion (auto-loads for all test work; extends the existing "Module-Level Mocking anti-pattern" material).
- **`.claude/skills/test-standards/`** — the how-to for DI seam / fetch stub / props-based conversion.
- **Mechanical lint is out of scope for this PR.** Deciding whether a mock target is real-imported elsewhere requires import-graph analysis. File a follow-up Issue for a lint that gates *new* poisoners using the allowlist-baseline pattern from `workflow.md`.

### 4. Regression verification — a green full suite is NOT sufficient

The #970 class (green on Linux CI, red on macOS) came precisely from readdir-order dependence, so ordering must be exercised explicitly.

- **Polarity per converted poisoner (mandatory).** Run the poisoner file and its victim file in the **same process, poisoner first**. Before the conversion, the victim must be observed receiving the mocked module (poisoning demonstrated); after the conversion, the victim must receive the **real** module. For `routes/__root`: run a file that mocks `__root`, then a victim test that real-imports `__root`, and assert the victim gets the real one.
- **Order independence.** After conversion, demonstrate green in **both** orders for each polarity pair (poisoner→victim and victim→poisoner). Exhaustive permutation is not required; the pair-level demonstration is the minimum.
- A plain "full suite is green" only shows that the current readdir order happens not to expose the poisoning. Say so explicitly in the PR body.

### Implementation scope

Phase 1 classification table → Phase 2 conversion of active poisoners → `testing.md` + `test-standards` additions → polarity / ordering tests → follow-up Issue for the lint.

Test-only PR (mergeable by the Orchestrator, does not add to the owner's approval queue). Reversibility: **reversible**.

ms2sato:
Closes #977

## What

Converts the 5 remaining active `mock.module()` call sites in `packages/client` (across 4 target modules) to cross-file-safe patterns, and codifies the prohibition + conversion how-to. Full suite green, both directions of every poisoner/victim pair verified.

## Phase 1 — classification (all 5 call sites are active poisoners)

| Call site | Mocked target | Overridden export(s) | Real cross-file consumer | Verdict |
|---|---|---|---|---|
| `hooks/__tests__/useCreateWorktree.test.ts:7` | `routes/__root` | `useWorktreeCreationTasksContext` | any real re-import of `routes/__root` (verified directly); transitively, `routes/index.tsx`'s `RepositoryCard` (rendered by `routes/__tests__/index.test.tsx` for the >=1-repo case) | **ACTIVE** |
| `components/__tests__/SessionSettings.test.tsx:98` | `routes/__root` | `useWorktreeDeletionTasksContext`, `WorktreeDeletionTasksContext` | `__tests__/routes/WorktreeRow.test.tsx` (imports `WorktreeDeletionTasksContext` directly) | **ACTIVE** |
| `__tests__/routes/WorktreeRow.test.tsx:10` | `lib/capabilities` | `hasVSCode`, `getVSCodeOpenMode`, `getVSCodeRemoteHost` (not `setCapabilities`) | `routes/__tests__/index.test.tsx` (spies on the real module) | **ACTIVE** |
| `components/sessions/hooks/__tests__/useSessionPageState.test.ts:24` | `hooks/useAppWs` | `useAppWsEvent`, `useAppWsState` (the module's only 2 exports) | `routes/__tests__/index.test.tsx`, `__tests__/routes/agents/index.test.tsx` (both spy on the real module) | **ACTIVE** |
| `components/worktrees/__tests__/QuickWorktreeDialog.test.tsx:11` | `hooks/useCreateWorktree` | `useCreateWorktree` (the module's only export) | `hooks/__tests__/useCreateWorktree.test.ts` (its own unit-under-test) | **ACTIVE** |

Zero benign/self-consumed cases this round; scope matched the `routes/__root`-centric shape the Issue anticipated (2 of 5 call sites target `__root`), so no PR-split was needed.

### A note on classification methodology (worth flagging)

Bun's `mock.module(specifier, factory)` **merges** `factory()`'s return onto the real module rather than fully replacing it — an export the factory doesn't declare falls through to the real implementation for every importer. This means "does the victim's test suite still pass" is **not** a reliable signal for classification: a victim can receive a genuinely poisoned export and still pass every assertion if it happens to be structurally tolerant (e.g. it re-derives both a Provider and consumer from the same poisoned reference, so both sides stay mutually consistent without ever touching the real module).

Classification here was done by object-identity checks (does the victim's imported Context `===` the poisoner's locally-created mock Context?) and source-string checks (does the victim's resolved function's `.toString()` match the poisoner's factory body?), forcing deterministic poisoner-then-victim load order via a lexicographically-earlier-sorting temp file copy (Bun's scheduler does not respect CLI argument order or simple directory position — verified empirically, not assumed). Documented in `test-standards.md`'s new section for future classification work.

## Phase 2 — conversions

- **`routes/__root` (both mockers):** replaced with the real `WorktreeCreationTasksContext` / `WorktreeDeletionTasksContext` Providers from `contexts/root-contexts.ts` (the module `routes/__root.tsx` re-exports from, without pulling in the router-registration barrel) wrapping an injected fake value. Zero production changes.
- **`lib/capabilities`:** replaced with the module's own real `setCapabilities()` setter (already existed for exactly this purpose). Zero production changes.
- **`hooks/useAppWs`:** replaced with `spyOn()` + `.mockRestore()` in `afterEach`, matching the pattern already used by `routes/__tests__/index.test.tsx` / `__tests__/routes/agents/index.test.tsx` for the same module.
- **`hooks/useCreateWorktree`:** replaced with `spyOn()` + `.mockRestore()` in `afterEach`.

## Regression verification — polarity + order-independence

A plain "full suite is green" only shows the current readdir order happens not to expose the poisoning (that's exactly how #970/PR #976's bug hid locally and only surfaced in CI). Explicit polarity evidence per converted poisoner, all pairs run in the same bun:test process with deterministically forced load order:

- **Before the fix** (pre-fix content re-materialized from `HEAD` for just the poisoner+victim pair, forced poisoner-first): every pair demonstrably leaks.
  - `useCreateWorktree.test.ts` → `routes/__root`: a real re-import's `useWorktreeCreationTasksContext` resolves to the poisoner's canned closure body.
  - `SessionSettings.test.tsx` → `routes/__root`: `WorktreeRow.test.tsx`'s imported `WorktreeDeletionTasksContext` is `Object.is`-identical to the poisoner's locally-created mock Context.
  - `WorktreeRow.test.tsx` → `lib/capabilities`: a real re-import's `hasVSCode` resolves to the poisoner's `() => false`, not the real cached-value reader (`setCapabilities`, not overridden, stays real — confirming the merge, not full-replace, semantics).
  - `useSessionPageState.test.ts` → `hooks/useAppWs`: a real re-import's `useAppWsEvent` resolves to the poisoner's callback-capturing closure.
  - `QuickWorktreeDialog.test.tsx` → `hooks/useCreateWorktree`: forcing this pair against `useCreateWorktree.test.ts` (its own unit test) drops 4 of 13 tests to failing, because the victim's own hook-under-test is silently replaced by the poisoner's canned mock.
- **After the fix**: every pair re-run in **both** orders (poisoner-path-first and victim-path-first, using the real current file paths as explicit `bun test` arguments) is green with the full expected test count, for all 5 pairs.

## Phase 3 — rule / skill updates

- `.claude/rules/testing.md` Anti-Pattern #2: added the cross-file-imported-target prohibition, a prohibited example (`routes/__root`) and a permitted example (self-consumed target), plus a pre-implementation-checklist bullet.
- `.claude/skills/test-standards/test-standards.md`: added the 3-pattern conversion how-to (DI seam / `spyOn` / real-module-plus-injected-fake) with worked code examples referencing this PR and PR #976, plus the merge-semantics + load-order-control methodology note above.

## Out of scope (per Issue AC)

A mechanical lint gating *new* cross-file-imported `mock.module()` targets requires import-graph analysis and is filed as a follow-up: #1226.

## Test plan

- [x] `bun run test` (full workspace: client 2075, shared 596, integration 73, server 3653, embedded-agent 287, scripts/hooks 483) — all green, `TEST_EXIT: 0`
- [x] `bun run typecheck` — clean (part of `bun run test`)
- [x] `bun run check:lang` — clean (113 files scanned)
- [x] `node .claude/skills/orchestrator/preflight-check.js` — clean (no coverage gaps, no rule/skill duplication, no language violations, no blame-shift comments)
- [x] Per-pair polarity + both-orders-green verification (above)
- [ ] `coderabbit review --agent --base main`

## CodeRabbit status (2026-07-27, final) — CLEAN

`main`'s red state (#1225) resolved via PR #1229; head rebased onto it (`5cfe359f`). First genuine CodeRabbit walkthrough on `5cfe359f` (after two rate-limit rounds -- local CLI: headless-worktree `automatic_login_failed`, structural; GitHub bot: rate-limited at PR open, then the "Review finished... does not re-review" quirk on a same-commit retrigger, then a fresh rate-limit on auto-review-on-push) returned 2 actionable findings, both in `test-standards.md`:

1. **Major** -- the `spyOn()` cross-file-safety claim was stated as an unconditional guarantee; it actually depends on this repo's test files running sequentially (no `--concurrent`, no `concurrentTestGlob`). Fixed: named the precondition explicitly and the concurrent-execution case where it would not hold.
2. **Minor** -- the `useAppWsState` mock example used an arbitrary `false as T` cast instead of exercising the real selector contract. Fixed: replaced with a `selector(fakeState)` example, kept the cast form only as a documented exception (matching the existing `routes/__tests__/index.test.tsx` precedent).

Fix commit `abc6efb7` pushed; verified against actual repo config before writing (no `--concurrent` flag in `packages/client/package.json`'s test script, no `concurrentTestGlob` in `bunfig.toml`). Re-review on `abc6efb7` (after a third rate-limit round, cleared and manually retriggered) returned a genuine walkthrough: **"No actionable comments were generated in the recent review."** Zero new findings; both prior findings confirmed resolved, not carried forward.

**3-layer verdict on `abc6efb7`**: pre-merge checks clean, CodeRabbit commit status `state: success` / `description: "Review completed"` (verified via the walkthrough body, not just the status line, per the state-vs-description trap), inline actionable comments = 0, `reviewDecision` empty (walkthrough-exists clean-equivalent case per `coderabbit-ops` troubleshooting -- a genuine review ran, found nothing to flag, and CodeRabbit does not always submit a formal `APPROVED` event for zero-finding passes). Independent Architect code-appropriateness audit (session `56b4d446`, gen-2): **CLEAN** -- confirmed zero production changes, verified the context-identity chain between `routes/__root.tsx` and `contexts/root-contexts.ts`, verified no hidden dependency on the removed `SessionSettings` mock's no-Provider fallback branch, and independently confirmed the CR-Major fix's config claims and the documented-cast exception's premise.

## Note on CI

The `test` job's failure is a confirmed pre-existing `main`-branch flake (`this.#handle.unref is not a function` in `packages/client/src/lib/api.ts`'s `openPath`), already tracked in #1225 and reproduced identically on `main`'s own most recent CI run (2026-07-23). This PR's diff never touches `packages/client/src/lib/api.ts`; locally `bun run test` (full workspace) is green with `TEST_EXIT: 0`. All other CI jobs (preflight, language-lint, comment-blame-shift-lint, check-structure, CodeQL/Analyze) pass.

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by removing process-global module mocking in favor of per-test spies, real context providers, and resettable setup (covering worktree rows, session settings, WebSocket hook behavior, and worktree creation dialogs/hooks).
  * Kept existing assertions for UI states and navigation/DELETE behavior while making setup and cleanup more reliable.

* **Documentation**
  * Expanded testing guidance with a new “cross-file module mocking poisoning” procedure, clearer semantics/examples, and a repo-wide checklist to prevent unsafe mocking patterns.

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -21,6 +21,12 @@ Test file re-implements production logic instead of importing it. Signs: test-on
 ### 2. Module-Level Mocking
 Using `mock.module()` or `vi.mock()` instead of fetch-level mocks. Problems: bypasses actual API function logic, `mock.module()` is permanent in bun:test, tests pass even when integration is broken. Has caused production incidents (mocking `config.js` broke 26+ unrelated tests). **Preferred: dependency injection over module mocking.**
 
+**Never `mock.module()` a target that any other test file imports for real.** `bun:test`'s `mock.module()` is process-global and irreversible for the life of the test process, so it poisons every test file loaded afterward in the same process — and bun runs test files in directory readdir order (not CLI-arg order), which differs across operating systems. A suite that is green on one OS and red on another (or green locally and red in CI) is the signature of this failure mode (Issue #970, PR #976, Issue #977).
+
+- **Prohibited example:** `mock.module('../../routes/__root', () => ({ useWorktreeCreationTasksContext: () => ({ ... }) }))`. `routes/__root` is the route root, so multiple other test files (route tests, sibling component tests) import it for real; a mock factory that only re-declares a subset of its exports silently breaks any file that loads afterward and needs an export the factory omitted.
+- **Permitted example:** a module consumed exclusively by the one test file mocking it (e.g. a component's own tightly-scoped internal helper with no other importer) may still use `mock.module()`, but confirm the exclusivity first — grep the repo for other real importers before relying on this exception.
+- **Before adding a new `mock.module()` call**, grep the repository for other files that import the target module without mocking it. If any exist, do not use `mock.module()` — use one of, in order of preference: (1) a DI seam (prop / injected factory) on the component or hook under test, (2) `spyOn()` on the module's named export (restorable per-test via `.mockRestore()` in `afterEach`), (3) fetch-level request stubbing, (4) a real store/context with an injected fake value. See `test-standards` skill for worked conversion patterns.
+
 ### 3. Private Method Testing
 Attempting to test internal/private methods directly. Test through public interface instead, or extract to a separate module if complexity warrants it.
 
@@ -74,3 +80,4 @@ Before writing tests, verify:
 - [ ] Not following existing bad patterns blindly
 - [ ] Not changing production code just for testing without discussion
 - [ ] **Target code is mockable via DI** — check if the code under test imports module-level singletons. If it does, DI refactoring is required before the test can be written safely. Do NOT use `mock.module()` to work around missing DI. See Anti-Pattern #2.
+- [ ] **If a new `mock.module()` call is unavoidable**, grep the repo for other real importers of the target module first. See Anti-Pattern #2's cross-file-imported-target prohibition.
--- a/.claude/skills/test-standards/test-standards.md
+++ b/.claude/skills/test-standards/test-standards.md
@@ -195,7 +195,9 @@ render(<TerminalAdapter sessionId="s" workerId="w" createInstance={mockGetOrCrea
 
 ### Pattern 2 — `spyOn()` on a named export
 
-Best for a hook or function the component-under-test imports and calls directly, when adding a DI prop is not warranted. `spyOn` is restorable per-test (unlike `mock.module()`), so it must be paired with `.mockRestore()` in `afterEach` — without that, the spy leaks into the next test in the same file (not other files, since `spyOn` targets a specific module-namespace object each test file imports independently).
+Best for a hook or function the component-under-test imports and calls directly, when adding a DI prop is not warranted. `spyOn` is restorable per-test (unlike `mock.module()`), so it must be paired with `.mockRestore()` in `afterEach` — without that, the spy leaks into the next test in the same file.
+
+**Cross-file safety is conditional, not unconditional.** Bun shares the module cache within a single process; `spyOn`'s restore-on-`mockRestore()` is what keeps it file-scoped only because this repo's `bun run test` invocation (`bun test --preload ./src/test/setup.ts src/` in `packages/client`) runs test files **sequentially** — no `--concurrent` flag on the script, no `concurrentTestGlob` in `bunfig.toml`. Under that condition, one file's spy is always torn down (via `afterEach`'s `mockRestore()`) before the next file's module-level code runs, so it cannot bleed across files. If a future test opts into Bun's `--concurrent` flag, `test.concurrent()`, or a `concurrentTestGlob` config, this guarantee no longer holds: a spy installed by one concurrently-running test file becomes observable by another before its `mockRestore()` fires, since they'd share the same in-process module cache at the same time. Do not use `spyOn()` on a shared module inside a `test.concurrent()` block without independently re-verifying isolation.
 
 ```typescript
 import * as useAppWsModule from '../../hooks/useAppWs';
@@ -211,11 +213,19 @@ afterEach(() => {
 });
 ```
 
-A generic function (`useAppWsState<T>(selector: (state) => T): T`) needs an explicit generic on the mock implementation so the cast is not silently `any`:
+A generic function (`useAppWsState<T>(selector: (state: AppWebSocketState) => T): T`) needs a real fake state run through the caller's own `selector`, not an arbitrary cast — a cast-through-`T` (e.g. `<T,>() => false as T`) type-checks for any selection but silently returns the wrong value the moment a test's `selector` reads a field the cast didn't account for:
 
 ```typescript
-useAppWsStateSpy = spyOn(useAppWsModule, 'useAppWsState').mockImplementation(<T,>() => false as T);
+import type { AppWebSocketState } from '../../lib/app-websocket';
+
+const fakeWsState: AppWebSocketState = { /* ...minimal real shape... */ } as AppWebSocketState;
+
+useAppWsStateSpy = spyOn(useAppWsModule, 'useAppWsState').mockImplementation(
+  <T,>(selector: (state: AppWebSocketState) => T) => selector(fakeWsState),
+);
 ```
+
+If a test genuinely never observes the mocked value (the component under test doesn't call `useAppWsState` on any path the test exercises), a cast is acceptable but must say so in a comment — see `routes/__tests__/index.test.tsx`'s `<T,>() => false as T` for the documented-exception form.
 
 (Worked examples: `routes/__tests__/index.test.tsx`, `__tests__/routes/agents/index.test.tsx`, `components/sessions/hooks/__tests__/useSessionPageState.test.ts`, `components/worktrees/__tests__/QuickWorktreeDialog.test.tsx`.)
 
--- a/.claude/skills/test-standards/test-standards.md
+++ b/.claude/skills/test-standards/test-standards.md
@@ -168,6 +168,105 @@ describe('Client-Server Boundary', () => {
 
 This gives you a single test file that exercises the full round-trip: user event → form serialization → HTTP body → server handler → persisted state. If any step drops or mistransforms data, the test fails.
 
+## Converting Cross-File `mock.module()` Poisoning
+
+See [rules/testing.md](../../rules/testing.md) Anti-Pattern #2 for the prohibition (never `mock.module()` a target another test file real-imports) and when the exception applies. This section is the conversion how-to, with three worked patterns in order of preference. Before converting, grep the repo for other real importers of the target module to confirm the poisoning classification (Issue #977 Phase 1).
+
+### Pattern 1 — DI seam (prop / injected factory)
+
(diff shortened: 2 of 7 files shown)


ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

Issue #1211, opened 2026-07-19
bug(docker-dev-stack): vite /ws proxy blocked by upstream Bun issue — provide working Browser QA path (was: proxy hangs)


ms2sato:
## Symptom

Detected via A6 dogfood delegate (session `e5dcd3b9`) during multi-user AGENTS.md verification for [Issue #1123](https://github.com/ms2sato/agent-console/issues/1123), 2026-07-19.

Docker dev stack (`docker/docker-compose.yml`, host `DEV_CLIENT_PORT=15173 DEV_SERVER_PORT=13457`):
- Login to `alice` succeeds (auth_token cookie issued, `Logout (alice)` visible)
- Dashboard stuck at "Loading sessions..." indefinitely
- Root cause: `packages/client/vite.config.ts` `/ws` proxy (`ws: true`) establishes backend upgrade but **never returns response to client**

## Diagnostic evidence (delegate-verified)

1. Server-side log: `App WebSocket connected, sending initial sync` fires every attempt (backend sees upgrade success)
2. Container-internal direct WS to app server (bypasses vite): `ws://localhost:3457/ws/app` → **immediate** `schema-version` → `agents-sync` → `repositories-sync` → `sessions-sync` (all 4 messages received, normal)
3. Browser page context `new WebSocket('ws://localhost:15173/ws/app')` (via vite proxy) → **12s wait, `readyState === 0` (CONNECTING), no open/message/close/error events**
4. Same manual WS to `ws://localhost:13457/ws/app` (host-mapped port, bypasses vite) → **26ms OPEN, all 4 messages received**
5. Container-internal probe to vite's own port (5173 internal) `/ws/app` also hangs (readyState 0 after 6s)
6. Container restart reproduces identically (persistent, not transient)

## Suspected cause

The vite `/ws` proxy entry lacks `changeOrigin: true` (`/api` proxy has it and works). Unverified but consistent with the "backend upgrade succeeds, response never propagates back" symptom.

## Impact

- **Docker dev stack's core "AI agents can drive it end-to-end" premise (per `docker/README.md`) is broken** — any browser-driven verification via the dev stack cannot progress past dashboard load
- **Browser QA delegates using the Docker dev stack are effectively blocked** — must fall back to production-bundle `docker/docker-compose.verification.yml` (no vite) or host-side `scripts/dev-multiuser.sh` (owner-driven)
[...]

Issue #1210, opened 2026-07-18
bug(json-repo): JsonSessionRepository.update() silently ignores recoveryState/orphanedAt/orphanedReason (breaks orphan-marking on JSON backend)


ms2sato:
## Context

Discovered as incidental finding during Sprint 2026-07-18b PR [#1207](https://github.com/ms2sato/agent-console/pull/1207) (#1197 Part A, `killOrphanWorkers` fix) R1 test rewrite.

While rewriting `session-manager.test.ts` per CR's R1-d proposal (persist session with `recoveryState: 'orphaned'` + dead `serverPid` → `initializeSessions` preserves it, leaving `cleanupOrphanProcesses` as the sole resolver caller), the delegate discovered that **`JsonSessionRepository.update()` silently ignores the `recoveryState`, `orphanedAt`, and `orphanedReason` fields** — the update returns success but the persisted JSON is unchanged for these fields (no-op).

Workaround in the R1 test: `findById()` + `save()` (full-record replace) instead of `update()`.

## Impact

**Production impact (JSON repository backend):** `detectOrphans()` writes recovery state via `update()`, so on JSON repo deployments **orphan marking silently fails**. This can cause:
- Orphaned sessions never entering `recoveryState: 'orphaned'` state
- `initializeSessions`'s orphan-preservation branch never triggering (all sessions treated as either healthy or missing)
- `cleanupOrphanProcesses` never running its cross-user orphan-kill path (which is the very code path #1197 Part A was fixing!)

This may explain why the #1197 Part A umbrella hardening only surfaced defects in the SQL repo path and not the JSON repo path — the JSON path never actually populates the state that would exercise those defects in the first place. This is a **silent scope-limiting bug**, not an active corruption.

## Repro (delegate-observed)

In test setup:
1. Create session via `sessionRepository.save(session)`
2. `sessionRepository.update(sessionId, { recoveryState: 'orphaned', orphanedAt: Date.now(), orphanedReason: 'crashed' })` — returns success
3. `sessionRepository.findById(sessionId)` — returned session **does NOT have** the orphan fields set

Fix path candidates:
- Explicit field allowlist in `JsonSessionRepository.update()` includes `recoveryState`, `orphanedAt`, `orphanedReason`
- OR generic passthrough (accepts any field defined in the Session type) — depends on repo's serialization convention

## Scope note

[...]

PR #1228, opened 2026-07-27
fix(docker-dev-stack): own soft-fail for git-absence in install-hooks.mjs + missing embedded-agent volume

### LINKED ISSUE #1214: bug(docker-dev-stack): bun install --frozen-lockfile exits 1 on fresh container start due to postinstall .git resolution
## Summary

`docker/docker-compose.yml`'s dev stack (`dev-entrypoint.sh`) always exits with code 1 on a fresh container start, because `bun install --frozen-lockfile` returns exit code 1 even though all real dependency packages install successfully. `set -eu` in the entrypoint then kills the container before the dev servers (`vite` + `bun --watch`) ever start.

This contradicts the documented contract in `CLAUDE.md`:

> `bun install` runs a `postinstall` step that installs the repository's git hooks... The step fails soft when `.git` is not present (e.g., Docker images, non-git checkouts) so it never breaks `bun install` itself.

In practice, inside the Docker dev stack, it DOES break `bun install`.

## Reproduction

```bash
DEV_CLIENT_PORT=15173 DEV_SERVER_PORT=13457 docker compose -f docker/docker-compose.yml up -d
# wait a few seconds
docker ps -a --filter "name=agent-console-dev" --format "table {{.Names}}\t{{.Status}}"
# -> Exited (1)
docker compose -f docker/docker-compose.yml logs
```

Representative log tail (clean volumes, fresh install):

```
+ typescript@5.9.3
...
1318 packages installed [5.53s]
Failed to install 1 package
```

Manually reproducing inside the container as the `agentconsole` service user (bypassing the entrypoint's `set -eu`), with `--verbose`, shows the "1 package" summary appears immediately after the `agent-console` root package's own lifecycle scripts run:

```
[Scripts] Starting scripts for "agent-console"
$ bun scripts/check-bun-version.mjs
$ bun scripts/install-hooks.mjs || echo 'hooks:install skipped (likely no git or non-interactive env)'
hooks:install — `git rev-parse --git-common-dir` failed:
fatal: not a git repository: /var/lib/agent-console/source-repos/agent-console/.git/worktrees/wt-007-szyw

hooks:install skipped (likely no git or non-interactive env)
[Scripts] Finished scripts for "agent-console"

Failed to install 1 package
[210.00ms] done
```

Confirmed this is the `.git`-not-resolvable case working as intended by `scripts/install-hooks.mjs` (the repo's bind mount at `/app` only contains the worktree checkout, not the main repo's `.git` directory the worktree's `.git` file references — expected in a Docker bind mount). What's NOT expected is that `bun install`'s own exit code goes non-zero for this soft-failed postinstall.

Isolated checks that rule out a real dependency/network/permission problem:
- `bun scripts/check-bun-version.mjs` run standalone: exit 0.
- `bun scripts/install-hooks.mjs || echo done` run standalone (the exact postinstall command string): exit 0.
- `bun install --ignore-scripts`: still reports "Failed to install 1 package" (rules out the hooks script itself being the literal cause bun tracks; also rules out a native-module postinstall like `bun-pty`/`esbuild` failing, since scripts were disabled entirely).
- `bun install --frozen-lockfile --concurrent-scripts 1 --verbose`: only the `agent-console` root package's scripts ever run, and the failure is still reported right after — no other package's lifecycle script executes or fails.
- Manually running each `esbuild` variant's `install.js`, and the postinstall command string directly: all exit 0.
- `BUN_CONFIG_MINIMUM_RELEASE_AGE=0` (disabling the `bunfig.toml` supply-chain age gate): no change — rules out `minimumReleaseAge` as the cause.

This points at Bun 1.3.8's own lifecycle-script exit-status bookkeeping for a `cmdA || cmdB` compound `postinstall` script, not at anything specific to this repo's dependency graph.

## Impact

- Every fresh `docker compose -f docker/docker-compose.yml up` (or `up -d`, or a volume-cleared restart) fails to start the dev servers at all, blocking the Docker dev stack's stated purpose: "AI agents can drive it end to end... no privilege elevation, no password prompt" (`docker/README.md`).
- This was discovered as a compound blocker while investigating #1211 (vite `/ws` proxy hang) — a QA-only, uncommitted entrypoint override (`bun install --frozen-lockfile || echo ...` to ignore the exit code) was needed to get the stack running at all for that investigation's Browser QA.
- Host-side `scripts/dev.sh` / `scripts/dev-multiuser.sh` are NOT affected in practice, since they don't typically hit a fresh `bun install` against an unresolvable `.git` the same way.

## Suggested directions (not yet investigated in depth)

- Confirm whether Bun's lifecycle-script runner mis-tracks the exit status of a `cmdA || cmdB` postinstall string (possible Bun bug/limitation), vs. some other root cause specific to how `install-hooks.mjs`'s failure interacts with bun's script summary.
- If confirmed, either restructure `install-hooks.mjs`'s soft-fail invocation to avoid the shell `||` pattern (e.g., handle the git-repo-check failure inside the script itself with `process.exit(0)`), or bypass/ignore `bun install`'s own exit code specifically inside `dev-entrypoint.sh` if this is deemed acceptable for a dev-only container context.

## Refs

Refs #1211 (found as a compound blocker while investigating that issue; separate root cause, filed independently per repo scope-hygiene convention).

---

## Acceptance Criteria (architect-authored, 2026-07-27)

Current code has **two** `exit(1)` paths in `scripts/install-hooks.mjs` that both fire when git is absent: `resolveRepoRoot()` (`git rev-parse --git-common-dir` failure) and `resolveHooksDir()` (`git rev-parse --git-path hooks` failure).

### 1. Soft-fail ownership moves into the script; the shell `||` is removed

The `|| echo` in `package.json`'s `postinstall` currently owns soft-fail, and it is not working under Bun 1.3.8 (the lifecycle bookkeeping appears to count the left-hand script's exit 1 as a package failure rather than honoring the compound's final exit status). Per `design-principles.md` "enforce constraints through structure, not convention", **the script itself must own the contract** that `CLAUDE.md` already documents ("fails soft when `.git` is not present ... so it never breaks `bun install` itself").

- Change both git-absence paths from `process.exit(1)` to an early return / `null` result that leads to `process.exit(0)`, with a clear `console.error` warning (`not a git repo; skipping hook install`).
- Simplify `postinstall` to `bun scripts/install-hooks.mjs`. With the script exiting 0, `bun install` stays green regardless of whether the Bun bookkeeping hypothesis is true.
- **No `--postinstall` / `--soft` context flag.** "This is not a git repository, so hooks cannot be installed" is the same fact whether reached via `postinstall` or a manual `bun run hooks:install`. Exit 0 plus an explicit stderr warning informs both contexts without breaking either — this is not a silent skip.

### 2. Soft-fail is strictly limited to the two git-resolution conditions

Failures *after* git resolves — symlink failure, copy failure, target exists with different content (overwrite refusal) — **stay hard failures**. These represent the hook genuinely not being installed; swallowing them creates a silent gap where a developer believes the `commit-msg` language check is active when it is not (the same silent-no-op class as #1210).

"Make everything soft in the postinstall path" is **rejected**: an unknown non-Docker failure would then let `bun install` pass with the hook missing.

### 3. Bun behavior is recorded, not fixed here

The fix makes the script exit 0 on its own, so it no longer depends on Bun's compound-script bookkeeping — it is correct whether or not the hypothesis holds. Record **why the `|| echo` was removed** in the PR body, with a minimal reproduction of the Bun 1.3.8 behavior, so a future reader can trace the decision. Filing upstream is out of scope (record and defer).

### 4. Verification

**Unit** (`scripts/__tests__/install-hooks.test.mjs`) — both polarities are required:

- **(a) Soft path polarity** — with git unavailable (stub `spawnSync`, or remove `git` from `PATH`): fails (exit 1) before the fix, exits 0 with the stderr warning after.
- **(b) Hard path polarity — mandatory, prevents over-softening** — assert that symlink/copy failure or a conflicting existing target still exits non-zero. Without (b), a future regression to "everything soft" is undetectable.

**Docker smoke** (per `os-environment-coupling.md`):

- `docker compose -f docker/docker-compose.yml up` must get **past `dev-entrypoint.sh`'s `bun install` step and reach the point where the server process begins spawning**.
- **Do not gate on full healthy.** Issue #1211 (vite `/ws` proxy hang → Dashboard stuck on "Loading sessions...") is a separate, currently-open defect in the same stack; waiting for healthy would make this smoke false-fail on #1211's hang. What #1214 broke is specifically "`set -eu` kills the container at `bun install` before anything starts", so the smoke's scope ends at that entrypoint death being resolved. Explicitly note `/ws` health as out of scope, tracked by #1211.

### Implementation scope

- `scripts/install-hooks.mjs` — soften the two git-absence branches; keep install-operation failures hard.
- `package.json:25` — drop `|| echo` from `postinstall`.
- Two unit tests (soft polarity, hard polarity) + the Docker smoke with the #1211 carve-out.
- PR body: Bun 1.3.8 minimal reproduction note + rationale for retaining hard failures.

Reversibility: **reversible** (repairs a broken postinstall; behavior in git-present environments is unchanged).

ms2sato:
## Scope note (read this first)

This PR contains **two independent fixes**, bundled together because the second was needed to actually demonstrate the first. Issue #1214's acceptance criteria require a Docker smoke test proving `docker compose up` gets past `bun install` on a fresh checkout. While building that smoke, I hit a **second, unrelated defect** that also breaks `bun install` on a fresh checkout — without fixing it too, the smoke for #1214 cannot pass, so the PR's own verification would be unsatisfiable if split. Commits are kept separate for review clarity:

- Commit 1 (`3c4d8e19`) + commit 2 (`4e136971`): the actual #1214 fix — `install-hooks.mjs` soft-fail ownership.
- Commit 3 (`4e152d05`): the second, independent docker-compose.yml fix, needed only to make the smoke executable.

## Merge disposition note (2026-07-27, updated)

**CR state (initial)**: local CodeRabbit CLI returned `automatic_login_failed` (headless-worktree auth failure — needs interactive browser login, not available in this delegated session). GitHub-side bot posted a rate-limit warning at PR open ("Review limit reached... Next review available in: 15 minutes"). Both channels unavailable for this commit set at the time.

**Resolution**: per Orchestrator direction, waited out the bot's rate-limit countdown rather than using the documented fallback disposition (no urgency to spend the exception since `main` was independently red on unrelated Issue #1225 at the time). Retriggered with `@coderabbitai review` after the countdown; a genuine walkthrough landed (CodeRabbit status `SUCCESS` / "Review completed" at 02:13:40Z, accurately describing both fixes in this PR). Final state: inline actionable comments = 0, pre-merge checks 4/5 passed (1 non-blocking "Docstring Coverage" warning — not applicable to this repo's shell/JS utility script conventions, no actual finding), `reviewDecision` empty (known CodeRabbit behavior for a zero-finding pass — see the `coderabbit-ops` skill's "walkthrough-exists rubric," treated as clean-equivalent). A separate independent code-appropriateness audit from the Architect also returned CLEAN on this diff.

CodeRabbit + Architect layers are both clean. Per Orchestrator: this PR still requires **owner approval** before merge (touches `package.json`, `scripts/install-hooks.mjs`, and `docker/*` behavior — not test/docs-only), gated additionally on `main` going green (Issue #1225, unrelated, in progress).

## Summary

`docker/dev-entrypoint.sh`'s `bun install --frozen-lockfile` returned exit 1 on every fresh Docker dev-stack start, even though every dependency installed successfully, because `set -eu` then killed the container before the dev servers started.

### Fix 1: `scripts/install-hooks.mjs` soft-fail ownership (the #1214 fix)

`package.json`'s `postinstall` delegated soft-fail to a shell `bun scripts/install-hooks.mjs || echo '...'`. Under Bun 1.3.8, this does not reliably work: **minimal reproduction** —

```bash
# Standalone, the exact postinstall command string: exit 0, works fine.
$ bun scripts/install-hooks.mjs || echo done; echo $?
hooks:install — not a git repository; skipping hook install.
...
done
0

# But via `bun install` itself, inside a directory where git is unresolvable,
# the postinstall's own exit 0 does not stop bun from recording the
# left-hand script's exit 1 as a package failure:
$ bun install --frozen-lockfile
...
$ bun scripts/install-hooks.mjs || echo 'hooks:install skipped ...'
hooks:install — `git rev-parse --git-common-dir` failed: ...
hooks:install skipped ...
1318 packages installed [...]
Failed to install 1 package     # <- non-zero exit despite the compound's own exit 0
```

Per `design-principles.md` ("enforce constraints through structure, not convention"), the fix moves soft-fail ownership into the script itself rather than depending on Bun's compound-script exit-status bookkeeping:

- `resolveRepoRoot()`'s `git rev-parse --git-common-dir` failure (no git repo resolvable at all — e.g. a Docker bind mount of a worktree without its main repo, or a non-git checkout) now returns early and the script exits **0** with a stderr warning, instead of exit 1.
- `postinstall` simplified to `bun scripts/install-hooks.mjs` (the shell `||` removed entirely) — `bun install`'s exit code no longer depends on Bun's compound-script bookkeeping at all, so the fix holds regardless of whether the exact bookkeeping hypothesis above is the full story.
- **Asymmetric refinement** (post-review): only `resolveRepoRoot()`'s failure is soft. `resolveHooksDir()`'s `git rev-parse --git-path hooks` failure — reached only *after* a git repo was already confirmed to exist — stays a **hard failure** (exit 1). At that point "no repo" has already been ruled out, so a failure here means something abnormal (corrupted repo, git internal error), which should surface loudly rather than silently skip installing the commit-msg language check (the same silent-no-op class as #1210).
- All other failures (symlink failure, copy failure, conflicting existing target) are unchanged — still hard failures, since those mean the hook genuinely didn't install.

Filing the exact Bun 1.3.8 compound-script bookkeeping behavior upstream is out of scope for this PR (recorded above for a future reader; the fix does not depend on the hypothesis being correct).

### Fix 2: `docker-compose.yml` missing `packages/embedded-agent` volume (discovered by the smoke test)

Running the required Docker smoke against a **genuinely fresh checkout** (`git clone`, no pre-existing `node_modules` — exactly the "fresh container start" scenario in #1214) surfaced a second failure, independent of Fix 1:

```
ENOENT: No such file or directory: failed to symlink dependencies for package: @agent-console/embedded-agent@workspace:packages/embedded-agent (symlink)
...
Failed to install 1 package
```

Root cause: `docker/docker-compose.yml`'s dev-stack volumes list has named `node_modules` volumes for root/client/server/shared/integration, but not for `packages/embedded-agent` (a workspace member via `packages/*`, evidently added after this list was last updated). Without a dedicated volume, bun's workspace-symlink step for `@agent-console/embedded-agent` falls through to the bare bind mount and fails with `ENOENT` on a truly fresh checkout.

Confirmed fully independent of Fix 1 / git resolution:
- Reproduces with `--ignore-scripts` (all lifecycle scripts disabled).
- Reproduces with a fully resolvable git repo (`git clone --local`, not a broken worktree pointer).
- Does NOT reproduce when the same tree is copied to a container-native (non-bind-mounted) directory — only manifests over a bind mount + missing volume.
- My own dev worktree masked this bug because `packages/embedded-agent/node_modules` already existed there from an earlier host-side `bun install` — this is exactly the kind of state that let the defect survive undetected (see `os-environment-coupling.md`).

Fix: added `embedded-agent-node-modules` to both `docker-compose.yml`'s volumes list/mounts and `dev-entrypoint.sh`'s stage-1 `chown`/`chmod` loop (a named volume mounted over a path the entrypoint doesn't fix up is created root-owned, which would trade the ENOENT for an EACCES). Also updated `docker/README.md`'s volume-name list to match.

This is the Docker smoke test doing exactly what it's for — it caught a second real defect on the same code path that a narrower verification would have missed.

## Verification

- `bun run test:scripts` — 485 pass, 0 fail (scripts/ + hooks/ + skills tests).
- `bun run test` (full suite) — 1 pre-existing, unrelated failure: `MultiUserMode > Issue #918 ... does NOT inject any SSH_AUTH_SOCK` fails locally because the real host's 1Password SSH agent socket leaks into the test environment. Confirmed via baseline comparison (`git diff` isolate + `git checkout` to unmodified `main`-tip code, re-ran the same test file, same failure) — pre-existing and unrelated to this change, not introduced here. `main` is also independently known-red right now on two unrelated `@agent-console/integration` `System API > openPath` cross-file-pollution failures (Issue #1225, owned by another workstream) — not observed in my local run but noted per current sprint status.
- `bun run typecheck` — clean.
- `node .claude/skills/orchestrator/preflight-check.js` (run after commit, before push) — clean: no coverage gaps, no rule/skill duplication, no language violations, no blame-shift comment references.
- **Docker smoke** (both AC carve-outs honored: no `--soft`/`--postinstall` context flag anywhere; smoke asserts entrypoint gets past `bun install` and the server begins spawning, does NOT gate on full healthy since Issue #1211's `/ws` proxy hang is a separate open defect in the same stack):
  - Built a genuinely fresh `git clone --local` checkout (no pre-existing `node_modules`) of this branch.
  - `DEV_CLIENT_PORT=15173 DEV_SERVER_PORT=13457 docker compose -f docker/docker-compose.yml up --build -d` against that fresh checkout.
  - Result: container reached `Up ... (healthy)` — `bun install` exited 0, both dev servers (vite + `bun --watch` server) started, and the compose healthcheck (which itself only checks `/api/config` + the vite root, not `/ws`) passed. This exceeds the AC's minimum bar (get past `bun install` + server spawning) — full healthy was reached, and #1211's `/ws` hang did not block it since the healthcheck doesn't probe that path.
  - Before the docker-compose.yml fix, the identical fresh-checkout smoke reproduced the `ENOENT ... embedded-agent` failure described above (exit 1, container dies at `bun install`), confirming the fix was necessary.

## Unit tests

`scripts/__tests__/install-hooks.test.mjs` (12 tests total, TDD-verified: each new/changed assertion fails on unmodified code and passes after its corresponding fix):

- Soft-path polarity: `install-hooks.mjs` exits 0 with a stderr warning when no `.git` is reachable at all.
- `package.json#postinstall` (now `bun scripts/install-hooks.mjs`, no `||`) exits 0 in the same no-git scenario.
- Hard-path polarity (mandatory, prevents future over-softening): a conflicting existing hook target still exits 1 once git resolves.
- New: a `git rev-parse --git-path hooks` failure stays hard (exit 1) even when the repo itself was found (stubs `git` on `PATH` to isolate this from the "no repo at all" case).

Closes #1214

## Summary by CodeRabbit

* **New Features**
  * Improved development container support by persisting the embedded agent’s dependencies across restarts.

* **Bug Fixes**
  * Installation hooks now complete successfully when Git is unavailable, while still reporting a warning.
  * Hook installation continues to fail clearly for genuine configuration or repository errors.

* **Documentation**
  * Clarified instructions for removing Docker development volumes.

* **Tests**
  * Added coverage for Git-unavailable scenarios and confirmed that genuine hook installation failures remain visible.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -89,8 +89,9 @@ and must never be reachable from the LAN.
 
 Named volumes (all under the `agent-console-dev` compose project):
 `root-node-modules`, `client-node-modules`, `server-node-modules`,
-`shared-node-modules`, `integration-node-modules`, `dev-data`. Remove them
-with `docker compose -f docker/docker-compose.yml down -v` for a from-scratch
+`shared-node-modules`, `integration-node-modules`,
+`embedded-agent-node-modules`, `dev-data`. Remove them with
+`docker compose -f docker/docker-compose.yml down -v` for a from-scratch
 start.
 
 ### Driving the stack as an AI agent
--- a/docker/dev-entrypoint.sh
+++ b/docker/dev-entrypoint.sh
@@ -28,6 +28,7 @@ if [ "${1:-}" != "stage2" ]; then
     "$APP_DIR/packages/server/node_modules" \
     "$APP_DIR/packages/shared/node_modules" \
     "$APP_DIR/packages/integration/node_modules" \
+    "$APP_DIR/packages/embedded-agent/node_modules" \
     /var/lib/agent-console; do
     chown "$SERVICE_USER:$SERVICE_GROUP" "$dir"
     chmod 2775 "$dir"
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -68,6 +68,7 @@ services:
       - server-node-modules:/app/packages/server/node_modules
       - shared-node-modules:/app/packages/shared/node_modules
       - integration-node-modules:/app/packages/integration/node_modules
+      - embedded-agent-node-modules:/app/packages/embedded-agent/node_modules
       # Dev data root (DB, jwt-secret, worktrees) persists across
       # `docker compose down` / `up` cycles.
       - dev-data:/var/lib/agent-console
@@ -87,4 +88,5 @@ volumes:
   server-node-modules:
   shared-node-modules:
   integration-node-modules:
+  embedded-agent-node-modules:
   dev-data:
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "check:pty-fd-leak": "bun scripts/smoke/check-pty-fd-leak.ts",
     "hooks:install": "bun scripts/install-hooks.mjs",
     "preinstall": "bun scripts/check-bun-version.mjs",
-    "postinstall": "bun scripts/install-hooks.mjs || echo 'hooks:install skipped (likely no git or non-interactive env)'",
+    "postinstall": "bun scripts/install-hooks.mjs",
     "lint": "bun run lint:structure",
     "lint:structure": "bun run lint:deps && bun run lint:cycles && bun run lint:unused",
     "lint:deps": "depcruise --config .dependency-cruiser.cjs --ignore-known .dependency-cruiser-known-violations.json packages",
--- a/scripts/__tests__/install-hooks.test.mjs
+++ b/scripts/__tests__/install-hooks.test.mjs
@@ -279,6 +279,46 @@ describe('scripts/install-hooks.mjs hard-failure polarity (Issue #1214)', () => 
     expect(result.status).toBe(1);
     expect(result.stderr).toContain('exists with different content');
   });
+
+  it('a `git rev-parse --git-path hooks` failure stays hard (exit 1) even though the repo itself resolved', () => {
+    // Stub `git` on PATH so `--git-common-dir` succeeds (as a real repo
+    // would) but `--git-path hooks` fails, simulating a corrupted /
+    // abnormal repo state distinct from "no git repo at all". Only
+    // resolveRepoRoot()'s git-absence branch is soft; a hooks-path failure
+    // AFTER the repo resolves must surface loudly, not be swallowed as if
+    // the repo were simply missing (that would silently skip installing
+    // the commit-msg language check — the #1210 silent-no-op class).
+    const stubBin = mkdtempSync(join(tmpdir(), 'install-hooks-stubgit-'));
+    const gitStub = join(stubBin, 'git');
+    writeFileSync(
+      gitStub,
+      [
+        '#!/bin/sh',
+        'if [ "$1" = "rev-parse" ] && [ "$2" = "--git-common-dir" ]; then',
+        '  echo "/tmp/fake-common-dir"',
+        '  exit 0',
+        'fi',
+        'if [ "$1" = "rev-parse" ] && [ "$2" = "--git-path" ]; then',
+        '  echo "simulated corrupted repo" 1>&2',
+        '  exit 1',
+        'fi',
+        'exit 1',
+        '',
+      ].join('\n'),
+    );
+    chmodSync(gitStub, 0o755);
+
+    try {
+      const result = spawnSync('bun', [INSTALL_SCRIPT], {
+        encoding: 'utf8',
+        env: { ...process.env, PATH: `${stubBin}:${process.env.PATH}` },
+      });
+      expect(result.status).toBe(1);
+      expect(result.stderr).toContain('git rev-parse --git-path hooks');
+    } finally {
+      rmSync(stubBin, { recursive: true, force: true });
+    }
+  });
 });
 
 describe('scripts/install-hooks.mjs invoked via package.json#postinstall in a git repo (Issue #735)', () => {
--- a/scripts/__tests__/install-hooks.test.mjs
+++ b/scripts/__tests__/install-hooks.test.mjs
@@ -187,7 +187,7 @@ describe('scripts/install-hooks.mjs', () => {
   });
 });
 
-describe('scripts/install-hooks.mjs in a non-git environment (Issue #735)', () => {
+describe('scripts/install-hooks.mjs in a non-git environment (Issue #735, #1214)', () => {
   let nonGitSandbox;
 
   beforeEach(() => {
@@ -220,23 +220,22 @@ describe('scripts/install-hooks.mjs in a non-git environment (Issue #735)', () =
     return { ...process.env, GIT_CEILING_DIRECTORIES: nonGitSandbox };
   }
 
-  it('install-hooks.mjs exits non-zero when no .git is reachable', () => {
+  it('install-hooks.mjs exits 0 with a stderr warning when no .git is reachable (Issue #1214)', () => {
     const result = spawnSync(
       'bun',
       [join(nonGitSandbox, 'scripts/install-hooks.mjs')],
       { encoding: 'utf8', cwd: nonGitSandbox, env: envWithCeiling() },
     );
-    expect(result.status).toBe(1);
-    expect(result.stderr).toContain('git rev-parse');
+    expect(result.status).toBe(0);
+    expect(result.stderr).toContain('not a git repository');
   });
 
-  it('package.json#scripts.postinstall fails soft (exit 0) when no .git is reachable', () => {
+  it('package.json#scripts.postinstall exits 0 with no shell `||` needed when no .git is reachable (Issue #1214)', () => {
     const pkg = JSON.parse(
       readFileSync(resolve(REPO_ROOT, 'package.json'), 'utf8'),
     );
     const postinstallCmd = pkg.scripts.postinstall;
-    expect(postinstallCmd).toContain('bun scripts/install-hooks.mjs');
-    expect(postinstallCmd).toContain('||');
+    expect(postinstallCmd).toBe('bun scripts/install-hooks.mjs');
 
     const result = spawnSync('sh', ['-c', postinstallCmd], {
       encoding: 'utf8',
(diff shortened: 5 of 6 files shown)


ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

Issue #1242, opened 2026-07-29
MCP delegate_to_worktree: sentinel output never reaches the server in the create path — agent never starts (restart path works)


ms2sato:
## Summary

With #1237 and #1241 deployed and **working as designed**, MCP `delegate_to_worktree` still fails to start the agent — but through a **different, pre-existing mechanism** than #1234: in the create path, the PTY's output (including the login-shell sentinel) **never reaches the server-side sentinel handler at all**, so the (now short) agent command is never injected. The same worker restarted via `POST /api/sessions/:sessionId/workers/:workerId/restart` works every time, and since #1241 redelivers the prompt file, restart is a full workaround.

This issue was masked by #1234: both bugs produce "delegated agent never starts", but #1234 dies *after* sentinel detection (typeahead truncation) while this one dies *before* it (sentinel never observed). The 0-byte worker output log is the discriminator — pre-sentinel output is swallowed by design (`worker-manager.ts` `setupWorkerEventHandlers` early-returns before `bufferOutput` until the sentinel matches), so "log stays 0 bytes forever + interactive bash alive with no child" means the server never matched the sentinel.

Verified against deployed build 2026-07-29 08:52 (deployed `worker-manager.ts` extracted from `server.js.map` is byte-identical to repo HEAD `e84e19d4`). macOS (darwin 25.5.0), single-user mode, launchd-managed server, bun 1.3.5, default `PTY_PROVIDER` (= `bun-terminal`).

## Observed facts

Failing delegation (2026-07-29, session `76e425e3`, worker `f9edd7d1`, PTY pid 66655):

- Worktree, session, agent + git-diff workers all created; MCP call returns success; no `level>=40` server log entries for the worker.
- Prompt file `~/.agent-console/prompts/f9edd7d1-….prompt` written correctly (0600, 5,480 bytes) — **#1237 works**.
- PTY process tree: `bash` (plain, post-`exec $SHELL`, `Ss+`) with **no child** — the two-stage login shell ran to completion, so `echo <sentinel>` **did execute on the tty**.
- Worker output log: **0 bytes**, still 0 after minutes. UI terminal blank.
- `POST …/restart` on the same worker: agent starts within seconds, `ps` shows `claude <full prompt>` (prompt redelivered via #1241 — argv confirms verbatim content), output log grows normally.

[...]

Issue #1234, opened 2026-07-27
Delegated agent never starts: sentinel typeahead injection exceeds tty canonical input buffer for long initialPrompt


ms2sato:
## Summary

MCP `delegate_to_worktree` creates the worktree, session, and workers, but the delegated agent (claude) **never starts**: the PTY's interactive bash sits at a prompt with no child process and the worker terminal stays blank forever.

Root cause: the login-shell sentinel mechanism injects the full agent command — including the delegation `initialPrompt` (up to 5000 chars) shell-quoted into one line — as **terminal typeahead** via `pty.write(pendingCommand + '\r')`. The write lands while the tty is still in canonical mode (between the login shell's sentinel echo and the interactive shell's readline taking over). On macOS the canonical-mode input line buffer is ~1KB; input beyond that is discarded, so the command line never executes. Short commands survive; long ones deterministically die.

Verified against deployed build 2026-07-27 (repo `a4437630`; sentinel machinery from `41e93cd8`).

## Production evidence

Observed on macOS (darwin 25.5.0, single-user mode, `$SHELL` = homebrew bash):

- 3/3 MCP delegations with a ~3.5KB prompt failed identically: worktree + session + workers created, PTY bash alive (`Ss+`, post-`exec $SHELL`, **no child**), worker terminal blank.
- 5/5 agent-worker **restarts** succeeded — `restartAgentWorker` passes no `initialPrompt`, so the injected command is short.
- UI-created sessions and terminals: unaffected.
- `templateVars` ruled out (failure reproduces without it). Browser/WS layer ruled out (server receives the WS fine; the swallowed pre-sentinel gate means a missed command leaves the output stream empty, matching the blank terminal).

## Deterministic reproduction

Same spawn shape as `spawnDirectPty` + `activateAgentWorkerPty` (`sh -c "exec $SHELL -l -c 'echo <sentinel>; exec $SHELL'"`, then on sentinel detection `pty.write(command + '\r')`), via bun-pty directly:

| Injected command | Result |
| --- | --- |
| 22 bytes (`echo REPRO_DONE_SHORT`) | executes ✅ |
| 3,523 bytes (same + long trailing comment) | **never executes** ❌ — echoed input truncates at ~2KB, no prompt line, marker never printed |

The repro also shows the injected line being echoed twice (canonical echo + readline redisplay) even in the passing case — the same underlying "typed before readline is ready" window.

[...]

PR #1243, opened 2026-07-29
fix: close PTY pre-attach byte-loss windows, add sentinel watchdog

ms2sato:
## Summary

- Close two proven silent byte-loss windows in `bunTerminalProvider` / `BunTerminalPtyAdapter` (`packages/server/src/lib/pty-provider.ts`) with a single bounded (64 KiB, keep-first-bytes) pre-attach buffer, shared between the provider closure and the adapter, created **before** `Bun.spawn()`:
  1. Native `data` callback firing before the `adapter` variable is assigned (the callback is handed to `Bun.spawn` before it returns).
  2. `_emitData` firing before the first `onData()` attach.
  This extends the same late-attach-safe replay contract `onExit` already had to `data`.
- Add a 15s sentinel watchdog in `worker-manager.ts`'s `setupWorkerEventHandlers`: if an agent worker's login-shell sentinel is never detected, log an `ERROR` with the PTY adapter's native data-callback fire count, buffered/dropped byte counts, and a 256-byte sample of pre-sentinel bytes actually received — instead of leaving the worker silently stuck with a 0-byte output log forever. Cleared on sentinel detection, on natural worker exit, and (via `disposables`) on managed kill; two negative tests assert no spurious `ERROR` in either case.
- Add `scripts/smoke/check-pty-early-output.ts`: a real-PTY smoke that spawns a real child emitting output immediately, deliberately delays the `onData` attach past an event-loop turn, and asserts the early bytes are not lost. Wired as `bun run check:pty-early-output`.

Fix plan followed exactly per Issue #1242's architect-redlined plan; no design decisions were re-opened.

## Epistemic status

> This PR fixes two proven silent byte-loss windows in the bun-terminal PTY provider and adds sentinel-timeout instrumentation. It does NOT claim these windows are the cause of the macOS create-path failure (#1242). One instrumented delegation on the owner environment (see the issue's checkpoint section) decides that; this issue stays open until then.

Refs #1242. This PR does not close the issue.

### Full suite (`bun run typecheck && bun run test`)

Typecheck: exit 0, all four packages report `Done`.

Full test suite tail:

```
[server] 3720 pass / 1 skip / 0 fail -- 10209 expect() calls -- Ran 3721 tests across 164 files
[embedded-agent] 287 pass / 0 fail -- 637 expect() calls -- Ran 287 tests across 26 files
[client] 2075 pass / 0 fail -- 4361 expect() calls -- Ran 2075 tests across 142 files
[shared] 596 pass / 0 fail -- 891 expect() calls -- Ran 596 tests across 22 files
[integration] 73 pass / 0 fail -- 424 expect() calls -- Ran 73 tests across 27 files
test:scripts (scripts/ + .claude hooks/skills) 521 pass / 0 fail -- 1168 expect() calls -- Ran 521 tests across 17 files
TEST_EXIT: 0
```

(Counts above are from the final commit, after the post-review copy-on-push fix below; +1 test vs. the original commit.)

### `bun run check:pty-fd-leak` (required by `test-trigger.md`, `pty-provider.ts` touched)

```
==> ran 100 spawn/kill cycles via bunTerminalProvider
==> ptmx fd count (this process, via /proc/<pid>/fd)
  OK    ptmx fd count non-increasing (before=0, after=0)
==> kernel pty counter (/proc/sys/kernel/pty/nr)
  OK    kernel pty counter non-increasing (before=44, after=44)
PASSED: 2 assertion(s) passed
```

### Polarity — `pty-provider.test.ts`'s new `describe('pre-attach data buffer (Issue #1242)', ...)` (9 tests)

Verified by reverting only `pty-provider.ts` (via a diff/patch apply-and-revert, not `git stash`, to avoid the shared-worktree stash collision risk) and running the file against the unbuffered implementation, then restoring and re-running:

- **7/9 tests polarity-flip** (fail against the unbuffered adapter, pass against the fix): "buffers chunks that arrive before onData is attached...", "preserves a UTF-8 multi-byte sequence split across the pre-attach boundary", "caps buffered bytes at 64 KiB...", "keeps exactly the cap with zero drops...", "does NOT replay on a second onData attach...", "getDataDiagnostics().fireCount counts every native data callback fire...", "live chunks after the first attach still flow straight through onData...". Pre-fix run: 7 fail / 25 pass (2 of the 32 pre-existing + new tests already existed and pass unaffected). Post-fix (restored) run: 149/149 pass across both touched test files.
- **2/9 are new-mechanism contract tests, not bug-polarity tests** (per architect ruling — no pre-fix counterpart mechanism exists to flip against):
  - `"does NOT deliver anything on attach when no pre-attach bytes arrived (empty buffer is not flushed as an empty string)"` locks the property *an empty pre-attach buffer must never synthesize a spurious empty-string `onData` event*. Pre-fix, there is no buffer/flush concept at all, so this property has no pre-fix analog to flip against — it guards a NEW code path (`PreAttachDataBuffer.flush()` / the `if (buffered.length > 0)` guard in `onData()`) that could regress independently of the byte-loss bug (e.g. a future refactor that always calls `listener('')` on attach).
  - `"dispose() before any onData attach discards buffered bytes (no retention, no replay)"` locks the property *`dispose()` must free the pre-attach buffer so it can never leak into a later listener*. Pre-fix, `dispose()` has nothing to discard (no buffer exists), so again there is no pre-fix mechanism to flip against — it guards `PreAttachDataBuffer.discard()`, a NEW method.

  Both properties are genuinely unwritable against the unfixed code (there is no buffer to assert boundary behavior on), not cases where a flip was possible and skipped.

### `scripts/smoke/check-pty-early-output.ts` — real-PTY polarity (Fix plan 3)

Pre-fix run (unbuffered `pty-provider.ts`, same revert/restore as above):

```
==> running 20 early-output cycles via bunTerminalProvider (Issue #1242), each with a deliberately-late (50ms) onData attach
  FAIL  cycle 0: marker not observed after a 50ms late onData attach; diagnostics=undefined; captured output: ""
  ... (all 20 cycles FAIL identically)
FAILED: 20/20 cycle(s) lost the early marker
```
Exit code: 1

Post-fix run:

```
==> running 20 early-output cycles via bunTerminalProvider (Issue #1242), each with a deliberately-late (50ms) onData attach
  OK    cycle 0: early marker observed complete after late attach
  ... (all 20 cycles OK)
PASSED: 20/20 cycles observed the complete early marker after a deliberately-late onData attach
```
Exit code: 0

### Dev-server regression evidence (per the issue's ✅ list — regression evidence only, NOT cited as fix proof)

**Why not the shared production orchestrator.** This worktree's own session runs on the shared "agent-console" orchestrator (port 6340, ~10 unrelated active sessions across other repositories). Two independent reasons rule it out for this check: (1) that server is running the deployed (pre-fix) binary — my worktree's source changes do not apply until the server is rebuilt and restarted, so any `delegate_to_worktree` call against it tests old code and carries zero evidence either way (the exact "old binary looks the same as new binary" proxy trap this Issue's own methodology warns about); (2) restarting that shared server to pick up this build is not a delegate-level decision under any circumstances — it would disrupt ~10 other active sessions and is owner/orchestrator territory, not something to do unilaterally for a single PR's regression check.

**What was done instead.** A fully isolated single-user dev instance was started from this worktree's own fixed build (`bun run dev`, custom ports, isolated `AGENT_CONSOLE_HOME`, and a from-scratch local git clone registered as its own repository — no shared state, no shared `.git` worktree registry, no interaction with the production orchestrator or any other session). All of the following were exercised against the FIXED build via that instance's REST API (the same API `delegate_to_worktree` / the UI itself calls) and confirmed by (a) a real OS pid on the worker, (b) `ps --forest` showing the login shell's child is the actual `claude` process (not a bare idle shell — the exact symptom this Issue reports on macOS), and (c) zero `Login-shell sentinel not detected within timeout` ERROR lines anywhere in the server log across the whole session:

- **Direct create path (N=2)**: two `POST /repositories/:id/worktrees` calls (custom branch, `autoStartSession: true`) — both produced a login shell with a live `claude ...` child process within seconds.
- **Worker restart**: `POST /sessions/:id/workers/:id/restart` on the first worker — produced a fresh login shell with a fresh `claude` child.
- **Terminal worker**: `POST /sessions/:id/workers` with `type: terminal` — produced a real PTY pid.
- **UI-created session equivalent**: `POST /sessions` with `type: quick` (the same endpoint the dashboard's "new session" UI calls, exercised directly rather than through a browser) — produced a real PTY pid for the agent worker.

The instance and all its processes, worktrees, and the throwaway git clone were torn down after verification (ports freed, process tree killed).

**Elevated (multi-user, privileged-spawn) create path: skipped, by argument, not by inability.** This PR's diff touches `pty-provider.ts` (`bunTerminalProvider` / `BunTerminalPtyAdapter`) and `worker-manager.ts`'s `setupWorkerEventHandlers` watchdog — both are shared identically by the direct and elevated spawn paths. The elevated path differs from direct only in spawn-argv/chain construction (`buildElevationArgs`, the elevation helper's elevated-login-shell invocation), which this diff does not touch. Direct-path E2E above + the untouched-elevation-argv argument + the full unit suite (including the pre-existing elevated-path-adjacent tests, all green and unchanged) together constitute adequate elevated-path regression evidence without needing a disruptive multi-user instance for this PR.

**This section is regression evidence only, per the issue's explicit prohibited-evidence clause: it is NOT cited as proof that the byte-loss windows are the cause of the macOS failure.** Ubuntu direct-path `delegate_to_worktree` success carries no polarity for that question regardless of which server it runs against (see Epistemic status above).

### `bun run check:pty-early-output` real-native-path note

The pre-fix/post-fix polarity above is the mandatory Fix-plan-3 evidence. The dev-server regression pass above additionally exercised the exact production entry point (`activateAgentWorkerPty` → `spawnPty` → `bunTerminalProvider.spawn` → `setupWorkerEventHandlers`) end-to-end on Linux with the fix applied, for free.

### CodeRabbit

`coderabbit review --agent --base main` failed with an **authentication failure** in this headless worktree (`automatic_login_failed` / `authentication_failed` — browser-based OAuth cannot complete without a display), not a rate-limit. This is the known headless-CLI-auth constraint (no local CLI review available in this environment).

The GitHub-side bot DID review the first commit (1777f3d2) and posted one actionable finding (the aliasing hazard, fixed in 0799b374 — see "Post-review fix" above) plus one nitpick (fixed in 4cbd9fc2 — the `PtyDataDiagnostics` extraction). After both fix commits, `@coderabbitai review` was retriggered; the bot's commit-status for the current head (`4cbd9fc2`) reads `success | Review rate limited | ...` (checked via the 4th-surface `commits/{sha}/status` endpoint, not just `statusCheckRollup`) — it has not actually re-reviewed the fix commits despite the retrigger and a ~15+ minute wait.

**Merge disposition note (2026-07-29).** CR state: round-1 walkthrough on 1777f3d2 was genuine (1 actionable + 1 nitpick, both now fixed). Round-2 verification of the fix commits (0799b374, 4cbd9fc2) is unobtainable from either channel — local CLI headless-auth-fail (structural, not time-bound) AND GitHub bot rate-limited on retrigger (checked past the abandon-the-wait window). Per `coderabbit-ops/troubleshooting.md`'s "local CLI headless auth-fail AND GitHub bot rate-limited... both channels unavailable for the SAME commit" disposition: substituting the Architect's independent code-appropriateness audit + Orchestrator acceptance as the dual-clean gate for the two fix commits. Both the copy-on-push fix and the type extraction were independently read and confirmed line-by-line by the Orchestrator during this PR's review thread; the Architect's full audit (7-point focus list) is the next step per the Orchestrator.

### Preflight

`node .claude/skills/orchestrator/preflight-check.js` — exit 0. Test coverage check, rule/skill duplication check, language check, and source-comment blame-shift check all pass clean.

## Post-review fix: CodeRabbit caught a real aliasing hazard

CodeRabbit's GitHub-side review flagged that `PreAttachDataBuffer.push()` stored the terminal `data` callback's `Uint8Array` (or a `subarray()` VIEW of it) directly into `this.chunks`, instead of copying it. The callback's buffer is only guaranteed valid for the synchronous duration of that callback; if Bun's `Terminal` binding reuses or mutates the backing memory for a later event (a common native-binding pattern to avoid per-event allocation), a retained view would silently corrupt the buffered pre-attach bytes before `flush()` ever reads them — reintroducing, inside the new buffer itself, the exact class of silent data loss this PR exists to close. Fixed by switching both storage sites to `.slice()` (which copies) instead of `.subarray()` / a bare reference, with a comment explaining why: copying removes the need to know or verify whether Bun's Terminal binding actually reuses its callback buffer across events at all — the same don't-trust-native-behavior-assumptions principle the pre-attach buffer itself exists for (`.claude/rules/os-environment-coupling.md`), applied recursively to its own internals.

**Regression test**: pushes a mutable `Uint8Array`, mutates it in place (`fill()`) immediately after the push call but before `flush()` — simulating a later event reusing/overwriting the same backing buffer — and asserts the pre-mutation content is what `flush()` returns.

**Polarity verified by actual execution** (diff/apply `-R` on just the two `.slice()` call sites, not `git stash`, then restored — same technique as the rest of this PR's polarity checks): running this test alone against the reverted (`subarray()`/bare-reference) code gives
```
Expected: "original"
Received: "XXXXXXXX"
```
i.e. it fails exactly as the hazard predicts (the mutation reached the stored reference). Restoring the fix and re-running: 1 pass, 0 fail.

**Follow-up nitpick fixed**: the same CodeRabbit review also flagged the `{ fireCount, bufferedBytes, droppedBytes }` diagnostics shape being inlined independently in 4 places across `pty-provider.ts` and `mock-pty.ts`. Extracted into one exported `PtyDataDiagnostics` type, reused everywhere (pure refactor, no behavior change).

**Why the existing unit tests and the real-PTY smoke did not already catch this**: the unit-test harness passes a fresh `TextEncoder().encode(...)` literal on every mock `data` call — no two calls ever share a backing buffer, so aliasing/reuse is structurally unobservable in that harness, independent of what the production code does or how many such tests are added; more tests of that shape cannot ever exercise this hazard. Whether Bun's actual native `Terminal` binding reuses or mutates its callback buffer across events, and on what cadence, is not something this PR has evidence for either way — the real-PTY smoke's single-marker-per-process, immediate-flush shape may simply never land inside such a reuse window if one exists, but that is not established; the fix is written to not need an answer to that question.

## Addendum (2026-07-29, post-merge-readiness): filling a dev-verification gap for #1237 / #1241 discovered during owner follow-up

Owner asked whether #1234 had been dev-verified; it had not (PR #1237 shipped without a `delegate_to_worktree` E2E — its own smoke is a mechanism probe, not the shipping path). This worktree (`fix/1242-pty-pre-attach-buffer`, based on `origin/main`, 0 commits behind) contains #1237 + #1239 + #1241 + this PR's fix, so it is the correct base to close that gap. Recorded here rather than on #1237/#1241 directly because this is the worktree it was run from; treat this as "two already-merged items whose dev verification was missing, filled in on this occasion."

**Verification start conditions (explicit, not inferred).** Per the discipline this PR itself is built on (copy-on-push removes a native-behavior assumption instead of arguing about it), the dev server was stopped and restarted immediately before verification, not assumed fresh from HMR/watch state:
```
git status --short  ->  (clean)
git log -1           ->  4cbd9fc2 2026-07-29T15:19:58+09:00
process restart time  ->  2026-07-29T18:10:36+09:00   (> commit time, confirmed after an explicit stop+restart, not inferred from uptime)
```
Run via `bun run dev` with custom ports (`PORT=17242 CLIENT_PORT=17243`, isolated `AGENT_CONSOLE_HOME`) from this worktree directly — NOT the shared systemd orchestrator (port 6340), which was not touched. Note: an unrelated pre-existing dev instance was found listening on the default ports 3457/5173 from a **different checkout** (`source-repos/agent-console`, launched independently, unrelated to this branch) — flagging since it could otherwise be mistaken for "the" dev server for this PR; it was not used or touched.

### Verification 1 (primary): #1234 AC item 1 — 5000-character `initialPrompt` E2E via `delegate_to_worktree`'s underlying API

Built a 5000-byte prompt containing single quotes, double quotes, backticks, `$`, and newlines, submitted via the same worktree-creation endpoint `delegate_to_worktree` uses (`autoStartSession: true`).

- **Agent process actually started**: `ps --forest` showed the login shell's child was a live `claude ...` process (not idle).
- **Prompt delivered verbatim**: read `/proc/<pid>/cmdline` (NUL-separated, unambiguous — `ps`'s space-joined display cannot distinguish "one argv with spaces" from "many argvs," so it was not relied on). Result: `argc=2`, `argv[1]` length 5000, and a byte-for-byte `diff` against the source prompt file showed **zero differences**.
- **Prompt file permissions**: `-rw-------` (0600), confirmed via `ls -la`.
- **Injected line stays bounded**: the actual line typed into the PTY (captured in the worker's own output log) was `claude "$(cat '/path/to/N.prompt')"` — **114 bytes**, independent of the 5000-byte prompt size, matching `check-login-shell-sentinel.ts`'s own `<512` bytes assertion for this mechanism.

All four AC sub-items confirmed on this build.

### Verification 2: #1241 — restart redelivery

Attempted to catch a worker in the "not yet delivered" state before restarting it, to exercise the redelivery path directly. **Both attempts lost the race** (the sentinel handshake + prompt injection completed before the restart could land):

- Attempt 1: detect the new worker's PID via a DB poll loop, then fire the restart HTTP call. The gap between polling-loop-detects-PID and the actual restart-triggered kill landing server-side was measured at **~21 seconds** in the server log (an LLM-tool-call round trip is not free), while the sentinel handshake itself completed in under 5 seconds (`activity-detector` idle→active→idle transition observed at +4.4s / +6.8s from worker creation). `initial_prompt_delivered` was already `1` by the time of restart; the post-restart process correctly launched with a bare `claude` (no prompt argv) — i.e. it correctly did NOT re-deliver an already-delivered prompt.
- Attempt 2: tightened the race into a single shell invocation (poll loop + immediate `kill` on first sight of the PID, no tool-call round trip in between) to shrink the reaction window. Still lost — by iteration 228 of a sub-millisecond-interval poll, `initial_prompt_delivered` was already `1`. (The tight loop itself caused `SQLITE_BUSY` read-lock contention against the live server, which added its own noise/overhead — noted as a limitation of this specific probing technique, not a product finding.)

**Disposition**: not proven live via a genuine "redeliver after failed first delivery" race in this session — the healthy-path delivery window (single-digit seconds) is faster than an external prober's reaction time by tool-call round trip. This was not treated as a reason to fabricate success; both failed attempts are reported as found. Two mitigating facts: (a) the DB-level contract was inspected directly — `session.initial_prompt` is persisted independently of the in-memory/on-disk prompt file that `killWorker` unconditionally deletes on every kill (including restart), and gating is driven by the persisted `initial_prompt_delivered` flag, not file existence, which is the shape #1241's fix requires; (b) PR #1241's own unit test suite (unchanged, still green in this PR's full-suite run) covers the redelivery branch directly at the unit level. Per the request's own allowance ("if difficult, report as not performed with a reason" — this was attempted twice and reported, not skipped outright).

### Verification 3: #1243 — watchdog does not spuriously fire

Across all of the above (6 total worktree/session creations including the two race attempts, 1 restart): `grep -c "Login-shell sentinel not detected"` across the full server log = **0**. No spurious watchdog ERRORs in this build.

## Scope

- `bunPtyProvider` (legacy, `PTY_PROVIDER=bun-pty`) is untouched, per the issue's explicit out-of-scope list.
- No change to sentinel comparison/gating semantics, spawn sequencing, or any speculative fix (sleeps/retries/event-loop yields).
- The Issue's optional/non-gating A/B parallel experiment (unfixed-build baseline on a direct-path Ubuntu instance, N=5) was not run — timeboxed against verifying this PR's own content, and explicitly authorized to drop without penalty by both the architect and the Orchestrator. This does not change this PR's content or verification.

## AC checklist (Issue #1242 PR-1)

- [x] Pre-attach buffer implemented per Fix-plan 1
- [x] Deterministic polarity tests (7/9 flip; 2 classified as new-mechanism contract tests — see Verification)
- [x] Watchdog implemented per Fix-plan 2 with both positive and both negative tests
- [x] Real-PTY early-output smoke implemented per Fix-plan 3, run on the dev server, pre-fix and post-fix runs both in this PR body
- [x] `bun run check:pty-fd-leak` run locally and pasted above
- [x] Dev-server regression checks recorded as regression evidence only (not cited as fix proof)
- [x] Epistemic status section present verbatim
- [x] Restart path, UI-created sessions, terminal workers, and `bunPtyProvider` behavior unchanged

Refs #1242

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "check:mock-module": "bun scripts/check-mock-module-poisoners.mjs",
     "check:preview-sandbox-browser": "bun scripts/run-preview-sandbox-browser-check.mjs",
     "check:pty-fd-leak": "bun scripts/smoke/check-pty-fd-leak.ts",
+    "check:pty-early-output": "bun scripts/smoke/check-pty-early-output.ts",
     "hooks:install": "bun scripts/install-hooks.mjs",
     "preinstall": "bun scripts/check-bun-version.mjs",
     "postinstall": "bun scripts/install-hooks.mjs",
--- a/packages/server/src/__tests__/utils/mock-pty.ts
+++ b/packages/server/src/__tests__/utils/mock-pty.ts
@@ -1,5 +1,5 @@
 import { mock } from 'bun:test';
-import type { PtyProvider } from '../../lib/pty-provider.js';
+import type { PtyProvider, PtyDataDiagnostics } from '../../lib/pty-provider.js';
 
 /**
  * Disposable interface matching bun-pty's IDisposable.
@@ -33,7 +33,7 @@ export class MockPty {
    * "diagnostics absent" branch worker-manager.ts must handle gracefully
    * for `bunPtyProvider`.
    */
-  dataDiagnostics?: { fireCount: number; bufferedBytes: number; droppedBytes: number };
+  dataDiagnostics?: PtyDataDiagnostics;
   private sentinelEmitted = false;
   private autoEmitSentinel: boolean;
 
@@ -94,7 +94,7 @@ export class MockPty {
   }
 
   /** Mirrors PtyInstance's optional getDataDiagnostics() -- see `dataDiagnostics` field doc. */
-  getDataDiagnostics(): { fireCount: number; bufferedBytes: number; droppedBytes: number } | undefined {
+  getDataDiagnostics(): PtyDataDiagnostics | undefined {
     return this.dataDiagnostics;
   }
 
--- a/packages/server/src/__tests__/utils/mock-pty.ts
+++ b/packages/server/src/__tests__/utils/mock-pty.ts
@@ -26,6 +26,14 @@ export class MockPty {
   currentCols = 120;
   currentRows = 30;
   loginShellSentinel?: string;
+  /**
+   * Optional diagnostics stub for the sentinel watchdog (Issue #1242).
+   * Mirrors `BunTerminalPtyAdapter.getDataDiagnostics()`. Tests set this
+   * directly; left unset it returns `undefined`, exercising the same
+   * "diagnostics absent" branch worker-manager.ts must handle gracefully
+   * for `bunPtyProvider`.
+   */
+  dataDiagnostics?: { fireCount: number; bufferedBytes: number; droppedBytes: number };
   private sentinelEmitted = false;
   private autoEmitSentinel: boolean;
 
@@ -83,6 +91,11 @@ export class MockPty {
   /** Mirrors PtyInstance's optional dispose() so detachPty wiring is assertable. */
   dispose() {
     this.disposed = true;
+  }
+
+  /** Mirrors PtyInstance's optional getDataDiagnostics() -- see `dataDiagnostics` field doc. */
+  getDataDiagnostics(): { fireCount: number; bufferedBytes: number; droppedBytes: number } | undefined {
+    return this.dataDiagnostics;
   }
 
   // Test helpers - simulate PTY events
--- a/packages/server/src/lib/__tests__/pty-provider.test.ts
+++ b/packages/server/src/lib/__tests__/pty-provider.test.ts
@@ -342,5 +342,143 @@ describe('pty-provider', () => {
         expect(mockTerminal.close).toHaveBeenCalledTimes(1);
       });
     });
+
+    describe('pre-attach data buffer (Issue #1242)', () => {
+      it('buffers chunks that arrive before onData is attached and delivers them on first attach, in order', () => {
+        const pty = bunTerminalProvider.spawn('sh', [], {});
+        const dataCb = lastSpawn!.options.terminal!.data!;
+
+        // Fires with no adapter listener attached yet -- window 2.
+        dataCb(null, new TextEncoder().encode('hello '));
+        dataCb(null, new TextEncoder().encode('world'));
+
+        const received: string[] = [];
+        pty.onData((data) => received.push(data));
+
+        expect(received).toEqual(['hello world']);
+      });
+
+      it('preserves a UTF-8 multi-byte sequence split across the pre-attach boundary', () => {
+        const pty = bunTerminalProvider.spawn('sh', [], {});
+        const dataCb = lastSpawn!.options.terminal!.data!;
+
+        // U+1F600 "GRINNING FACE" is F0 9F 98 80, split across two
+        // pre-attach chunks.
+        const bytes = new Uint8Array([0xf0, 0x9f, 0x98, 0x80]);
+        dataCb(null, bytes.slice(0, 2));
+        dataCb(null, bytes.slice(2));
+
+        const received: string[] = [];
+        pty.onData((data) => received.push(data));
+
+        expect(received.join('')).toBe('\u{1F600}');
+      });
+
+      it('does NOT deliver anything on attach when no pre-attach bytes arrived (empty buffer is not flushed as an empty string)', () => {
+        const pty = bunTerminalProvider.spawn('sh', [], {});
+        const received: string[] = [];
+        pty.onData((data) => received.push(data));
+
+        expect(received).toEqual([]);
+      });
+
+      it('caps buffered bytes at 64 KiB, keeps the FIRST bytes, and counts the rest as dropped', () => {
+        const pty = bunTerminalProvider.spawn('sh', [], {});
+        const dataCb = lastSpawn!.options.terminal!.data!;
+
+        const CAP = 64 * 1024;
+        const overflowBy = 100;
+        const chunk = new Uint8Array(CAP + overflowBy).fill(0x61); // 'a' repeated
+        dataCb(null, chunk);
+
+        const received: string[] = [];
+        pty.onData((data) => received.push(data));
+
+        expect(received.join('').length).toBe(CAP);
+        expect(received.join('')).toBe('a'.repeat(CAP));
+        expect(pty.getDataDiagnostics!().droppedBytes).toBe(overflowBy);
+        expect(pty.getDataDiagnostics!().bufferedBytes).toBe(0); // freed after flush
+      });
+
+      it('keeps exactly the cap with zero drops when a chunk lands exactly at the boundary', () => {
+        const pty = bunTerminalProvider.spawn('sh', [], {});
+        const dataCb = lastSpawn!.options.terminal!.data!;
+
+        const CAP = 64 * 1024;
+        dataCb(null, new Uint8Array(CAP).fill(0x62)); // 'b' repeated, exactly the cap
+
+        const received: string[] = [];
+        pty.onData((data) => received.push(data));
+
+        expect(received.join('').length).toBe(CAP);
+        expect(pty.getDataDiagnostics!().droppedBytes).toBe(0);
+      });
+
+      it('does NOT replay on a second onData attach (listener replacement)', () => {
+        const pty = bunTerminalProvider.spawn('sh', [], {});
+        const dataCb = lastSpawn!.options.terminal!.data!;
+
+        dataCb(null, new TextEncoder().encode('pre-attach'));
+
+        const received1: string[] = [];
+        const disp1 = pty.onData((data) => received1.push(data));
+        expect(received1).toEqual(['pre-attach']);
+
+        disp1.dispose();
+
+        // No listener attached right now -- this chunk is counted as
+        // dropped (buffer is already flushed/retired), NOT re-buffered.
+        dataCb(null, new TextEncoder().encode('post-flush-gap'));
+
+        const received2: string[] = [];
+        pty.onData((data) => received2.push(data));
+
+        expect(received2).toEqual([]);
+        expect(pty.getDataDiagnostics!().droppedBytes).toBe('post-flush-gap'.length);
+      });
+
+      it('dispose() before any onData attach discards buffered bytes (no retention, no replay)', () => {
+        const pty = bunTerminalProvider.spawn('sh', [], {});
+        const dataCb = lastSpawn!.options.terminal!.data!;
+
+        dataCb(null, new TextEncoder().encode('will be discarded'));
+
+        pty.dispose?.();
+
+        const received: string[] = [];
+        pty.onData((data) => received.push(data));
+
+        expect(received).toEqual([]);
+      });
+
+      it('getDataDiagnostics().fireCount counts every native data callback fire, pre- and post-attach', () => {
+        const pty = bunTerminalProvider.spawn('sh', [], {});
+        const dataCb = lastSpawn!.options.terminal!.data!;
+
+        dataCb(null, new TextEncoder().encode('a'));
+        dataCb(null, new TextEncoder().encode('b'));
+        dataCb(null, new TextEncoder().encode('c'));
+
+        pty.onData(() => {});
+
+        dataCb(null, new TextEncoder().encode('d'));
+        dataCb(null, new TextEncoder().encode('e'));
+
+        expect(pty.getDataDiagnostics!().fireCount).toBe(5);
+      });
+
+      it('live chunks after the first attach still flow straight through onData (no buffering once attached)', () => {
+        const pty = bunTerminalProvider.spawn('sh', [], {});
+        const dataCb = lastSpawn!.options.terminal!.data!;
+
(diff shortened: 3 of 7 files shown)


ms2sato:
@coderabbitai review

PR #1275, opened 2026-08-03
fix: invalidate worktree queries on worktree session-created events

### LINKED ISSUE #1266: Dashboard repository card misses MCP-created worktrees until refetch (split from #1260 PR-3)
## Summary

Split from [#1260](https://github.com/ms2sato/agent-console/issues/1260) (PR-3, ruled separable — staleness, not deadlock). MCP-created worktrees appear on the Dashboard repository card only after a manual refetch:

- `worktree-creation-completed` is fired only by the REST route (`packages/server/src/routes/worktrees.ts:167`); the MCP `delegate_to_worktree` path never fires it.
- `onSessionCreated` does not invalidate the worktree queries (`packages/client/src/hooks/useSessionSideEffects.ts`), so the Dashboard repository card (which joins sessions to REST-fetched worktree rows by `locationPath`, `packages/client/src/routes/index.tsx:780-789`) misses the new row until an unrelated refetch.

With #1260's Gaps 1-3 fixed (PRs #1262 / #1265), the delegated session itself is visible and working — this residual only affects the repository card's worktree row freshness.

## Scope sketch (verify against code at implementation time)

Either fire the same completion event from the MCP path, or invalidate worktree queries in `onSessionCreated` — pick the one consistent with "server is the source of truth" and the existing side-effects layer's conventions. AC to be drafted at delegation time.

## Related

- #1260 (origin; architect ruled PR-3 separable)

Reversibility: reversible.

# Architect ruling + AC (2026-08-03 — binding)

## Ruling — invalidate on `session-created` (client-side); do NOT fire `worktree-creation-completed` from the MCP path

The Issue offers two candidates. They are not symmetric, because `worktree-creation-completed` is **two facts fused into one event**:

1. *"your task finished"* — client-task lifecycle, keyed by a **client-generated `taskId`** the browser minted before calling REST, carrying form-flow affordances (`branchNameFallback`, `setupCommandResult`, `fetchFailed`);
2. *"a worktree now exists"* — the cache-invalidation fact.

The MCP path has only (2). Firing the event from there would require synthesizing a `taskId` for a task no client ever started, and shipping form-dialog payloads to a user who is not looking at a form — propagating a browser-flow-shaped event into a path with no browser flow.

Meanwhile the authoritative signal for (2) **already reaches every client from both paths**: `session-created` broadcasts unconditionally (`websocket/routes.ts`, verified during #1260's investigation). The real defect is narrower than the Issue's framing: **the client's worktree query cache has no invalidation trigger for worktrees created outside the REST form flow** — the REST flow only happens to carry one, bundled inside a task event.

**Ruling**: react to the authoritative fact. In `useSessionSideEffects`, wrap `handleSessionCreated` so that a **worktree-type** session-created also invalidates the worktree queries for that session's repository — the same wrapper shape the file already uses for `handleSessionDeletedWithCleanup` / `handleSessionPausedWithCleanup`. Server-side stays untouched: no new event, no wire-schema change, and "server is the source of truth" is respected because the client is reacting to a server fact rather than being told what to cache.

This also fixes entry points the Issue does not name (any future non-REST worktree-session creator) for free, which the event-mirroring option would not.

## Acceptance criteria

- [ ] `useSessionSideEffects` wraps `handleSessionCreated`: on a session whose type is `worktree` (and which carries a `repositoryId`), invalidate the worktree queries for that repository via the existing `worktreeKeys` helpers — following the wrapper convention already in that file. `handleSessionCreated`'s own behavior is called through unchanged.
- [ ] **Quick sessions do not invalidate** (containment: no wasted refetch, and it pins the wrapper to the worktree case).
- [ ] **Polarity**: a `session-created` event for a worktree session triggers the worktree query invalidation — asserted at the hook level (sibling `useSessionSideEffects.test.ts`), and **failing against current `main`**, where no invalidation occurs. State the pre-change check in the PR body.
- [ ] Boundary rows: worktree session with `repositoryId` undefined → no invalidation, no throw; quick session → no invalidation; two worktree sessions created in rapid succession → both invalidate (React Query's own dedupe is the mechanism; no bespoke throttling); a worktree session created for a worktree the client already has cached (restore/resume) → invalidation still fires and is harmless (a refetch), documented as accepted rather than special-cased.
- [ ] Existing REST form flow unchanged: `worktree-creation-completed` handling, task removal, and the fallback / setup-result dialogs behave exactly as today (existing tests untouched and green). The two mechanisms may both fire for a REST-created worktree — confirm that is idempotent at the cache layer and say so in the PR body.
- [ ] **No `mock.module`** (the `mock-module-lint` gate is live) — use the established patterns per `testing.md`.
- [ ] Q10 N/A — no shared-type or wire-schema change (client-only reaction to an existing event). State it in the PR body.
- [ ] **Shipping-path verification (this is the item Q12/Q13 will be audited against)**: on an isolated dev instance, call MCP `delegate_to_worktree` and observe the Dashboard repository card showing the new worktree row **without any manual refetch or page reload**. That is the reported symptom; a hook-level test proves the invalidation call, not the user-observable outcome. Screenshot or equivalent evidence in the PR body. If it is substituted, a documented joint-skip decision must accompany it.
- [ ] Full suite + preflight green.

### Verification design (role-switch self-pass, applied at issuance)

- **(a) Production-real or proxy?** Two layers, deliberately: the hook test is a proxy for the user-visible outcome but the production code path for the invalidation contract; the dev-instance MCP delegation is production-real for the symptom itself. The second is required precisely because the first cannot see whether the card actually updates.
- **(b) Polarity?** Yes on both: the hook assertion fails on `main` (no invalidation exists), and the dev-instance check reproduces the reported staleness before the change.

ms2sato:
## Summary

MCP-created worktrees (`delegate_to_worktree`) never fired `worktree-creation-completed` (that event is only fired by the REST route), so the Dashboard repository card stayed stale until an unrelated refetch. Per the architect's ruling on this Issue, the fix reacts to the authoritative fact instead: `session-created` broadcasts unconditionally regardless of how the worktree was created, so `useSessionSideEffects` now wraps `handleSessionCreated` to also invalidate that repository's worktree query cache for `worktree`-type sessions. Server is unchanged — no new event, no wire-schema change.

- Quick sessions do not invalidate (containment).
- A worktree session's `repositoryId` is required by both the TypeScript type and the wire schema (`WorktreeSessionSchema`), so the guard's "no repositoryId" branch is unreachable over the wire; it exists purely as defense-in-depth against a bad runtime cast. Verified directly via a pure exported helper (`worktreeInvalidationKeyFor`), not just inferred.

## Pre-change (TDD polarity) check

Stashed only `packages/client/src/hooks/useSessionSideEffects.ts` (the fix), keeping the new/modified test file in place, and ran the hook test file against unmodified `main`:

```
$ bun test --preload ./src/test/setup.ts src/hooks/__tests__/useSessionSideEffects.test.ts
SyntaxError: Export named 'worktreeInvalidationKeyFor' not found in module '.../useSessionSideEffects.ts'.
0 pass / 1 fail / 1 error
```

The whole test file fails to even load without the fix (missing export used by the new tests). Restoring the fix: all 17 tests pass (`17 pass / 0 fail`). No partial polarity — every new assertion depends on behavior introduced by this change.

## Q10 (shared-type / wire-schema change)

N/A — this is a client-only reaction to an existing, unchanged wire event (`session-created`). No shared type or schema was added or modified.

## REST-flow idempotency

Existing REST form flow (`worktree-creation-completed` handling, task removal, fallback/setup-result dialogs) is untouched and its existing tests are green.

For a REST-created worktree, both `session-created` and `worktree-creation-completed` fire. Traced `useWorktreeCreationTasks.handleWorktreeCreationCompleted` (the REST path's completion handler): it only updates local sidebar task-list state (`status`, `sessionId`, `sessionTitle` on the in-memory task) — it never touched `queryClient`/`worktreeKeys` before or after this change. The worktree query-cache invalidation added here is fired exactly once, from `session-created`, regardless of which path created the worktree. There is no double-invalidation of the same cache key to reason about; the two events operate on disjoint state (local task list vs. TanStack Query cache).

**Restore/resume boundary (explicit):** a `session-created` event for a worktree the client already has cached (e.g. a restore/resume of an existing session) still fires the invalidation. This is accepted as harmless — it is just an extra `queryKey` refetch of already-fresh data — and is intentionally not special-cased to detect "already cached" and skip invalidation.

## Shipping-path verification

Per the AC, verified on an isolated dev instance (outside `~/.agent-console*`: `AGENT_CONSOLE_HOME=/tmp/agent-console-1266-verify`, ports 13457/15173, since the shared default ports were already in use by another running instance):

1. Registered a throwaway local git repo as a Repository.
2. Opened the Dashboard in Chrome DevTools MCP — repository card showed 0 worktrees.
3. Called the isolated instance's MCP endpoint directly (`POST http://localhost:13457/mcp`, `tools/call` → `delegate_to_worktree`) — the real production MCP tool-call code path, invoked via raw JSON-RPC since the pre-registered `agent-console-dev`/`agent-console` MCP clients in this environment point at different (shared) instances.
4. Re-snapshotted the Dashboard **without any manual refetch or reload** — the new worktree row (branch `chore/echo-hello`, session in the sidebar) was present.
5. Confirmed via the network log that this was invalidation-driven, not stale-time coincidence: a fresh `GET /api/repositories/<id>/worktrees` request appears immediately after the MCP call, matching `worktreeInvalidationKeyFor`'s query key.

Evidence (DOM snapshot + network trace of the post-MCP-delegate Dashboard state): see PR comments below.

Cleaned up afterward: removed the throwaway repo/session data, shut down the isolated instance, and deleted its data root — no artifacts left in the shared environment.

## CodeRabbit

Local CLI cannot authenticate in this headless worktree (structural auth-fail, not a quota condition). GitHub-side bot's commit-status description reads "Review rate limited" (rollup shows `CodeRabbit=SUCCESS`, which is not itself a verdict — see `coderabbit-ops` "the fourth surface"); `reviewDecision` is empty. Both channels are unavailable for this commit — see PR comments for the full disposition note; this needs the Architect + Orchestrator dual-clean disposition before merge per `coderabbit-ops` troubleshooting.

## Test plan

- [x] `bun run test` — full suite, second clean run: `TEST_EXIT: 0` (first run had 1 unrelated flake in `terminal-store-paging.test.ts`, unrelated to this change, confirmed non-reproducing on isolated re-runs and on a clean second full run).
- [x] `bun run typecheck` — clean.
- [x] `node .claude/skills/orchestrator/preflight-check.js` — clean (test coverage, rule/skill duplication, language check, blame-shift check all pass).
- [x] Shipping-path E2E via MCP on an isolated dev instance (see above).

Closes #1266

### DIFF
--- a/packages/client/src/hooks/__tests__/useSessionSideEffects.test.ts
+++ b/packages/client/src/hooks/__tests__/useSessionSideEffects.test.ts
@@ -9,7 +9,8 @@ import type {
 import type { UseWorktreeCreationTasksReturn } from '../useWorktreeCreationTasks';
 import type { UseWorktreeDeletionTasksReturn } from '../useWorktreeDeletionTasks';
 import type { UseSessionStopTasksReturn, SessionStopTask } from '../useSessionStopTasks';
-import { useSessionSideEffects } from '../useSessionSideEffects';
+import { useSessionSideEffects, worktreeInvalidationKeyFor } from '../useSessionSideEffects';
+import { worktreeKeys } from '../../lib/query-keys';
 import { clearDraftsForSession, _getDraftsMap } from '../useDraftMessage';
 import { _reset as resetWebSocket } from '../../lib/app-websocket';
 import { MockWebSocket, installMockWebSocket } from '../../test/mock-websocket';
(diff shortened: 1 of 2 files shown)


ms2sato:
Shipping-path verification screenshot (Dashboard repository card showing the new worktree row immediately after MCP delegate_to_worktree, no manual refetch/reload):

ms2sato:
Shipping-path verification evidence (isolated dev instance, MCP `delegate_to_worktree`, no manual refetch/reload):

**Dashboard accessibility-tree snapshot immediately after the MCP call** (repository card now shows the new worktree row, branch `chore/echo-hello`, session appears in the sidebar too):

```
uid=1_29 heading "agent-console-1266-verify-repo" level="2"
...
uid=2_1 StaticText "1"
uid=2_2 StaticText "Echo hello"
uid=2_4 StaticText "chore/echo-hello"
uid=2_6 StaticText "Idle"
uid=2_7 button "/tmp/agent-console-1266-verify/repositories/agent-console-1266-verify-repo/worktrees/wt-001-yb4v in Finder"
uid=2_8 link "Open" url="http://localhost:15173/sessions/9354f5d4-fcde-45fd-a0a5-1d4fae66e968"
```

**Network log proving invalidation-driven refetch** (not a stale-time coincidence): a fresh `GET /api/repositories/d25f0cc5.../worktrees` request (reqid=276) appears immediately after the MCP `delegate_to_worktree` call, following the initial load's `GET .../worktrees` (reqid=274) that showed 0 worktrees.

I did capture a full-page PNG screenshot locally during verification but gh CLI has no straightforward way to attach a binary image to a PR comment from this headless session; the DOM snapshot + network trace above is the text-equivalent evidence.

ms2sato:
## Note on CodeRabbit

Local CodeRabbit CLI cannot authenticate in this headless worktree (structural, not a quota condition). The GitHub-side bot's status is `CodeRabbit=SUCCESS` in the rollup, but the commit-status `description` reads **"Review rate limited"** — no real review has run for this commit on either channel. `reviewDecision` is empty (unreviewed, not approved).

Per `coderabbit-ops` troubleshooting ("Local CLI is a headless auth-fail AND the GitHub bot is rate-limited... both channels are unavailable for the SAME commit"), this needs the Architect + Orchestrator dual-clean disposition before merge — flagging to the Orchestrator rather than deciding unilaterally, since disposition/merge authority is not mine as delegate.

ms2sato:
@coderabbitai review

Issue #1273, opened 2026-08-03
embedded-agent: interrupted turn produces no failure/completion signal to delegating parent


ms2sato:
## Summary

Documented residual from Issue #1264's obligation-scoped auto-activation ruling (option d): a worker whose initial prompt **was already delivered** and that was mid-turn when the server died stays deactivated after revival, exactly as it did before #1264. Auto-activating it on revival would not help — transcript restore (#1123) rebuilds the conversation, but no turn-resumption mechanism exists, so the revived agent would simply sit idle after restore, having received no new instruction to act on.

The delegating parent (e.g. a session that used `delegate_to_worktree` and is awaiting a `send_session_message` reply, or an orchestrator polling for completion) has no way to learn that the turn it was waiting on will never complete. It waits on a callback that never arrives. This is a missing completion/failure signal, not a revival-policy problem — auto-activating this class would produce an idle process, not progress.

## Why this was not folded into #1264

Issue #1264's architect ruling deliberately scoped auto-activation to workers with an **undelivered** initial-prompt obligation, because that is the only class where eager activation on revival produces work rather than an idle process. See #1264's ruling section "Residual this ruling does NOT close (and cannot)" for the full reasoning — reproduced here:

> A worker whose prompt **was** delivered and that was mid-turn when the server died stays deactivated until opened or messaged. Auto-activating it would not help: restore rebuilds the transcript but **no turn-resumption mechanism exists**, so the revived agent sits idle exactly as it would have. The delegating parent, waiting on a callback that will never come, is the real problem there — and it is a *missing completion/failure signal*, not a revival-policy problem.

## Scope (to be drafted by the architect before implementation)

Candidate directions, not yet evaluated:
- A revived embedded-agent worker whose prompt was already delivered but has no live subprocess could, on the delegating parent's next poll / on revival itself, receive a synthetic "interrupted, restart required" event routed back to the parent session.
- Alternatively, the delegating parent's own waiting mechanism (whatever it is — a conditional wakeup, a blocking MCP call, a poll loop) could gain a timeout/liveness check independent of the embedded-agent worker's own state.
- Whether this should generalize to interrupted terminal-agent (PTY) turns too, or is embedded-agent-specific.

## Related

- #1264 (obligation-scoped auto-activation on revival; this Issue's origin, architect ruling text quoted above)
- #1123 (transcript restore across restart — the mechanism that rebuilds the conversation but does not resume the turn)

Reversibility: reversible.

PR #1276, opened 2026-08-03
fix: obligation-scoped auto-activation for embedded-agent revival (closes #1264)

### LINKED ISSUE #1264: embedded-agent: delegated headless workers stay deactivated after server restart (residual from #1260 PR-1)
## Summary

Documented residual from #1260 PR-1 ([PR #1262](https://github.com/ms2sato/agent-console/pull/1262), architect Ruling A): after a server restart, revived embedded-agent workers are deactivated until a browser opens them — including workers that were originally created by `delegate_to_worktree` and were running headless before the restart. A delegated, headless embedded-agent task therefore does not survive a server restart without human intervention (post-PR-2, a `send_session_message` to it would also revive it via activate-on-delivery, which narrows but does not close the gap).

## Why it was not bundled into PR-1

Ruling A explicitly scoped PR-1 to the delegate path ("file it, do not bundle"): headless revival is a distinct policy decision — *which* revived workers should auto-activate (all previously-active? only delegate-created? none, with a visible "paused by restart" state?) — with resource-model implications (a restart would re-spawn every previously-active embedded subprocess at once). It interacts with #1123 (transcript restore across restart) and the restart-semantics contract (activation resets the conversation), so the right shape likely depends on #1123's landing.

## Decision needed before implementation

1. Auto-activation policy on revival (all / delegate-originated only / opt-in flag on the worker / none + UI affordance).
2. Ordering/stampede control if multiple workers revive at once.
3. Interaction with #1123: reviving before restore lands means the conversation resets — is that acceptable for a headless task mid-flight?

## Acceptance criteria

To be drafted (architect) once the policy decision is made. Placeholder until then — do not delegate before the AC lands.

## Related

- #1260 / PR #1262 (the residual's origin; Ruling A documents the scope cut)
- #1123 (transcript restore across restart)

Reversibility: reversible.

# Architect ruling + AC (2026-08-03 — binding)

## Reframing: the candidates optimize the wrong axis

(a) all / (b) delegate-originated / (c) lazy-plus-wider-delivery all sort workers by **origin**. The property that actually matters is **whether the worker has an outstanding obligation nobody will trigger**. One code fact settles it: activation by itself does not start work. `maybeDeliverInitialPrompt` (`embedded-agent-worker-service.ts:640-656`) returns early unless *all* of `worker.deliverInitialPromptOnActivation`, a non-empty `session.initialPrompt`, and `!session.initialPromptDelivered` hold. An activated worker outside that set spawns a subprocess, restores its transcript, reports ready — **and then sits idle**.

So the classes divide cleanly:

| Class | What revival buys | Who can trigger it today |
|---|---|---|
| Outstanding initial prompt, never delivered | the task actually starts | **nobody** — no browser is coming, and no message will arrive to trigger activate-on-delivery |
| Prompt already delivered, was mid-turn at restart | an idle subprocess (see residual below) | a human opening the tab, or a `send_session_message` |
| Prompt already delivered, conversation complete | an idle subprocess | same |

Only the first class is genuinely stranded, and it is the only one where eager activation produces work rather than an idle process.

## Ruling — (d) obligation-scoped auto-activation

**On startup revival, auto-activate exactly those embedded-agent workers that carry an undelivered initial-prompt obligation.** Everything else keeps today's lazy activation.

- **(a) rejected**: activates the two idle classes for nothing, and makes restart cost scale with total historical sessions rather than with pending work — an unbounded subprocess spawn at the least convenient moment.
- **(b) rejected**: "delegate-originated" is a *proxy* for the real property and would need a new persisted marker to express. The real property is already modelled and already persisted — `deliverInitialPromptOnActivation` (workers table, v26) plus `session.initialPromptDelivered` (sessions table, v24), both verified round-tripping across restart in #1241. Do not add state to express something the schema already says.
- **(c) rejected as a complete answer**: activate-on-delivery requires *someone to send something*. For the stranded class, by definition nobody does.

### Interaction with #1123 (transcript restore) — why (d) is safe where (a) is not

The Issue asks whether reviving mid-flight resets the conversation. For the class (d) activates, the question does not arise: an undelivered initial prompt means the agent never received its first user message, so there is no meaningful transcript to lose. (d) auto-activates precisely the workers with nothing at risk, and leaves the workers that *do* carry conversation history to the lazy path, where restore runs with a human present. That asymmetry is the ruling's main safety argument — state it in the PR body.

### Residual this ruling does NOT close (and cannot)

A worker whose prompt **was** delivered and that was mid-turn when the server died stays deactivated until opened or messaged. Auto-activating it would not help: restore rebuilds the transcript but **no turn-resumption mechanism exists**, so the revived agent sits idle exactly as it would have. The delegating parent, waiting on a callback that will never come, is the real problem there — and it is a *missing completion/failure signal*, not a revival-policy problem. **File it as a separate follow-up Issue** ("an interrupted embedded-agent turn produces no failure signal to its delegating parent"); do not attempt it here.

## Acceptance criteria

- [ ] **Single-writer predicate.** Extract the eligibility condition currently inlined in `maybeDeliverInitialPrompt` into one exported predicate (name it for the obligation, e.g. `hasUndeliveredInitialPrompt(worker, session)`), and have BOTH `maybeDeliverInitialPrompt` and the new revival hook call it. Duplicating the three-part condition is the drift this AC exists to prevent (I-2 Single Writer for Derived Values).
- [ ] **Revival hook** on the startup revival path (`session-initialization-service.ts`; locate the exact site and name it in the PR body). For each revived embedded-agent worker satisfying the predicate, call the same idempotent entry every other activation path uses (`activateEmbeddedAgentWorker`) — no second activation entry point, per the standing Ruling A shape.
- [ ] **Bounded and isolated**: activations run sequentially (or at a small fixed concurrency — state the number and why), each wrapped so a failure logs and continues; one worker's activation failure must never abort the remaining revivals or fail server startup. Log one line per attempted activation with sessionId/workerId and outcome.
- [ ] **Polarity**: a revived worker WITH an undelivered obligation is activated and its prompt delivered, with **no worker WebSocket opened at any point in the test** — and the test fails against current `main` (where it stays deactivated). State the pre-change check in the PR body.
- [ ] **Containment (pins this ruling against option (a))**: a revived worker WITHOUT the obligation — prompt already delivered, or never eligible — stays deactivated after startup, with no subprocess spawned. This test fails if someone implements "activate everything".
- [ ] **Boundary rows**: session with no `initialPrompt`; whitespace-only prompt; `deliverInitialPromptOnActivation === false` with an undelivered prompt (not eligible — the flag is the authority); eligible worker whose activation throws (logged, startup completes, other revivals unaffected); zero eligible workers (startup path is a no-op).
- [ ] Terminal-agent revival behavior unchanged (existing tests untouched and green).
- [ ] Q10 N/A — no shared-type or wire-schema change (the predicate reads existing persisted fields). State it in the PR body.
- [ ] Full suite + preflight green.

### Verification design (role-switch self-pass, applied at issuance)

- **(a) Production-real or proxy?** The polarity test must drive the **real startup revival path** (the service method the server calls at boot), not a direct `activateEmbeddedAgentWorker` call — a direct call proves the entry works, which was never in doubt, and would not detect the hook being absent. The loop subprocess may be a fake through the existing `spawnAsUserFn` seam; that substitution is pre-existing and recorded here.
- **(b) Polarity?** Yes on the load-bearing item (fails on `main`, where revived workers stay deactivated), and the containment test carries the inverse polarity against option (a).

ms2sato:
## Summary

Implements the architect's binding ruling on #1264 (option "d", obligation-scoped auto-activation): on session revival (server-restart auto-resume AND manual pause→resume — both funnel through the same site), auto-activate exactly the embedded-agent workers that carry an **undelivered initial-prompt obligation**. Every other revived embedded-agent worker (prompt already delivered, or never eligible) keeps today's lazy activation unchanged.

- Extracted the three-part eligibility check (`worker.deliverInitialPromptOnActivation`, non-empty trimmed `session.initialPrompt`, falsy `session.initialPromptDelivered`) out of `EmbeddedAgentWorkerService.maybeDeliverInitialPrompt` into a single exported predicate, `hasUndeliveredInitialPrompt(worker, session)` (`embedded-agent-worker-service.ts`) — I-2 single writer, consumed by both the existing delivery path and the new revival hook.
- **Named revival site**: `SessionPauseResumeService.resumeSessionInternal` (`packages/server/src/services/session-pause-resume-service.ts`, the per-worker restoration loop). This is the single site funnelling both the startup auto-resume path (`SessionManager`'s private `initialize()`, driven by `SessionInitializationService.initialize()`'s `autoResumeSessionIds`) and the manual `POST /api/sessions/:id/resume` route — the architect ruling's placeholder reference to `session-initialization-service.ts` pointed at the wrong file; that service only prepares the auto-resume candidate list, it never itself re-activates a worker.
- The new `else if (worker.type === 'embedded-agent')` branch calls `hasUndeliveredInitialPrompt` and, when true, calls `this.deps.activateEmbeddedAgentWorker` (the same idempotent `EmbeddedAgentWorkerService.activate` every other activation path uses — no second entry point). The call is wrapped in a **locally-scoped** try/catch so one worker's activation failure is logged and never propagates to the outer rollback logic that aborts the whole session resume on a PTY-worker activation failure — deliberately different failure-handling shape from the branch above it, since a revived-but-idle embedded-agent worker is a much lower-stakes miss than a broken resume.
- `docs/design/embedded-agent-worker.md` and `docs/glossary.md` updated as the spec, ahead of the code change.

## Why this is safe with #1123 (Transcript Restore)

The eligible class is defined as "never received its first user message" — an undelivered initial prompt means the worker's loop has never run at all, so there is no conversation transcript at risk. #1123's mid-turn repair machinery only matters for workers that already have a transcript; this change deliberately never touches that class (a worker that was mid-turn when the server died stays on the existing lazy "next access" path, unaffected). This asymmetry is the ruling's stated safety argument, restated here per the AC.

## Residual not addressed here (follow-up filed)

The architect ruling explicitly identifies a second class this change does NOT and cannot fix: a worker whose prompt **was** delivered and was mid-turn when the server died stays deactivated after revival — auto-activating it would just produce an idle process (no turn-resumption mechanism exists), and the real gap is that its delegating parent never learns the turn will not complete. Filed as a separate follow-up per the AC's instruction, not attempted here:

**#1273 — embedded-agent: interrupted turn produces no failure/completion signal to delegating parent**

## Sibling grep (design-principles.md)

Grepped for `initialPromptDelivered` repo-wide. Found a structurally similar three-part redelivery gate for **terminal-agent** workers in `worker-lifecycle-manager.ts` (`restartAgentWorker`'s `shouldRedeliverInitialPrompt`) — operates on `InternalAgentWorker` (a different type, different persisted field, different delivery mechanism: sentinel-injected PTY command vs `sendUserMessage`), in a different context (worker restart's redelivery gate, not session revival). Bundling it is explicitly out of scope: the AC requires "Terminal-agent revival behavior unchanged (existing tests untouched and green)". Not touched.

## Test plan

- `hasUndeliveredInitialPrompt` unit tests (`embedded-agent-worker-service.test.ts`): all-three-true, each condition individually false, whitespace-only prompt.
- `session-pause-resume-service.test.ts`, new `embedded-agent worker revival (Issue #1264)` describe block: polarity (undelivered obligation → activated), containment (already-delivered / flag-false → not activated), boundary rows (prompt undefined, prompt whitespace-only), failure isolation (one worker's activation throw does not abort resume or a second eligible worker's revival, no rollback triggered), and a zero-eligible-workers regression guard (PTY activation unaffected).
- `session-manager.test.ts`: an end-to-end DI-seam test proving the real `activateEmbeddedAgentWorker` dep wiring (added in `session-manager.ts`) actually reaches `EmbeddedAgentWorkerService.activate` via `resumeSession`, not just the mocked-deps coverage above.
- Existing `maybeDeliverInitialPrompt` / initial-prompt-delivery tests untouched and green (proof the predicate extraction is behavior-preserving).
- Existing terminal-agent / PTY `resumeSession` tests untouched and green.

**Polarity check (pre-change verification, per workflow.md TDD discipline).** For both the `session-pause-resume-service.test.ts` polarity/isolation tests and the `session-manager.test.ts` wiring test, the new `else if (worker.type === 'embedded-agent') { ... }` branch body in `session-pause-resume-service.ts` was temporarily replaced with a no-op comment, confirmed the affected tests FAIL (`Expected: [...] But it was not called.`), then the exact original code was restored (`git diff --stat` showed identical insertion count before/after — no net change from the round trip) and the tests confirmed PASS. Done independently by both the implementing delegate and this PR's author as a second pass.

```
output=$(bun run test 2>&1); exitcode=$?; echo "$output" | tail -100; echo "TEST_EXIT: $exitcode"
```
Final run: server 3801 pass / 1 skip (pre-existing, unrelated memfs probe) / 0 fail; client 2107/0; shared 597/0; integration 73/0; embedded-agent 295/0; scripts/hooks/skills 522/0. **TEST_EXIT: 0.**

`bun run typecheck` — all workspace packages exit clean.

Note: one full monorepo run hit an unrelated single-millisecond-boundary timestamp flake in `pty-notification.test.ts` (a file untouched by this diff); confirmed flaky by re-running that file alone (29/29 pass) and by two subsequent full clean runs (0 fail both times).

## Other notes

- Q10 (schema-type parallel maintenance): **N/A** — no shared-type or wire-schema change. The predicate reads existing persisted fields (`deliverInitialPromptOnActivation` since migration v26, `initialPromptDelivered` since migration v24); nothing new crosses the server/client wire.
- CodeRabbit local CLI: skipped (cannot authenticate in this headless worktree per project convention); GitHub-side bot review still applies post-push.
- Not merging — merge is owner-gated per this PR's logic-change classification.

Closes #1264

## Merge disposition note (2026-08-03)

**CR state**: local CLI unavailable (headless-worktree auth failure — no interactive login possible in this delegated session, per `coderabbit-ops` troubleshooting). GitHub-side bot posted a rate-limit reply at 10:15:57Z: "Review limit reached... next review available in: 30 minutes." No genuine walkthrough for this commit from either channel. inline: 0, reviewDecision: empty.

**Fallback disposition applied** (`coderabbit-ops` skill, "Local CLI is a headless auth-fail AND the GitHub bot is rate-limited... both channels are unavailable for the SAME commit"): both channels genuinely exhausted for this commit — this is the dual-unavailable case, not the plain both-rate-limited case (local side is structural, not a countdown). Per the skill's disposition, this needs either (a) waiting out the bot's 30-minute countdown and re-checking, or (b) routing through the Architect as a substitute verification layer (independent code-appropriateness audit) for architect+Orchestrator dual-clean. Reported to the Orchestrator for the wait-vs-proceed call; not decided unilaterally here.

## Summary by CodeRabbit

* **New Features**
  * Resuming a paused session now automatically reactivates eligible embedded-agent workers with undelivered initial prompts.
  * Workers activate only when prompt delivery is enabled and the prompt contains content.
  * A failed worker activation no longer prevents the session from resuming or other eligible workers from activating.

* **Documentation**
  * Updated embedded-agent worker and Initial Prompt documentation to describe resume-time reactivation behavior.


ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai Thanks for catching this — confirmed real. Fixed in 27b6f50a:

- Tracked successfully-activated embedded-agent worker ids in a separate array (`activatedEmbeddedAgentWorkerIds`), since `workerManager.killWorker` (used for the existing `activatedWorkers` PTY-only array) silently no-ops on non-PTY worker types — pushing embedded-agent workers into that array would have been a no-op fix.
- Deactivation runs via `Promise.allSettled` (not the `Promise.all` shape in the suggested diff) so one worker's deactivation failure can't abort the rest of the rollback sequence (`deleteSession` / DB paused-state restore).
- Deactivation runs before `deleteSession` in both rollback paths — `deactivateEmbeddedAgentWorker` looks up the session in memory, so running it after would silently no-op on an already-removed session and leave the subprocess/MCP token orphaned regardless.
- Added the two tests you suggested (PTY-failure-after-embedded-success, DB-failure-after-embedded-success) plus an explicit call-order assertion locking in the deactivate-before-deleteSession ordering.

Independently reviewed by this repo's architect before landing.

Issue #1310, opened 2026-08-16
Terminal-agent exit-127 output capture: identify why the shell's 'not found' stderr never reaches the output file (I-9 instance)


ms2sato:
## Context

Split out of #1294 per Architect ruling R-b: the exit-127 legibility fix in #1294 synthesizes a diagnostic message directly into the worker's output stream/file, bypassing the PTY output pipeline entirely. That fix does NOT depend on this issue and must not be blocked by it. This issue tracks the separate, still-open question of why the *original* shell output (e.g. `sh: 1: claude: not found`) never reached the worker's output file in the first place -- the output log was observed at 0 bytes despite a shell that failed to resolve a command normally printing to stderr.

## The gap

Per #1294's investigation:

> The worker's output log is 0 bytes, yet a shell that fails to resolve a command prints `sh: 1: claude: not found` to stderr. So either:
> - nothing was written (the failure happened before the command line was ever fed to the PTY), or
> - bytes were written and the capture path did not retain them -- which #1243's pre-attach ring buffer exists specifically to prevent.

This is an instance of I-9 "No Silent Drop of Pre-Attach Events" in `.claude/skills/architectural-invariants/SKILL.md` (the same invariant #1242/#1243 fixed for the `data`/`exit` pre-attach window).

## Candidate swallow points (from #1294's delivery trace)

1. **The sentinel gate itself** (`worker-manager.ts`'s `pty.onData` handler): pre-sentinel bytes are unconditionally dropped (`if (idx === -1) { preSentinelCarry = ...; return; }`). If the shell's error output lands in the same read window as (or ahead of) the sentinel echo -- plausible given the ~12ms window between spawn and the process dying -- it is discarded here with no buffering, unlike the #1243 pre-attach buffer's bounded-keep behavior.
2. **The pending flush buffer's lifetime**: `WorkerOutputFileManager.bufferOutput` accumulates bytes in an in-memory `pendingFlushes` map with a deferred (threshold- or timer-triggered) flush to disk. If the worker/session teardown path (or a process exit, in a dogfood/production restart) races this deferred flush, buffered-but-unflushed bytes could be lost before ever reaching disk.

## Investigation task

[...]

Issue #1291, opened 2026-08-12
embedded-agent: default EMBEDDED_AGENT_BUN_PATH to process.execPath, paired with a multi-user boot-time WARN


ms2sato:
## Summary

`EMBEDDED_AGENT_BUN_PATH` currently defaults to the bare name `'bun'` (`packages/server/src/lib/server-config.ts:190`), so an unconfigured deployment resolves the embedded-agent's runtime through PATH. Change the default to `process.execPath`, so "the embedded-agent runs the same bun binary as the server" holds **by construction** rather than by PATH coincidence — paired with a boot-time WARN in multi-user mode performing **two complementary checks** (binary identity and target-user reachability), so a misconfiguration surfaces at startup instead of as an opaque activation failure later.

## Current behaviour and why it is weak

```ts
// packages/server/src/lib/server-config.ts:190
EMBEDDED_AGENT_BUN_PATH: process.env.EMBEDDED_AGENT_BUN_PATH || 'bun',
```

Consumed at `packages/server/src/services/embedded-agent-worker-service.ts:323`. The subprocess entry is resolved from the **server's own install root** (`resolveEmbeddedAgentEntryPath`, returning `source: 'bundle' | 'package' | 'fallback'`), so the code executed is always the server's shipped build artifact. Only the runtime is in question.

With the bare-name default, the binary that executes that artifact depends on PATH resolution at spawn time. It may or may not be the binary the server itself is running, and nothing reports a divergence.

The elevated case is worse than "may differ". `elevation-args.ts` lists PATH in `PRIVILEGE_BOUNDARY_PROTECTED` and strips it from injectable env, so the elevated shell's PATH comes from the **target user's** login-shell init — not the server's. A bare name therefore resolves in the target user's minimal login PATH, which is the exact defect #1221 was filed for (exit 127 inside the elevated shell, invisible to the operator).

**Observed on the live instance (2026-08-13).** The deployed unit predates #1222: `ExecStart=/home/agentconsole/.bun/bin/bun run start`, with **no `Environment=EMBEDDED_AGENT_BUN_PATH=` line at all**. Embedded-agent workers fall back to `'bun'`, which resolves to `/usr/local/bin/bun` (present from an earlier setup run, on the default PATH). Server and workers therefore execute **two different binaries** that can drift on any `bun upgrade` — and nothing reports it. It works by accident.

[...]

PR #1315, opened 2026-08-16
fix: surface exit-127 (agent binary not found) as a worker output message

### LINKED ISSUE #1294: agent CLI missing for a user: the worker dies silently — surface the root cause as an error message
## Status: mechanism PINNED (2026-08-13). Remaining scope is one sub-question plus the fix.

Step 0 (the watchdog / journal grep) produced the answer. The Issue's original framing — "investigate before designing" — was right, and it took one log grep rather than a reproduction.

## The mechanism

```
worker-manager  WARN  "Agent worker activated without session.createdBy; skipping MCP token mint"
worker-manager  INFO  workerId=addac5af… pid=3416575 exitCode=127 signal=null  "Worker exited"
```

**`exitCode: 127` — command not found.** All four dead workers show the identical shape, and each retried on restart with the same 127. Deterministic, not a race.

So the chain is:

1. `createdBy` is NULL (caller omitted `parentSessionId` — the #1293 route, or any other ownerless path).
2. `resolveSpawnUsername` falls back to the **server-process user** (`agentconsole`), exactly as documented.
3. `MultiUserMode.spawnPty` takes the elevation-skip **direct spawn** branch — also as documented.
4. The agent command is `claude --dangerously-skip-permissions --model … ` — a **bare binary name**.
5. `claude` is installed **only** at `/home/ms2sato/.local/bin/claude` (a symlink into `~/.local/share/claude/versions/…`). There is no `/usr/local/bin/claude` and no `/usr/bin/claude`.
6. `/home/ms2sato` is `drwxr-x---` and `/home/ms2sato/.local` is `drwx------` — **structurally unreachable to any other user**, including the service account.
7. A shell with the standard system PATH does not resolve `claude`. Verified directly: `env -i PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin sh -c 'command -v claude'` finds nothing.
8. → exit 127, ~12 ms after spawn.

**The PTY did spawn.** Real pids exist (`3416575`, `3428550`, `3429805`, and again on restart). The earlier reading in #1293's first draft — "activation fails, no process" — was wrong on this point; the process was created and died immediately. The documented ownerless fallback works exactly as designed. What it cannot do is run a binary the fallback user cannot see.

## This is the #1221 class, one binary over

`os-environment-coupling.md` Discipline 3 already states the rule: **elevated (or otherwise cross-user) commands must not resolve binaries by PATH-only name.** That rule was written for `bun`, and #1221 fixed `bun` specifically — `EMBEDDED_AGENT_BUN_PATH` plus a copy to a location every target user can traverse.

The **agent CLI has no equivalent**. `commandTemplate` invokes `claude` by bare name, which works for every agent worker whose spawn user is the human who installed it, and fails closed for any spawn user who did not. Nothing in the codebase notices the difference until exit 127.

Note this is not solely an ownerless-session concern: the same failure would hit **any** spawn user without their own `claude` install — which is now a live possibility, since four additional interactive users were provisioned into `agent-console-users` today (#911).

## Fix decomposition (Architect ruling, 2026-08-13) — three layers, and the agent CLI is NOT bun's twin

Before anyone reaches for the `EMBEDDED_AGENT_BUN_PATH` treatment: **`claude` carries per-user authentication, and `bun` does not.** A shared `/usr/local/bin/claude` would fix *resolution* for real interactive users (their `~/.claude` credentials stay per-user, so sharing the binary is fine) — but it **cannot** fix the ownerless case. The service user would then resolve the binary and fail one step later on credentials, **less legibly than exit 127**. Copying the bun fix wholesale would trade a clear failure for a murky one.

The fix therefore decomposes into three layers, each with an existing home:

**Layer 1 — prevent ownerless spawn. Home: #1293 (R2/R3), already specified.** Once ownerless delegation errors at the tool boundary, the service-user-cannot-see-`claude` case stops being reachable from the delegate path at all.

**Layer 2 — legibility. Home: this Issue.** Both open points belong here:

- **(a) The WARN must name the operative consequence.** Something of the shape *"worker will spawn as server-process user `<x>`; command resolution and file access use that identity; per-user CLI installs will not resolve"*, with the token-mint skip demoted to a secondary clause rather than the headline. The empirical proof that the current wording misleads is that the Orchestrator read it during this very investigation and concluded the situation was benign.
- **(b) The zero-bytes question is an I-9 instance, and it is now on the critical path** rather than a pure debuggability concern — whether the user-facing message comes from the shell's own stderr or from a synthesized one, this is the gap that currently swallows the evidence. The exit observer fired — `exitCode` is in the journal — so the *exit* path attached in time. But the shell's "not found" stderr went to the PTY and none of it reached the output file. That is exactly the shape [I-9 "No Silent Drop of Pre-Attach Events"](../../.claude/skills/architectural-invariants/SKILL.md) (from #1242 / #1243) exists for. The investigation step is to find where the output-file writer attaches relative to spawn; the fix direction is that the **file writer becomes a first-class consumer of the pre-attach ring buffer** — attach before spawn, or drain the buffer into the file at exit/dispose.

  Note carefully: **the delegate path is precisely the path where no browser ever attaches**, so a "flush on first attach" design provably never fires there. Without this fix, every future 12 ms spawn failure eats its own diagnostic text.

**Layer 3 — per-user CLI provisioning. Home: operations, not code.** A section in `docs/multi-user-setup-guide.md` stating that each interactive user needs the agent CLI installed and authenticated in their own account. The #911 users hitting 127 is an onboarding gap; Layer 2 is what makes the next occurrence self-diagnosing rather than requiring a journal grep by someone with `adm`.

Keeping these separate is what stops the fix from being mis-scoped as "bun all over again" — and stops Layer 3 from being mistaken for a fix to Layer 1.

## Scope (owner, 2026-08-16): surface the root cause as an error message

The owner's direction, and it is the deliverable for this Issue:

> The agent command is not installed for that user — output **that** as the error.

And explicitly **not `claude`-specific**: *"not just claude"*. This instance already carries Codex and opencode agent definitions alongside the two Claude Code ones, and `commandTemplate` is user-editable, so the message must name whichever command that agent actually invokes, for whichever user the worker spawned as.

**What the user sees today:** nothing. The PTY dies in ~12 ms, the output log is 0 bytes, and the session sits there looking idle. There is no error anywhere the user can reach — the only record is `exitCode: 127` in the server journal, which requires `adm` group membership to read. That is how this Issue was diagnosed, and no user can be expected to do that.

**What they should see:** a message naming the command that could not be found, the account it was looked up for, and enough for them to act — the account needs that agent CLI installed and authenticated.

### The population this serves

Note carefully that this is **not** the ownerless case. #1293 closes the route that produces ownerless sessions, and that route is how this defect was originally found. What survives #1293 is the case the owner is pointing at: a **real, properly-owned user whose account simply lacks the agent binary**. That is reachable today — four interactive users were provisioned into `agent-console-users` on 2026-08-13 (#911), and whether each has their own agent CLI installed is unknown.

So the two Issues cover disjoint populations:

| | Route | Closed by |
|---|---|---|
| Ownerless session spawns as the service user | caller omits `parentSessionId`, or a legacy/stale owner | #1293 |
| Owned session, user lacks the agent binary | normal use by an unprovisioned account | **this Issue** |

### Design questions, with the Architect

- **Detect-and-synthesize, fix-the-capture, or both.** Fixing the capture gap (below) would let the shell's own `sh: 1: claude: not found` reach the user. Synthesizing from the exit status can produce a better message — naming the command, the account, and the remedy — but needs a judgment on whether exit 127 alone is a safe signal (127 is conventionally command-not-found, but a child can exit 127 for its own reasons).
- **Where the message surfaces.** The worker output stream is where the user is looking; activation failure also has the tool-result channel already discussed in this Issue's scope list.
- **How it stays correct across agent kinds** without hardcoding a binary name, given `commandTemplate` is user-editable.

## Remaining sub-question: why zero bytes?

The worker's output log is **0 bytes**, yet a shell that fails to resolve a command prints `sh: 1: claude: not found` to stderr. So either:

- nothing was written (the failure happened before the command line was ever fed to the PTY), or
- bytes were written and the capture path did not retain them — which #1243's pre-attach ring buffer exists specifically to prevent.

These are different defects and the evidence must distinguish them. This is the one thing Step 0 did **not** settle, and it is worth settling: if it is the second, the ring buffer has a gap on a path nobody has exercised, and that gap would silently hide the diagnostic text for every future spawn failure of this shape.

## The warning that was right there and pointed elsewhere

The server logged a WARN at the exact moment of the defect:

> `Agent worker activated without session.createdBy; skipping MCP token mint (MCP calls from this worker will be rejected if AGENT_CONSOLE_MCP_AUTH=enforce is set; see Issue #1107)`

It correctly detects the ownerless condition, and then names the **MCP-token** consequence — which is conditional (`if enforce`) and, under the current `warn` default, harmless. It says nothing about the consequence that actually killed the worker: the spawn user becomes the service account, which may not be able to run the agent binary at all.

Whoever fixes this should treat the message as part of the fix. A warning that fires on the right condition and describes the wrong consequence is worse than silence — it invites the reader to conclude the situation is benign.

## Scope now (this Issue = Layer 2)

1. **Settle the zero-bytes sub-question as an I-9 instance** — locate where the output-file writer attaches relative to spawn, with evidence rather than reasoning.
2. **Correct the WARN's consequence text** per Layer 2(a).
3. **Surface the activation outcome** — a worker that exits 127 within 12 ms should not leave the tool result claiming success (this is the terminal-path asymmetry the original Issue described; where the check belongs is now answerable, since the failure escapes at the exit observer).

## Original investigation framing (superseded, kept for the trail)

The Issue opened by insisting the tempting explanation — "no parent → no owner → activation fails" — did not hold at source, and that the `created_by` NULL ↔ dead-session correlation was an anchor observation rather than a mechanism. Both held up: the fallback did spawn, and the real cause was one layer further down. The anchor pointed at the right *condition* and the wrong *mechanism*, which is exactly the failure mode #1242 established.

Related: #1293 (the caller-error route into this state), #1221 / #1291 (the same bare-name class, for `bun`), #1243 (the ring buffer the zero-bytes question tests), #911 (newly provisioned users who would hit this the same way), #1107 (cited by the misleading WARN), `os-environment-coupling.md` Discipline 3 (the rule this violates).

---

# Architect rulings (2026-08-16) — binding, do not re-open

- **R-a** — synthesize from exit 127 with **non-assertive wording** plus a remedy. Timing inference and corroboration probes are out for v1.
- **R-b** — the **output-capture / zero-bytes repair is a separate Issue**; the synthesis must not depend on it. I-9 instance. The "deferred, not dismissed" note moves there.
- **R-c** — surface on the **worker output stream** plus a structured log. Showing it in the **tool return value is a non-goal** (the 127 exit races the already-returned response). **Corrected 2026-08-16**: `buildPtyNotificationText` is reusable as the *text formatter* only. Its delivery sibling `writePtyNotification` writes via `writeInput` — input injection into a live PTY — and there is no PTY left to write into 12 ms after exit. Format convention: reuse. Delivery convention: do not. See S2.
- **R-d** — the command name comes from the **first token of the expanded `commandTemplate`**, skipping `VAR=value` prefixes; **never** from the final wrapped string (that yields `env` / `sh` — our own plumbing). Terminal agents only.

## The delivery trace behind S2 (verdict facts, from source)

- `buildPtyNotificationText` (`pty-notification.ts:139-148`) is a pure text formatter. `writePtyNotification` (:159-171) delivers via `writeInput` — unusable here.
- The `pty.onData` handler (`worker-manager.ts:860-887`) runs sentinel-gate preprocessing **first**, then advances `worker.outputOffset`, then calls `workerOutputFileManager.bufferOutput` (:877), then fans out to attached connections. `bufferOutput` (:466-478) is an in-memory pending buffer with a **deferred** flush, not an immediate disk write.
- So the original stderr was lost at one of two points: the sentinel gate swallowing pre-sentinel bytes, or the pending flush dying with the worker's teardown. Pinning which is R-b's Issue. What matters here: **the synthesized message must enter the pipeline after the sentinel gate and carry its own flush guarantee** — then neither candidate swallow can eat it. T2 is designed so that a third, unknown swallow is found by the test rather than by dogfood.

# Acceptance Criteria (Architect, 2026-08-16)

Scope: exit-127 legibility for **terminal-agent workers**. Deliverable is code (owner's correction), producing a user-visible error message.

**S1 — Trigger and content.** In the PTY exit observer, when an agent worker exits with `exitCode === 127`: synthesize a message using `buildPtyNotificationText` **as the text formatter only** (the `[tag] field=value` single-writer convention). Content must include: the command token (per S3), the spawn username, the exit code, a hedged diagnosis ("usually means '<cmd>' is not installed or not on PATH for user '<Y>'"), and the remedy (install / authenticate the agent CLI for that user, or adjust the agent's command template). No timing heuristics, no corroboration probes (R-a).

**S2 — Delivery: synthetic append with a flush guarantee. This is the load-bearing section.**

- A small internal helper (e.g. `appendSyntheticOutput(worker, text)`) that advances `worker.outputOffset`, calls `bufferOutput`, and fans out to currently-attached connections — i.e. joins the existing pipeline at the tail of :870-886, **strictly after the sentinel gate**.
- Then **force the pending flush to disk before worker teardown proceeds** — awaited, not fire-and-forget. If `WorkerOutputFileManager` has no public force-flush, adding a minimal one is in scope.
- **Prohibited**: `writePtyNotification` / `writeInput` (input injection into a dead PTY); routing through the PTY data path or the pre-attach ring buffer. The message is server-authored output, not terminal traffic.

**S3 — Command-token extraction (R-d).** First whitespace-delimited token of the **expanded** `commandTemplate` (the agent command actually attempted), skipping leading `VAR=value` assignment prefixes. On any ambiguity (empty result, or a token still containing `{{`), fall back to displaying the full command string. **Never** extract from the final wrapped shell string — the env-unset / sentinel wrappers make its first token our plumbing (`env` / `sh`) and the message would name the wrong thing. If threading the expanded template to the exit observer requires widening a surface you did not expect, **stop and consult** — do not silently widen.

**S4 — Secondary (in scope).** Reword the ownerless WARN ("skipping MCP token mint"): the spawn-user consequence becomes the headline ("worker will spawn as server-process user '<x>'; command resolution and file access use that identity"), and the mint-skip becomes the trailing clause.

**S5 — Non-goals (state explicitly in the PR body).** Tool-result surfacing (the 127 exit races the already-returned response; possible follow-up). The output-capture repair — **file the separate I-9 Issue as part of this PR's Definition of Done** and move the "deferred, not dismissed" note there. Embedded agents (#1291 track). Non-127 exit codes.

**T — Tests** (classify each per `testing.md` in the PR body):

- **T1** (new-mechanism contract) — agent worker exits 127 → message appended containing the command token, the username, and "127". Fails pre-fix.
- **T2** (**the message-nobody-sees guard**) — worker spawns and exits 127 with **zero clients ever attached** (the delegate-path shape) → assert the message bytes are present **in the output file on disk** (read the file, not any in-memory buffer), and that a subsequent history read returns them. This fails pre-fix, fails if the flush guarantee is missing, and fails if any unknown third swallow exists.
- **T3** (containment) — non-127 exit → no synthetic message.
- **T4** (unit, S3) — plain template / `VAR=x cmd` prefix / ambiguous first token → fallback.
- **T5** (invariant preservation) — healthy-exit path byte-identical; state in the PR body which wrong implementation it catches (synthetic append firing on every exit).

**V — Verification.** Full suite + typecheck, pasted per `workflow.md` (last 100 lines + real exit code, not a summary). **Shipping-path E2E on a dev instance**: register a throwaway terminal-agent definition whose command is a nonexistent binary (e.g. `definitely-not-installed-xyz`), create a session with it, let it exit, open the worker view → the message is visible to a human. Screenshot or paste it in the PR body. This is production-real end-to-end (real spawn, real 127, real render) — no proxy. Q9: the mechanism is mode-independent (exit observer), so single-user dev E2E suffices, and the message names the spawn user correctly in both modes. Q10: N/A (existing output channel, no wire change). Put `Closes #1294` in the PR **body**.

---

# Mechanism refinement (delegate finding, verified 2026-08-16) — V's E2E recipe corrected

While executing V's shipping-path E2E on a throwaway single-user dev instance (real fixed code, real UI flow: registered an agent with `commandTemplate: "definitely-not-installed-xyz {{prompt}}"`, started a Quick Session, opened the worker view), the worker's PTY never exited. Root cause, confirmed at source and empirically:

- `buildDirectSentinelShellCommand` / `buildElevatedSentinelCommand` (`sentinel-spawn-command.ts`) both end in a bare `exec $SHELL` (no `-c`). The sentinel-injected agent command is delivered by `worker.pty.write(command + '\r')` -- **typed as input into an already-alive interactive login shell**, not exec'd as the spawned process's own argv.
- Interactive shells do not exit on command-not-found. Verified directly in the dev session: typing `definitely-not-installed-xyz ''` prints `command not found` and returns to the prompt; typing `exec definitely-not-installed-abc` directly prints `bash: exec: ...: not found` and **still** returns to the prompt (matches documented bash behavior -- interactive `exec` failure does not exit the shell unless `execfail` is set or the shell is non-interactive).
- This holds on both the direct and elevated spawn paths -- they share the identical `exec $SHELL` wrapper shape.

**Correction: the typed-into-live-shell architecture means a bad `commandTemplate` cannot produce a real PTY exit 127 via the documented UI flow.** V's literal recipe ("register an agent with a nonexistent-binary commandTemplate, let it exit, screenshot the worker view") cannot be executed as written -- there is no UI-reachable path that reproduces the trigger. Architect confirmed this at source (`sentinel-spawn-command.ts:14-25`) and ruled that the fix's own logic (S1-S5, T1-T5) is unaffected -- it is triggered by `exitCode === 127` regardless of cause -- but the V section's verification recipe needed replacing. The replacement: a smoke test (`scripts/smoke/check-exit-127-diagnostic.ts`) drives the REAL production chain (real `bunPtyProvider`, real `WorkerManager`, real `WorkerOutputFileManager`, zero attached clients) with `SHELL` pointed at a nonexistent binary, which fails the wrapper's own `exec $SHELL` before the sentinel is ever echoed -- a real, milliseconds-fast, pre-sentinel 127.

**Wrapper-`$SHELL` hypothesis for the original production 127s -- plausible origin of the production 127s; NOT verified against production.** The `SHELL`-triggered 127 above reproduces the exact signature this Issue's Step-0 investigation found in the journal (exit 127, ~12ms after spawn, real pid, zero output bytes, pre-sentinel) more completely than the originally-named suspect: it explains the zero-byte capture (the wrapper's own error, if any, is swallowed by the same pre-sentinel gate #1310 investigates) AND the timing, and it does not depend on `commandTemplate` content at all -- whereas a bad `commandTemplate` cannot 127 the PTY at all (see above). This does not overturn the original diagnosis (`claude` unreachable for the fallback spawn user) as A real, separately-verified problem -- it may still be correct, or the two may compound -- but the causal chain from "commandTemplate names a missing binary" to "PTY exits 127" specifically does not exist in this architecture, and `$SHELL` resolution for the spawn user is an unverified alternative/contributing explanation worth checking against the actual production host (not done here). Filed for future reference; not required for #1294's fix, which does not depend on which explanation is correct.

Related: #1310 (the still-open zero-bytes/pre-sentinel-swallow investigation this refinement bears on).

ms2sato:
## Summary

Terminal-agent workers whose agent binary is missing for the spawn user die ~12ms after spawn with `exitCode: 127`. The user sees nothing: 0-byte output log, session looks idle, the only record is `exitCode: 127` in the server journal (requires `adm` group membership to read).

This PR synthesizes a diagnostic message — command token, spawn username, exit code, hedged diagnosis, remedy — directly into the worker's own output stream on exit 127, and rewords the ownerless-session WARN log so the spawn-user consequence is the headline instead of the MCP-token detail.

Closes #1294

## What changed (S1-S5)

- **S1/S2**: new `WorkerManager.appendSpawnFailureNotification` / `appendSyntheticOutput` — on `worker.type === 'agent' && exitCode === 127`, builds the message via `buildPtyNotificationText` (text-formatter reuse only, per the Architect's R-c correction — `writePtyNotification`/`writeInput` is not usable, there's no live PTY 12ms after exit), joins the output pipeline strictly after the sentinel gate, then **awaits** a new `WorkerOutputFileManager.forceFlush(sessionId, workerId)` before the rest of `onExit`'s teardown proceeds — the delegate path (MCP-created session, no browser ever attaches) has no other trigger that would ever flush the pending buffer.
- **S3**: new `extractCommandToken` (`packages/server/src/lib/command-token.ts`) — first non-`VAR=value` token of the **expanded** `commandTemplate` (captured via a closure param on `setupWorkerEventHandlers`, since `worker.pendingCommand` is already cleared by the time `onExit` fires), never the final env/sentinel-wrapped shell string. Falls back to the full string on ambiguity (all-assignment / unexpanded placeholder / empty).
- **S4**: reworded the ownerless-session WARN (`Agent worker activated without session.createdBy...`) so `worker will spawn as server-process user '<x>'; command resolution and file access use that identity` is the headline and the MCP-token-mint skip is the trailing clause.
- **S5 non-goals**: tool-result surfacing (the 127 exit races the already-returned response), embedded agents (#1291 track), non-127 exit codes — none of these are touched. The output-capture repair (why the shell's own `sh: 1: <cmd>: not found` never reaches disk) is filed separately as **#1310** and out of scope here — this PR's synthesis does not depend on it (it bypasses the PTY data path entirely).

**Additional, from an Architect consult during E2E verification (see "E2E note" below)**: the exit-127 structured log line now also carries `wrapperCommand` — the actual spawned wrapper command string, reconstructed via the same `buildDirectSentinelShellCommand`/`buildElevatedSentinelCommand` builders production spawns with — so a future investigator can tell a pre-sentinel wrapper failure from a post-sentinel one without re-running an investigation. This is a structured-log-only addition; the user-facing message (S1) is unchanged.

## E2E note: the AC's original recipe was impossible (my finding, the Architect's premise)

The AC's V section specified: register an agent whose `commandTemplate` names a nonexistent binary, create a session, let it exit, confirm the message is visible in the worker view. I ran exactly this on a throwaway single-user dev instance running this branch's code — the worker's PTY never exited.

Root cause: `buildDirectSentinelShellCommand`/`buildElevatedSentinelCommand` (`sentinel-spawn-command.ts`) both end in a bare `exec $SHELL`. The sentinel-injected agent command is delivered via `pty.write(command + '\r')` — **typed as input into an already-alive interactive login shell**, not exec'd as the spawned process's own argv. Interactive shells do not exit on command-not-found. Verified directly: typing a plain bogus command prints `command not found` and returns to the prompt; typing `exec <missing-binary>` explicitly prints `bash: exec: ...: not found` and **still** returns to the prompt (matches documented bash behavior — interactive `exec` failure doesn't exit the shell unless `execfail`/non-interactive). Same wrapper shape on both direct and elevated spawn paths.

So a bad `commandTemplate` cannot produce a real PTY exit 127 via the documented UI flow — the premise behind the AC's V-section recipe (mine, per the Architect's own framing when they confirmed this at source) doesn't hold. The Architect ruled the fix's logic (S1-S5, T1-T5) is unaffected — it triggers on `exitCode === 127` regardless of cause — but prescribed replacing the recipe with a smoke test driving the **real** production chain, since a real trigger does exist: pointing the spawn env's `$SHELL` at a nonexistent binary makes the wrapper's own `exec $SHELL` fail before the sentinel is ever echoed — a real, millisecond-fast, pre-sentinel 127, reproducing the exact signature this Issue was diagnosed from.

See `scripts/smoke/check-exit-127-diagnostic.ts`'s header comment for the full writeup, and #1294's issue body (appended "Mechanism refinement" section) for the correction-trail record, including an unverified hypothesis about the original production 127s' actual cause.

**Scope of what the smoke proves, precisely.** The smoke forces a **wrapper** failure by pointing the spawn env's `SHELL` at a nonexistent binary (`/nonexistent-xyz-127-smoke`) — this is a deliberately-injected, synthetic trigger for verification purposes. It proves the capture path (exit observer → `appendSyntheticOutput` → `forceFlush` → on-disk bytes, zero clients attached) works correctly against a **real spawned process** exiting with a **real 127**. It does **not** prove that a broken `$SHELL` is how the original production 127s this Issue was diagnosed from actually happened — that remains a plausible, unverified hypothesis (see #1294's issue body). The fix itself doesn't care why `exitCode === 127`; the smoke's job is only to prove the capture chain, not to re-diagnose production.

## Tests (per `testing.md` classification)

- **T1** (new-mechanism contract, must fail pre-fix) — `worker-manager.test.ts`: exit 127 → message with command token/username/exitCode appended to `worker.outputBuffer`.
- **T2** (new-mechanism contract, must fail pre-fix) — zero clients ever attached (delegate-path shape) → message present **in the output file on disk** (`fs.readFile`, not any in-memory buffer) and via `readHistoryWithOffset`.
- **T3** (invariant preservation) — non-127 exit → no synthetic message. Verified it fails against a wrong implementation (guard removed, fires on every exit).
- **T4** (new-mechanism contract) — `command-token.test.ts`: plain template / `VAR=x` prefix / multi-prefix / all-assignment fallback / placeholder fallback / empty / whitespace-only.
- **T5** (invariant preservation) — healthy exit (0) leaves `worker.outputBuffer` byte-identical. Guards against "synthetic append fires on every exit" — verified it fails against that wrong implementation.
- **New**: `scripts/smoke/check-exit-127-diagnostic.ts` — real-process E2E replacement for the AC's UI recipe (see above). Not part of `bun run test` / CI, run manually per its header (matches the convention of its siblings `check-pty-early-output.ts` / `check-login-shell-sentinel.ts`).
- Sibling coverage also added/updated: `pty-notification.test.ts` (new `internal-agent-spawn-failed` variant), `worker-output-file.test.ts` (new `forceFlush` method).

T1/T2/T4 confirmed to fail pre-fix (stubbed/commented the new logic, restored after, verified with `git diff --stat`/`git status` clean). T3/T5 confirmed to fail against a plausible wrong implementation (guard removed).

## Duplication check (workflow.md step 6)

`appendSyntheticOutput`'s body is a near-duplicate of the `onData` handler's post-sentinel-gate tail (outputOffset advance / `bufferOutput` / connection fan-out), minus the sentinel-detection and activity-detector steps (documented in its own JSDoc). Reviewed and deliberately left as separate, documented code rather than refactored into one shared call site: the `onData` handler is sentinel-watchdog-critical and heavily commented/tested, and consolidating would add diff surface to that path for a 6-line stylistic gain. Flagging per the duplication-check step rather than silently leaving it.

## Verification

`bun run typecheck` — 0 errors, all 6 packages.

`bun run test` (full suite):
```
server:          3852 pass, 1 skip, 0 fail  (10599 expect() calls)
client:          2146 pass, 0 fail
shared:          606 pass, 0 fail
integration:     75 pass, 0 fail
embedded-agent:  295 pass, 0 fail
scripts/hooks/skills: 528 pass, 0 fail
TEST_EXIT: 0
```

`node .claude/skills/orchestrator/preflight-check.js` — coverage / duplication / language / blame-shift checks all clean.

`bun scripts/smoke/check-exit-127-diagnostic.ts`:
```
==> exit-observer chain
  OK    worker.pty became null within timeout (detachPty ran; onExit fired and completed)
  OK    the real spawned process exited with code 127
  OK    exit callback reported the expected sessionId
  OK    exit callback reported the expected workerId
==> output file on disk (real fs, zero clients ever attached)
  OK    output file contains the diagnostic tag
  OK    output file contains exitCode=127
  OK    output file contains username=ms2sato
==> readHistoryWithOffset (manager read API)
  OK    readHistoryWithOffset also returns the diagnostic message

PASSED: 8 assertion(s) passed
```

CodeRabbit CLI review: ran `coderabbit review --agent --base main`. It emitted `{"type":"status","phase":"auth","status":"starting_login"}` then `{"type":"status","phase":"auth","status":"awaiting_browser_auth","authUrl":"..."}` and then hung — no rate-limit countdown, no further output — until a 280s external `timeout` killed it (exit code **124**, from the `timeout` wrapper; the CLI process itself never exited on its own). Per this repo's `coderabbit-ops` troubleshooting doc, this is the **headless-worktree auth failure** shape: `--agent` needs an interactive browser login to complete `awaiting_browser_auth`, which a delegated worktree session has no way to do. It is a structural limitation of the session, not a quota condition, and will not resolve by waiting or retrying. This is explicitly **not** a rate-limit — no rate-limit message was ever shown. **Caution for anyone automating this check**: a naive read of "exit code" alone is misleading here — the auth-hang case does not self-report failure via a distinct non-zero code the way an explicit `authentication_failed` reply would; it simply never returns without an external timeout. Read the JSON status stream (`awaiting_browser_auth` with no follow-up), not just the exit code.

Relying on the GitHub-side bot review for this PR, scheduled by the Orchestrator against the shared repository-wide review window. **At PR-open time the commit status is expected to read `Review rate limited`** — the window was last consumed by PR #1304's review (completed 09:28:10Z, PR #1304 merged 09:30Z at `82c31a85`) and has not yet reopened. The Orchestrator will retrigger with an `@coderabbitai review` comment once the window reopens (~an hour after 09:28Z). This is a known, expected, tracked state — not an unnoticed gap.

## Definition of Done extras

- Filed #1310 ("Terminal-agent exit-127 output capture: identify why the shell's 'not found' stderr never reaches the output file (I-9 instance)") for the still-open output-capture gap, per S5.
- Appended a "Mechanism refinement" section to #1294's issue body (correction-trail convention: append, don't rewrite) documenting the E2E finding above.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved diagnostics when an agent fails to start with exit code 127.
  - Failure details now include the command, account, exit code, diagnosis, and recommended action.
  - Diagnostics are saved to worker output and delivered even when no clients are connected.
  - Buffered worker output is flushed immediately when needed, preventing delayed or missing messages.

- **Tests**
  - Added coverage for command parsing, failure notifications, output flushing, and exit-code handling.

### DIFF
--- a/packages/server/src/__tests__/utils/mock-pty.ts
+++ b/packages/server/src/__tests__/utils/mock-pty.ts
@@ -18,7 +18,7 @@ export class MockPty {
   // Note: Single callback that gets replaced, matching PtyInstance interface contract
   // which specifies "Only one callback is supported. Subsequent calls will replace the previous callback."
   private dataCallback: ((data: string) => void) | null = null;
-  private exitCallback: ((event: { exitCode: number; signal?: number }) => void) | null = null;
+  private exitCallback: ((event: { exitCode: number; signal?: number }) => void | Promise<void>) | null = null;
   killed = false;
   /** Set true when dispose() is called. Mirrors PtyInstance's optional dispose(). */
   disposed = false;
@@ -136,10 +136,8 @@ export class MockPty {
     }
   }
 
-  simulateExit(exitCode: number, signal?: number) {
-    if (this.exitCallback) {
-      this.exitCallback({ exitCode, signal });
-    }
+  async simulateExit(exitCode: number, signal?: number): Promise<void> {
+    await this.exitCallback?.({ exitCode, signal });
   }
 }
 
--- a/packages/server/src/__tests__/utils/mock-pty.ts
+++ b/packages/server/src/__tests__/utils/mock-pty.ts
@@ -56,7 +56,7 @@ export class MockPty {
     };
   }
 
-  onExit(callback: (event: { exitCode: number; signal?: number }) => void): MockDisposable {
+  onExit(callback: (event: { exitCode: number; signal?: number }) => void | Promise<void>): MockDisposable {
     this.exitCallback = callback;
     return {
       dispose: () => {
--- a/packages/server/src/lib/__tests__/command-token.test.ts
+++ b/packages/server/src/lib/__tests__/command-token.test.ts
@@ -1,0 +1,32 @@
+import { describe, it, expect } from 'bun:test';
+import { extractCommandToken } from '../command-token.js';
+
+describe('extractCommandToken', () => {
+  it('extracts the program name from a plain template', () => {
+    expect(extractCommandToken('claude --dangerously-skip-permissions')).toBe('claude');
+  });
+
+  it('skips a single VAR=value assignment prefix', () => {
+    expect(extractCommandToken('FOO=bar claude --flag')).toBe('claude');
+  });
+
+  it('skips multiple assignment prefixes', () => {
+    expect(extractCommandToken('A=1 B=2 mycmd --x')).toBe('mycmd');
+  });
+
+  it('falls back to the full string when every token is an assignment', () => {
+    expect(extractCommandToken('FOO=bar BAZ=qux')).toBe('FOO=bar BAZ=qux');
+  });
+
+  it('falls back to the full string when the candidate still has an unexpanded placeholder', () => {
+    expect(extractCommandToken('{{cmd}} --flag')).toBe('{{cmd}} --flag');
+  });
+
+  it('returns an empty string for empty input', () => {
+    expect(extractCommandToken('')).toBe('');
+  });
+
+  it('falls back to the original (unmodified) input for whitespace-only input', () => {
+    expect(extractCommandToken('   ')).toBe('   ');
+  });
+});
--- a/packages/server/src/lib/__tests__/pty-notification.test.ts
+++ b/packages/server/src/lib/__tests__/pty-notification.test.ts
@@ -279,4 +279,25 @@ describe('buildPtyNotificationText', () => {
     expect(result).toContain('type=ci:failed');
     expect(result).toContain('intent=triage');
   });
+
+  it('builds an internal-agent-spawn-failed notification with command, username, and exitCode (Issue #1294)', () => {
+    const result = buildPtyNotificationText({
+      kind: 'internal-agent-spawn-failed',
+      tag: 'internal:agent-spawn-failed',
+      fields: {
+        command: 'claude',
+        username: 'testuser',
+        exitCode: '127',
+        diagnosis: "usually means 'claude' is not installed or not on PATH for user 'testuser'",
+        remedy: "install and authenticate the agent CLI for user 'testuser'",
+      },
+      intent: 'triage',
+    });
+
+    expect(result).toContain('[internal:agent-spawn-failed]');
+    expect(result).toContain('command=claude');
+    expect(result).toContain('username=testuser');
+    expect(result).toContain('exitCode=127');
+    expect(result).toContain('intent=triage');
+  });
 });
--- a/packages/server/src/lib/__tests__/pty-notification.test.ts
+++ b/packages/server/src/lib/__tests__/pty-notification.test.ts
@@ -288,8 +288,8 @@ describe('buildPtyNotificationText', () => {
         command: 'claude',
         username: 'testuser',
         exitCode: '127',
-        diagnosis: "usually means 'claude' is not installed or not on PATH for user 'testuser'",
-        remedy: "install and authenticate the agent CLI for user 'testuser'",
+        diagnosis: "usually means a required program is missing for user 'testuser': the spawn shell itself, or the agent command 'claude', is not installed or not on PATH",
+        remedy: "install and authenticate the agent CLI for user 'testuser', or adjust this agent's command template -- check the server log's wrapperCommand field for this event to see which one was actually attempted",
       },
       intent: 'triage',
     });
--- a/packages/server/src/lib/__tests__/worker-output-file.test.ts
+++ b/packages/server/src/lib/__tests__/worker-output-file.test.ts
@@ -138,6 +138,28 @@ describe('WorkerOutputFileManager', () => {
     });
   });
 
+  describe('forceFlush (Issue #1294)', () => {
+    it('flushes buffered data below the threshold to the live file immediately', async () => {
+      manager.bufferOutput('session-force-flush', 'worker-force-flush', 'below threshold', quickResolver);
+
+      const filePath = manager.getOutputFilePath('session-force-flush', 'worker-force-flush', quickResolver);
+      // Below the 256-byte threshold and before the interval fires: not yet on disk.
+      expect(vol.existsSync(filePath)).toBe(false);
+
+      await manager.forceFlush('session-force-flush', 'worker-force-flush');
+
+      expect(vol.existsSync(filePath)).toBe(true);
+      expect(vol.readFileSync(filePath, 'utf-8')).toBe('below threshold');
+    });
+
+    it('is a no-op when nothing is pending', async () => {
+      await expect(manager.forceFlush('session-force-flush-noop', 'worker-force-flush-noop')).resolves.toBeUndefined();
+
+      const filePath = manager.getOutputFilePath('session-force-flush-noop', 'worker-force-flush-noop', quickResolver);
+      expect(vol.existsSync(filePath)).toBe(false);
+    });
+  });
+
   describe('readHistoryWithOffset', () => {
     it('should read full history when no offset specified', async () => {
       // Create file with content
--- a/packages/server/src/lib/command-token.ts
+++ b/packages/server/src/lib/command-token.ts
@@ -1,0 +1,26 @@
+/**
+ * Extract the display token (the program name) from an EXPANDED agent
+ * commandTemplate -- the string `expandTemplate` returns, before any of
+ * this server's own env/sentinel wrapping (see `UserMode.spawnPty` /
+ * `sentinel-spawn-command.ts`). Using the wrapped shell string instead
+ * would name our own plumbing ("env" / "sh") rather than the agent's
+ * command.
+ *
+ * Skips leading `VAR=value` shell-assignment-style tokens (a template idiom
+ * some custom agents use, e.g. `FOO=bar mycommand args`) so the returned
+ * token names the actual program. Falls back to the full command string
+ * when extraction is ambiguous -- no non-assignment token found, or the
+ * candidate still contains a literal `{{` (an unexpanded placeholder) --
+ * so callers always have something reasonable to display rather than a
+ * misleading truncation.
+ */
+export function extractCommandToken(expandedCommand: string): string {
+  const tokens = expandedCommand.trim().split(/\s+/).filter(Boolean);
+  const isAssignment = (token: string): boolean => /^[A-Za-z_][A-Za-z0-9_]*=/.test(token);
+  const candidate = tokens.find((token) => !isAssignment(token));
+
+  if (candidate === undefined || candidate.includes('{{')) {
+    return expandedCommand;
+  }
+  return candidate;
+}
--- a/packages/server/src/lib/pty-notification.ts
+++ b/packages/server/src/lib/pty-notification.ts
@@ -120,6 +120,19 @@ export interface InternalConditionalWakeupPtyNotification extends BasePtyNotific
   intent: PtyNotificationIntent;
 }
 
+export interface InternalAgentSpawnFailedPtyNotification extends BasePtyNotificationParams {
+  kind: 'internal-agent-spawn-failed';
+  tag: 'internal:agent-spawn-failed';
+  fields: {
+    command: string;
+    username: string;
+    exitCode: string;
+    diagnosis: string;
+    remedy: string;
+  };
+  intent: PtyNotificationIntent;
+}
+
 export type WritePtyNotificationParams =
   | InboundEventPtyNotification
   | InternalMessagePtyNotification
@@ -127,7 +140,8 @@ export type WritePtyNotificationParams =
   | InternalReviewCommentPtyNotification
   | InternalReviewedPtyNotification
   | InternalProcessPtyNotification
-  | InternalConditionalWakeupPtyNotification;
+  | InternalConditionalWakeupPtyNotification
+  | InternalAgentSpawnFailedPtyNotification;
 
 /**
  * Build the structured notification text (`\n[tag] key1=val1 key2=val2
--- a/packages/server/src/lib/worker-output-file.ts
+++ b/packages/server/src/lib/worker-output-file.ts
@@ -504,6 +504,19 @@ export class WorkerOutputFileManager {
   }
 
   /**
+   * Force an immediate flush of a worker's buffered output to disk,
+   * bypassing the size/interval-triggered scheduling. Callers that must
+   * guarantee synthesized bytes are durable before proceeding (e.g. the
+   * exit-127 diagnostic append in worker-manager.ts) use this
+   * instead of relying on `bufferOutput`'s deferred flush -- a worker that
+   * already has zero attached WebSocket connections has no other trigger
+   * that would ever flush the pending buffer.
+   */
+  async forceFlush(sessionId: string, workerId: string): Promise<void> {
+    await this.flushBuffer(sessionId, workerId);
+  }
+
+  /**
    * Flush the pending buffer. MUST be called inside the serialization domain.
    * Legacy `.log.gz` files are migrated to `.log` on first write.
    */
--- a/packages/server/src/services/__tests__/worker-manager.test.ts
+++ b/packages/server/src/services/__tests__/worker-manager.test.ts
@@ -688,6 +688,9 @@ describe('WorkerManager', () => {
   });
 
   describe('exit-127 diagnostic message (Issue #1294)', () => {
+    // T4 (S3 command-token unit coverage) lives in
+    // packages/server/src/lib/__tests__/command-token.test.ts -- not here,
+    // since it's pure-function coverage with no WorkerManager dependency.
     it('T1: appends a message with the command token, username, and exit code to the worker output stream', async () => {
       const worker = createTestAgentWorker();
       await workerManager.activateAgentWorkerPty(worker, defaultAgentActivationParams);
@@ -749,6 +752,37 @@ describe('WorkerManager', () => {
 
       expect(worker.outputBuffer).toBe(before);
       expect(worker.outputBuffer).not.toContain('internal:agent-spawn-failed');
+    });
+
+    // T6 -- CodeRabbit MAJOR regression guard (PR #1315). This is also a
+    // Q14-shaped case (pre-pr-completeness.md): a new resource-acquisition
+    // step (the diagnostic append+flush) was added inside a function whose
+    // existing failure path (the rest of onExit's teardown) predates it, and
+    // the new step must not block that existing rollback.
+    it('T6: teardown still completes (detachPty runs, exit callbacks fire) when a connected client\'s onData throws during the exit-127 fan-out', async () => {
+      const worker = createTestAgentWorker();
+      await workerManager.activateAgentWorkerPty(worker, defaultAgentActivationParams);
+
+      let onExitFired = false;
+      workerManager.attachCallbacks(worker, {
+        // The concrete reachable trigger CodeRabbit named: a connection's
+        // synchronous onData callback throwing during
+        // appendSyntheticOutput's fan-out loop.
+        onData: () => {
+          throw new Error('boom: synchronous onData throws');
+        },
+        onExit: () => {
+          onExitFired = true;
+        },
+      });
+
+      const mockPty = ptyFactory.instances[ptyFactory.instances.length - 1];
+      // Must not throw/reject: the fix wraps the synthesis call in try/catch
+      // so a throwing onData cannot abort the rest of onExit's teardown.
+      await mockPty.simulateExit(127);
+
+      expect(worker.pty).toBeNull();
+      expect(onExitFired).toBe(true);
     });
   });
 
--- a/packages/server/src/services/__tests__/worker-manager.test.ts
+++ b/packages/server/src/services/__tests__/worker-manager.test.ts
@@ -36,6 +36,7 @@ import type { LookupOsUserFn } from '../os-user-lookup.js';
 import type { runAsUser, RunAsUserOpts } from '../privilege-elevation.js';
 import type { listDescendantPids, signalPids } from '../../lib/process-tree.js';
 import * as os from 'node:os';
+import * as fs from 'fs/promises';
 
 const TEST_CONFIG_DIR = '/test/config';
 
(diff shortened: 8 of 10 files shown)


ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

PR #1319, opened 2026-08-16
probe: THROWAWAY 1/2 — #1318 minus artifact E2E (do not merge)

ms2sato:
Diagnostic-only probe for PR #1318's CI-only agent-form-boundary.test.tsx failure. Removes ONLY packages/integration/src/embedded-agent-artifact-e2e.test.ts from #1318's head, nothing else. If this comes back green, the new E2E file is the trigger. Will be closed and the branch deleted once the answer is in hand — do not merge.

PR #1320, opened 2026-08-16
probe: THROWAWAY 2/2 — #1318 minus all of packages/integration (do not merge)

ms2sato:
Diagnostic-only probe for PR #1318's CI-only agent-form-boundary.test.tsx failure. Reverts ALL of packages/integration/ to origin/main on #1318's head, keeping every other change. Paired with #1319 (which removes only the new E2E file) to distinguish: green here + red on #1319 -> a different integration change is the trigger; red here too -> the cause is outside packages/integration entirely (most likely the packages/shared barrel export change). Will be closed and the branch deleted once the answer is in hand — do not merge.


ms2sato:
Void by construction (typecheck failure from reverting a directory coupled to a type the rest of #1318 changed, not a real data point). Closing as throwaway.

Issue #1316, opened 2026-08-16
Full-suite test failures can lose the failing test's identity to turbo output elision


ms2sato:
## The problem

`workflow.md`'s Verification Checklist requires that test results be **pasted, not summarized** — the rule exists because a summary once hid 61 CI failures behind a reported "2338 pass". But when a test fails inside `bun run test` at the repo root, **the paste can be missing the one thing it is there to carry: which test failed.**

The root `test` script runs each package's `bun test` through turbo, and turbo's terminal-UI reconciliation compresses long successful stretches into `[N lines elided]` in its own output. Observed on 2026-08-16 during PR #1315: a server-package failure produced

```
[34 lines elided]
3851 pass
1 skip
1 fail
Exited with code 1
```

The elision happened **in turbo's output**, not in any downstream capture, so the failing test's name, assertion, and stack were never present in the text the agent could read. The aggregate counts and the exit code were correct and honestly reported — they just could not identify the failure.

## Why this matters more than one flake

The specific failure here was an unreproducible one-off in a string-literal-only diff, and two clean re-runs plus a standalone package run followed; the PR proceeded correctly. **That is not the point.** The point is the general case:

- **A flake and a real regression are indistinguishable without the failure's identity.** Facing an elided failure, an agent's only options are to re-run (and hope it reproduces) or to proceed on a judgment call about the diff's nature. Neither is verification.
- **It silently weakens the paste-not-summary rule at the exact moment the rule matters.** A green run pastes fine. A *failing* run — the case the rule was written for — is the one that can arrive identity-free.
- **The cost lands on whoever hits it next, not on whoever caused it.** A flake that fails an unrelated branch sends that author hypothesising about their own diff, which is `workflow.md`'s Sub-pattern 1 in its most expensive form: a real failure log they cannot read.

## What to investigate

1. Confirm the mechanism and its trigger conditions — is the elision purely a TTY/terminal-UI behavior (and therefore absent in CI logs and when piped), or does it also apply to non-TTY output? If CI logs are unaffected, the gap is specific to local runs, which is still where agents do their pre-push verification.
2. Decide the fix. Candidates, not a decision: a turbo output/log-order flag that disables the UI reconciliation; a `test:verbose` script for the diagnose-a-failure path; documenting a per-package `bun test` fallback as the standard first move after any full-suite failure.
[...]

Issue #1301, opened 2026-08-16
unregisterRepository deletes only the org/repo directory — the session-data tree (outputs / messages / memos) survives


ms2sato:
## Summary

Unregistering a repository deletes **one** of the two directories that repository owns. The other — the one holding every session's outputs, inter-session messages, and memos — is never touched and survives forever.

This is a defect that exists **today**, independent of #1300. #1300 is about the derivation being wrong going forward; this Issue is about cleanup only ever having known one of the two addresses. Fixing #1300's derivation does not close this, and closing this does not depend on #1300 landing.

## Mechanism

Two derivations, one repository, two directories — the split #1300 documents:

| Directory | Derived by | Shape |
|---|---|---|
| worktrees, templates | `getOrgRepoFromPath(repo.path)` → git remote, falling back to `basename` | `repositories/<org>/<repo>/` |
| **outputs, messages, memos** | `getRepositorySlug(id)` → `repositories.get(id)?.name` | `repositories/<repo>/` |

`RepositoryManager.cleanupRepositoryData` (`packages/server/src/services/repository-manager.ts:566-603`) computes only the first:

```ts
const orgRepo = await getOrgRepoFromPath(repoPath);
const repoDir = getRepositoryDir(orgRepo);
// ...
await this.jobQueue.enqueue(JOB_TYPES.CLEANUP_REPOSITORY, { repoDir, requestUsername, extraDir });
```

`repoDir` is the `org/repo` path. The session-data path — `computeSessionDataBaseDir(configDir, 'repository', slug)` with `slug = repo.name` — is not computed here, not enqueued, and not deleted anywhere else in the unregister path.

For any repository **with a remote**, the two paths differ, so the session-data tree is orphaned. For a remote-less repository the two coincide (both fall back to `basename`) and cleanup happens to work — which is why this has never been noticed.

## Evidence on the live instance

Both shapes are present side by side:

```
repositories/StreetAcademy/www.street-academy.com/   <- worktrees, templates
repositories/www.street-academy.com/{messages,outputs}   16M
repositories/ms2sato/agent-console/                  <- worktrees, templates
repositories/agent-console/{memos,messages,outputs}  13M, 965 files
```

Unregistering `www.street-academy.com` today would remove the first line and leave 16M under the second.

And there is a completed unregister that demonstrates it, not just predicts it. `octocat/Hello-World` was registered and unregistered at some point; it is absent from the `repositories` table and has no sessions. What remains on disk:

```
[...]

PR #1328, opened 2026-08-17
fix(server): converge repository session-data slug on the git-remote-derived org/repo shape

### LINKED ISSUE #1300: repository data directory resolves to two different addresses: fix the slug derivation forward, let flat directories drain
## Summary

One repository gets **two different directory addresses**, because two independent functions compute "where this repository's data lives" and neither knows about the other:

| Consumer | Derivation | Result for `www.street-academy.com` |
|---|---|---|
| worktrees, templates | `getOrgRepoFromRemote(repoPath) \|\| basename(repoPath)` — reads the git remote | `repositories/StreetAcademy/www.street-academy.com/` |
| outputs, messages, memos | `getRepositorySlug(id)` → `repositories.get(id)?.name` — returns the stored name, never reads the remote | `repositories/www.street-academy.com/` |

Both directories exist on the live server and both are actively written. This is I-1 (I/O Addressing Symmetry) and I-2 (Single Writer for Derived Values) — the same identity resolving to two addresses because two writers compute it.

## The design intent, and where the implementation diverges from it

`docs/design/session-data-path.md` already states that a slug has **two valid shapes**, and that both must be handled everywhere:

> 1. **`org/repo`** — for repositories mirrored from a remote host with an organization / owner namespace.
> 2. **`repo`** — for **local-only repositories** that have no remote namespace.

The owner's account of the intent (2026-08-15) completes it: **flat is for a git repository just created locally, with no remote yet.** Once that repository gains a remote, the flat value continues to work as a fallback — but ideally it should be corrected to `org/repo` at that point.

That correction step **does not exist**. `getRepositorySlug` returns `repositories.get(id)?.name`, which is set at registration to `path.basename(absolutePath)` and never revisited. So every repository gets the local-only shape regardless of whether it has a remote.

**Confirmed against production data: no session has ever had an `org/repo` slug.** Every row is flat:

```
slug=www.street-academy.com   6 sessions
slug=agent-console            2 sessions
slug=conteditor / es-rag / wsheet / ...   1 each
```

So the two-shape design is real, documented, and one of its two branches has never run against real data.

## Direction (owner, 2026-08-15): fix the derivation going forward; do not migrate

Considered and rejected: hooking "the moment a remote is added". A remote can be added with `git remote add` entirely outside the application, so there is no event to hook — the owner named this and it holds.

Also considered: an explicit user-triggered migration, and a read-time fallback (`org/repo` first, flat second). Both were set aside against two constraints:

1. **The slug is persisted per session and frozen by design.** `session-data-path.md`: *"the slug is captured at creation time and never updated. Repository rename does not affect existing sessions."* Moving directories means rewriting `data_scope_slug` on existing rows, in direct conflict with that rule.
2. **Live workers hold open output file handles.** The largest output log on the live server was being appended to during this investigation. Moving files under a running worker breaks writes.

A read-time fallback carries a third cost: `computeSessionDataBaseDir` currently touches the filesystem **zero times** (verified) and the design doc names that purity as a property. A fallback requires a `stat`, giving that up.

**Chosen direction: correct the derivation so new sessions get the right shape, and let the flat directories drain naturally as their sessions are deleted.**

- `getRepositorySlug` returns `org/repo` when the repository has a remote, and the flat name when it does not — restoring the documented two-shape intent.
- **Existing sessions are untouched.** They keep their frozen slug, keep reading their existing directory, and keep working. No data movement, no file-handle hazard, no conflict with the rename-decoupling rule.
- New sessions land under `org/repo`, alongside the `worktrees/` and `templates/` that already live there — which also resolves the confusion that surfaced this Issue.

The result looks like today (both shapes present) but means something different: today the split is permanent, and after this it converges.

## The caveat that makes this non-trivial

**The `org/repo` branch has never been exercised by production data.** Changing the derivation sends real data down a path that has only ever been covered by tests. This is the "removing a gate reveals the seams it was accidentally shielding" pattern.

Slug-consuming files, from a grep of `packages/server/src` (excluding tests):

```
database/connection.ts        database/mappers.ts           database/schema.ts
jobs/handlers.ts              lib/session-data-path.ts      lib/worker-output-manifest.ts
mcp/mcp-server.ts             repositories/sqlite-session-repository.ts
services/github-issue-service.ts        services/internal-types.ts
services/persistence-service.ts         services/repository-lookup-types.ts
services/repository-manager.ts          services/session-converter-service.ts
services/session-deletion-service.ts
```

Fifteen files. The audit question for each: **does it assume the slug is a single path segment?** Candidate failure shapes — using the slug as a filename, embedding it in a job payload that is later split, comparing it against a directory basename during orphan detection. The only splitting site found so far is `session-data-path.ts:92`, which splits segments specifically to reject `.` / `..` and is already slash-aware by design.

`SLUG_PATTERN` (`^[A-Za-z0-9._-]+(\/[A-Za-z0-9._-]+)?$`) already permits both shapes, so validation is not the gap.

## Open for the AC

1. **The 15-file audit** — which consumers assume a single segment, and what each does when it first receives `org/repo`.
2. **What "has a remote" means precisely**, and what happens when the remote is unparseable or the repository has several. `getOrgRepoFromRemote` already exists and already fails over to basename; whether `getRepositorySlug` should reuse it directly (single writer for the derivation — the I-2 fix) or duplicate the logic is a design call, though reuse looks obviously right.
3. **Whether repositories registered before this change should be offered a correction at all**, or whether draining is sufficient. The owner's instinct was that a unification mechanism would be nice to have; the draining approach makes it optional rather than required, and that call can be deferred until the derivation is fixed and the drain rate is observable.
4. **Whether the two directories should be visibly reconciled anywhere in the UI** — the confusion that produced this Issue was a human looking at the filesystem and reasonably concluding one of the two was dead.

## Out of scope

- Migrating or deleting existing flat directories. They hold live data — output logs and inter-session messages — and their sessions are still reading them.
- Changing the worktrees / templates derivation. That side already produces `org/repo` and is not the divergent one.
- The persisted-slug freeze rule itself, which this direction is specifically shaped to avoid disturbing.

Related: `docs/design/session-data-path.md` (the two-shape spec and the freeze rule), architectural-invariants I-1 / I-2, #631 (the fragmentation bug that produced I-1 — same shape at the resource-locator layer).

ms2sato:
## Summary

A repository's session-data directory and its worktree/template directory were computed by two independent, never-synchronized derivations: worktree/template placement read the git remote (`org/repo`), while session-data (`dataScopeSlug`, frozen at session creation) used the repository's flat display name. Any repository with a remote silently split its data across two addresses.

This PR hoists a single writer, `deriveRepositorySlug(repoPath, fallback)` (`packages/server/src/lib/git.ts`), and converges all three derivation sites on it. Forward-only, per the prior ruling on the Issue: no migration, existing sessions keep their frozen (flat) slug and continue reading their existing directory unchanged. New sessions for remote-backed repositories now land under `org/repo`, alongside `worktrees/`/`templates/` — the split converges over time as old sessions drain rather than being fixed retroactively.

## Changes

- **S1 — single writer.** `deriveRepositorySlug(repoPath, fallback)` in `lib/git.ts`, built on the existing `getRemoteUrl`/`parseOrgRepo` chain plus the existing `isValidSlug` validator (reused, not duplicated). **Total by contract**: any missing remote, unparseable URL, or a parsed value that fails `isValidSlug` returns `fallback` — never throws. The fallback argument is always `path.basename(repoPath)`, never a display name, at every call site.
- **S2 — three sites converge:**
  1. `worktree-service.ts` `createWorktree` — the existing `getOrgRepoFromRemote(repoPath) || basename` call becomes `deriveRepositorySlug(repoPath, basename)` verbatim; the now-dead local `getOrgRepoFromRemote` helper is deleted.
  2. `repository-manager.ts` `cleanupRepositoryData`'s `repoDir` — the private `getOrgRepoFromPath` wrapper (the same formula, duplicated) is deleted; the call site now calls `deriveRepositorySlug` directly.
  3. **Session creation** (`session-manager.ts` `createSession`) — `RepositoryLookup.getRepositorySlug(id)` is now `async` (`Promise<string | undefined>`); `RepositoryManager.getRepositorySlug` resolves the repo and calls `deriveRepositorySlug(repo.path, basename)`. `app-context.ts`'s two wiring lambdas needed no code change (TS infers the new `Promise` return through the forwarding arrow function).
- **S3 — `buildSessionDataCleanupTargets`** (`session-data-cleanup-candidates.ts`). Its canonical-candidate call migrates from the old `getRepositorySlug(id)` param to an injectable `deriveSlug(repoPath, fallback)`, defaulting to the real `deriveRepositorySlug`. **The #1301/#1304 sentinel lock test (`session-data-cleanup-candidates.test.ts`) was updated for the new writer, not weakened** — the assertion is unchanged (a sentinel value returned by the injected function must land in the target set); only the stubbed function's identity moved from `getRepositorySlug(id)` to `deriveSlug(repoPath, fallback)`. Its continued existence with the same assertion shape is the structural-coupling receipt.
- **S4 — `RepositoryManager.getRepositorySlug`'s fate, decided.** After S2/S3 I re-grepped its actual callers rather than assume: **2 wiring lambdas** (`app-context.ts:349`, `:729`, both just forwarding) **+ 1 real caller** (`session-manager.ts:805`, the `RepositoryLookup` seam for session creation). Not dangling — kept, with its JSDoc rewritten (the old "currently the slug is the repository name" claim is now false; it describes the actual `deriveRepositorySlug`-backed derivation and states it's async).

## S5 — audit table (produced before the fix)

Re-ran the "slug-consuming files" grep myself rather than trust the Issue's transcribed list — got **16 files**, not the Issue's stated 15, plus one extra found via a sibling `parseOrgRepo`/`getRemoteUrl` grep (`routes/repositories.ts`, which doesn't contain the literal word "slug").

| File | Tier | Evidence |
|---|---|---|
| `lib/git.ts` | origin (S1) | New single writer `deriveRepositorySlug` |
| `services/repository-manager.ts` | repository-scoped actor (S2/S4) | `getRepositorySlug` (session-creation freeze) + `cleanupRepositoryData`'s `repoDir` |
| `services/session-manager.ts` | repository-scoped actor (S2) | `createSession` awaits `repositoryLookup.getRepositorySlug`, validates via `isValidSlug`, freezes `dataScopeSlug` |
| `services/worktree-service.ts` | repository-scoped actor (S2) | `createWorktree`'s `orgRepo` now calls `deriveRepositorySlug` verbatim |
| `services/session-data-cleanup-candidates.ts` | repository-scoped actor (S3) | `buildSessionDataCleanupTargets`'s canonical-candidate now calls injectable `deriveSlug` |
| `services/repository-lookup-types.ts` | interface (type-only) | `RepositoryLookup.getRepositorySlug` signature widened to `Promise<string \| undefined>` |
| `app-context.ts` | pass-through | Two wiring lambdas forward `repositoryManager.getRepositorySlug(id)` unchanged |
| `database/connection.ts` | manipulator, **not migrated (out of S2 scope)** | v18 backfill (`data_scope_slug = row.repositoryName`, pre-existing, always-flat) + a separate one-time `worktree-indexes.json` migration computation at a different site — not named in the AC's S2 call-site list, left untouched |
| `database/mappers.ts` | pass-through | camelCase↔snake_case boundary mapping, opaque string, no split |
| `database/schema.ts` | pass-through | Type-only column declaration |
| `jobs/handlers.ts` | pass-through | `{ scope, slug }` fed opaquely into `computeSessionDataBaseDir` |
| `lib/session-data-path.ts` | manipulator (already correct, not a gap) | `computeSessionDataBaseDir`/`isValidSlug` already split on `/` to reject `.`/`..` segments |
| `lib/worker-output-manifest.ts` | not applicable (false positive) | Comment mentions "slug-like" re: `workerId`, unrelated concept |
| `mcp/mcp-server.ts` | pass-through | `send_session_message` resolves path via `sessionManager.getPathResolverForSessionId` (opaque resolver); a separate direct `getRemoteUrl` call exposes `remoteUrl` in tool output — not slug derivation |
| `repositories/sqlite-session-repository.ts` | pass-through | `data_scope_slug: sessionRow.data_scope_slug` direct field copy |
| `services/github-issue-service.ts` | not applicable (false positive) + **sibling pattern, not consolidated** | Its own "slug" is issue-title slugification, unrelated; separately calls `getRemoteUrl`+`parseOrgRepo` directly for GitHub-host-specific context — different semantics, not migrated |
| `services/internal-types.ts` | pass-through | Type-only field declaration |
| `services/persistence-service.ts` | pass-through | Type-only field declaration |
| `services/session-converter-service.ts` | pass-through | `dataScopeSlug: session.dataScopeSlug` direct copy |
| `services/session-deletion-service.ts` | pass-through | `{ scope, slug }` pair forwarded into cleanup job payload, opaque |
| `services/session-initialization-service.ts` | pass-through | `resolveSessionScopePayload(orphan)` at startup, opaque |
| `services/session-pause-resume-service.ts` | pass-through | `dataScopeSlug: persisted.dataScopeSlug` direct copy |
| `services/worker-lifecycle-manager.ts` | pass-through | `{ scope, slug }` cleanup-payload forwarding, opaque |
| `routes/repositories.ts` (sibling pattern) | **sibling pattern, intentionally not consolidated** | `GET /:id/github` redirect calls `getRemoteUrl`+`parseOrgRepo` directly to produce three distinct 400 error messages (no remote / not GitHub / unparseable) — `deriveRepositorySlug`'s single fallback-on-any-failure semantics would collapse that distinction, so this is correctly not migrated |

**Three files deliberately left untouched**, so they don't read as misses: `database/connection.ts` (a different, pre-existing one-time migration computation, not named in S2), `services/github-issue-service.ts` and `routes/repositories.ts` (both need the three-way error distinction that `deriveRepositorySlug`'s single-fallback semantics would collapse — different semantics, not a gap).

**Client-side**: `grep -rn "dataScopeSlug\|data_scope_slug" packages/client packages/shared` → **zero hits**, confirmed independently twice (once before delegating, once by the implementing agent). `dataScopeSlug` never crosses the wire to the client, so no client-side tier call was needed.

## Q14 audit (one sentence per S2 call site, re-read each function to confirm)

- `worktree-service.ts` `createWorktree`: the `deriveRepositorySlug` call is the function's first statement, before any directory is created or any resource acquired — nothing to roll back around it.
- `repository-manager.ts` `cleanupRepositoryData`: the call happens before the `CLEANUP_REPOSITORY` job is enqueued, no prior acquisition in that function.
- `session-manager.ts` `createSession`: the call happens before `this.sessions.set(id, ...)`, i.e. before the function's own `Promise.allSettled` + `deleteSession` rollback machinery has anything to roll back.

No site acquires-then-rolls-back around the new call; S1's totality means no new rejection route was introduced at any of the three sites.

## Tests (classified per `testing.md`)

- **T1 (bug-polarity)** — `session-manager.test.ts`, "freezes dataScopeSlug as org/repo for a remote-backed repository": wires a real `RepositoryManager` (not a stub) as `SessionManager`'s `repositoryLookup`. **I independently re-demonstrated the polarity myself** (not just trusting the implementing agent): temporarily reverted `RepositoryManager.getRepositorySlug` to the pre-fix `repo.name` body and re-ran this test — it failed as expected (`Expected: "acme/remote-repo" Received: "remote-repo"`); restored, `git status` clean afterward.
- **T2 (invariant)** — same file, "freezes dataScopeSlug as the repository basename for a local-only repository": passes identically against both pre-fix and post-fix code, correctly — it doesn't exercise the fix, it locks the no-remote fallback shape.
- **T3 (invariant, locks the forward-only premise)** — "resolves an existing session's data directory from its frozen slug, never re-deriving on read": creates a session under one derived slug, changes the mocked remote to resolve a different slug, re-resolves the same session's path — still the original slug. Also independently reconfirmed to fail against the pre-fix body (first assertion needs the new derivation to hold in the first place).
- **T4 (S1 totality + sanitization polarity)** — "rejects session creation via the existing slug validator when the resolved slug is hostile": confirms the async migration introduced no new rejection route around `createSession`'s pre-existing `isValidSlug` guard. Independently reconfirmed to fail against the pre-fix body (old code never calls the mocked derivation at all, so no rejection occurs).
- **T5 (S3 sentinel lock)** — see S3 above; same assertion shape, updated stub identity only.
- Direct unit tests for `deriveRepositorySlug` itself in `lib/__tests__/git.test.ts`: SSH URL, HTTPS URL, no remote → fallback, unparseable URL → fallback, traversal-segment org/repo → fallback, invalid-char org/repo → fallback, fallback-value-passthrough (7 tests).
- Mechanical test fallout: ~20 test files with a `RepositoryLookup`/`getRepositorySlug` stub updated to `async`; `mock-git-helper.ts` gained a `deriveRepositorySlug` mock entry; targeted fixes where a test's expected value depended on the old flat-name derivation (`repository-manager.test.ts`, `worktree-service.test.ts`, `mcp-server.test.ts`, `api.test.ts`).

## Verification

- `bun run typecheck` (root, all 6 packages): **exit 0**.
- `bun run test` (`packages/server`, direct — the root aggregate run is subject to #1316's elided-identity issue): **3979 pass / 0 fail / 1 skip (pre-existing, unrelated), TEST_EXIT: 0**. I re-ran this myself rather than trust the summary.
- `node .claude/skills/orchestrator/preflight-check.js`: coverage ✅ (4 files matched `COVERAGE_PATTERNS`; `git.ts` and `repository-lookup-types.ts` are outside the mechanical gate — `lib/` isn't in `COVERAGE_PATTERNS` and `*-types.ts` is excluded by convention — sibling tests were added for `git.ts` anyway), rule/skill duplication ✅, language check ✅, blame-shift ✅. I re-ran this myself, exit 0.
- **Local CodeRabbit CLI**: attempted, failed with `automatic_login_failed`. This is the **headless-worktree auth failure**, not a rate limit — it is structural to running the CLI in a delegated/headless session and does not self-resolve on retry, so it will not be retried in this PR. Worth noting for reviewers: the CLI can exit `0` while its JSON stream reports `authentication_failed`, so an exit-code-only check would misread this as clean. The GitHub-side bot (a separate token) is unaffected and is the review layer for this PR.
- **Dev-instance E2E — V section, two parts with different polarity, stated explicitly per the AC**:
  - **(i) nested-path-on-disk. Bug-polarity — fails on today's (pre-#1300) main.** Reverted `packages/server`/`packages/shared` to the pre-fix commit, started a throwaway server, registered a real local git repo with `origin` pointed at a GitHub-shaped remote URL (no network reachability needed — reading a local remote URL doesn't require it), created a worktree session via the real `POST /api/repositories/:id/worktrees` → `autoStartSession` flow. Confirmed on disk: the worktree correctly landed under the nested `org/repo` path while session data landed under a **separate flat** directory — reproducing the exact bug. Restored, re-ran typecheck+full suite clean as the round-trip check. As reported by the implementing agent; see the pre-merge condition below for the independently-executed re-run of the forward direction on the fix build.
  - **(ii) full lifecycle (session delete → repo unregister → both dirs gone). Regression guard as of `82c31a85`/#1304 — already passing on unmodified main before this PR.** Stated explicitly rather than presented as new bug-polarity evidence, per the AC's note that a polarity claim which quietly stopped being true is this sprint's twice-hit failure shape. Covered at the unit level by the pre-existing `repository-manager.test.ts` "session-data cleanup (Issue #1301)" describe block, which passed throughout this change (including its "coincide (local-only repo)" test, updated for the new mock shape).

## Pre-merge condition (Architect, satisfied) — forward-direction E2E re-run on the fix build

The Architect's verdict was CLEAN with one condition: the forward-direction on-disk evidence and the full unregister lifecycle had only been reported by the implementing agent — a two-hop chain (specialist ran it → I believed the report → the Architect didn't re-run it either). I re-executed both on the fix build myself, against a disposable server instance, and recorded raw receipts below rather than a summary.

**Setup**: `AGENT_CONSOLE_HOME=/tmp/agent-console-1300-verify PORT=17301 bun src/index.ts` (fix build, commit `941d424a`) against a real local git repo (`/tmp/1300-verify-remote-repo`, one commit, `origin` = `git@github.com:verify-org/verify-repo.git`).

1. **Register the repository** (`POST /api/repositories`):
   ```json
   {"id":"bdc662ad-...","name":"1300-verify-remote-repo","path":"/tmp/1300-verify-remote-repo","remoteUrl":"git@github.com:verify-org/verify-repo.git"}
   ```
   Note `name` (the flat display name) is `1300-verify-remote-repo` — the pre-#1300 slug shape — confirming the assertion below isn't vacuous.

2. **Create a worktree with `autoStartSession: true`** (`POST /api/repositories/:id/worktrees`, mode `custom`, branch `feature-verify`). Server log:
   ```
   Worktree creation completed  worktreePath=/tmp/agent-console-1300-verify/repositories/verify-org/verify-repo/worktrees/wt-001-sics branch=feature-verify
   Initialized worker output file + manifest  filePath=/tmp/agent-console-1300-verify/repositories/verify-org/verify-repo/outputs/7601ffb3-.../e3c47199-....log
   ```
   Worktree and session output both land under the **same nested `repositories/verify-org/verify-repo/`** tree — the fix's core claim, on disk.

3. **Direct DB read** (`bun:sqlite` against `data.db`, not the API):
   ```
   {"id":"7601ffb3-...","data_scope":"repository","data_scope_slug":"verify-org/verify-repo"}
   ```
   The frozen `dataScopeSlug` is `org/repo`, not the flat `1300-verify-remote-repo` name — direct confirmation of T1's claim from a live, non-test code path.

4. **Directory listing before cleanup**:
   ```
   /tmp/agent-console-1300-verify/repositories/verify-org/verify-repo/outputs/7601ffb3-...
   /tmp/agent-console-1300-verify/repositories/verify-org/verify-repo/worktrees/wt-001-sics
   ```
   No `repositories/1300-verify-remote-repo/` flat directory exists at any point — the fix prevents it from ever being created for a new session, not just from being read incorrectly.

5. **Full unregister lifecycle** — `DELETE /api/sessions/:id` (`{"success":true}`), then `DELETE /api/repositories/:id` (`{"success":true}`). Server log:
   ```
   Repository data cleanup completed
   cleanup:repository removed empty org-parent directory
   cleanup:repository session-data directory does not exist, skipping
   ```
   `find /tmp/agent-console-1300-verify/repositories -maxdepth 5` after unregister → **empty** (only the bare `repositories/` directory remains). Both the nested `org/repo` tree and the (never-created) flat directory are gone.

Instance stopped and both throwaway trees removed afterward. No stray state left in `/tmp`.

## CodeRabbit review findings addressed (commit `6bfe20a9`)

The GitHub-side bot's formal review (`Review completed`, one submitted review, no `Outside diff range` block) found two **Major** issues, both genuinely valid, neither caught by my own verification, the Architect's full-file read, or the Orchestrator's independent check:

1. **`lib/git.ts` — raw remote URL logged on the unparseable-URL branch.** `deriveRepositorySlug`'s WARN log on a URL that failed to parse as `org/repo` included the raw `remoteUrl` in its context. A git remote URL can carry embedded HTTP(S) credentials (`https://user:token@host/...`), and the branch that logs is specifically the one seeing URLs that deviate from the ordinary SSH/HTTPS shapes — exactly the shape a credential-bearing URL would take. The sibling `isValidSlug`-failure log two lines below already had the safe shape (it logs only the already-regex-extracted `orgRepo`, never the raw URL) — that was the template. **Fix**: added `describeRemoteUrlShapeForLogging`, which reduces a remote URL to `protocol//hostname` via the WHATWG `URL` parser (userinfo-free by construction) when it parses as a URL, or a fixed marker otherwise (covers scp-style `git@host:path`, which has no credential-leak risk either way, and any other malformed shape). The diagnostic signal is kept — the payload is not. 4 new tests directly assert credentials never appear in the returned string.

2. **`repository-manager.ts` — the org/repo slug derived twice.** `cleanupRepositoryData` computed `orgRepo` once (for `repoDir`) and `buildSessionDataCleanupTargets` derived the same value again independently (for `sessionDataDirs`), via its own default `deriveSlug`. Two separate reads of the same value, with a window for them to disagree if git state changed between the two calls. **This is the exact two-writer defect Issue #1300 exists to fix, found surviving inside the fix itself** — worth stating plainly for a future reader, since a de-duplication PR re-introducing a duplicate at a call site nobody was specifically looking at is an instructive shape. Neither call site was wrong in isolation, and the `deriveSlug` injection point made the second derivation look like the *intended* use of that seam rather than a duplicate of the first — which is why it read as correct under review-in-isolation and only shows up when both call sites are held in view at once. **Fix**: thread the already-captured `orgRepo` through `buildSessionDataCleanupTargets`'s existing `deriveSlug` injection point instead of letting it re-derive. One derivation, one value, both cleanup targets guaranteed to agree. New test asserts the underlying helper is called exactly once per unregister (previously twice), with the exact `(repoPath, fallback)` arguments — the property the S3 rider's "driven by a call, not a re-derivation" framing was already meant to hold, now assertable at this call site rather than only at the sentinel-locked one.

Both fixes: full suite (3984 pass / 0 fail), typecheck, and preflight re-run clean (exit 0 each) after applying.

**Optional nano, applied**: added a one-line `{@link}` cross-reference between `getOrgRepoFromPath` and `deriveRepositorySlug` in `lib/git.ts`'s JSDoc, naming the contract difference (nullable-identification vs. total-addressing) so a future reviewer doesn't file "consolidate these" — mirrors the `resolveSpawnUsername`/`resolveRequestUsername` pattern the Architect pointed at. Commit `941d424a`.

## Design doc / glossary

- `docs/design/session-data-path.md` §5: `RepositoryLookup.getRepositorySlug` signature updated to `Promise<string | undefined>` with a note on why (git-remote-derived, makes both documented slug shapes actually reachable — previously only the flat shape was ever produced); §6 step 1 updated to `await`.
- `docs/glossary.md` "Session Data Address" entry rewritten to describe the converged (post-#1300) derivation instead of the pre-#1300 split, including that existing sessions keep their frozen flat slug until drained.

## Non-goals (per the AC)

Migration of existing sessions (forward-only, settled on the Issue); the org/flat namespace-commingling residual; expiry of drained flat directories; any change to `computeSessionDataBaseDir` semantics.

Closes #1300

## Merge disposition (2026-08-17) — dual-clean over the fix commit

**What CodeRabbit did and did not review.** The genuine walkthrough covered `941d424a` and produced the two Majors addressed above. The fix commit `6bfe20a9` is **unreviewed by the bot**: a retrigger at 01:49:51Z returned `Review rate limited`.

**This rate limit came with no countdown**, unlike the earlier one on this PR ("next review available in: 6 minutes"). The walkthrough explains why, and it is worth recording as new information about the account's behaviour: Pro/Pro+ uses **adaptive** limits — sustained high-volume activity (95th percentile among CodeRabbit users) releases additional reviews only gradually as earlier ones age out. So this is not a wait-it-out window with a knowable end.

⚠ **A misread that nearly happened, recorded because the surface is genuinely deceptive.** The walkthrough comment's scope line reads *"Reviewing files that changed ... between `941d424a` and `6bfe20a9`"* — exactly the range we wanted reviewed. **That line states intent, not completion.** The body immediately beneath it is the rate-limit explanation, and its `updated_at` matches the commit status's `Review rate limited` **to the second** — one event, two faces. Reading the scope line alone would have supported merging in the belief that the fix commit had been reviewed. This is a distinct trap from the stale-head variant (a real review pinned to an earlier commit); both are ways this surface can read as clean while nothing covered the current head.

**Disposition applied — dual-clean**, per the `coderabbit-ops` substitute path:

1. **Architect independent audit of `6bfe20a9`** — read in full, not extended over an unread diff. Their stated principle: *a verdict extended over an unread diff is the two-hop-report shape we refused on the E2E.* Verdict: **CLEAN covers `6bfe20a9`**.
2. **Orchestrator acceptance** — both fixes read in the code before this disposition was written, not accepted from the delegate's report.

**Why the shape of these two fixes made the substitute reasonable rather than merely expedient**: both are *narrowings* carrying their own polarity. Fix 1 replaces a value that could contain a credential with one that is **userinfo-free by construction** (WHATWG `URL` accessors, not ad-hoc redaction), asserted by a test checking the credential substring is literally absent. Fix 2 **removes** a derivation rather than adding behaviour, pinned by an exactly-once-with-exact-args assertion. Neither adds a surface a review would need to explore.

**Audit-record note.** Both Majors were **misses of the Architect's own `941d424a` audit** — and the first was a defect of their AC's direction, which specified logging the URL on that branch. CodeRabbit's observation that the parse-**failure** branch is precisely where credentialed URLs live is an adversarial reading that the anchored pass did not perform. Recorded because it is the empirical case for asking the fix-commit-review question every time rather than treating it as ceremony.


ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

Issue #1339, opened 2026-08-17
Hand-maintained doUpdateSet column lists silently freeze fields that mutate after insert


ms2sato:
## The defect class

`packages/server/src/repositories/sqlite-session-repository.ts`'s `save()` performs two upserts — one for `sessions`, one for `workers` — and each spells out, **by hand**, the list of columns to write on conflict:

```ts
.onConflict((oc) =>
  oc.column('id').doUpdateSet({
    session_id: workerRow.session_id,
    type: workerRow.type,
    name: workerRow.name,
    updated_at: workerRow.updated_at,
    pid: workerRow.pid,
    agent_id: workerRow.agent_id,
    base_commit: workerRow.base_commit,
  })
)
```

The INSERT side (`.values(workerRow)`) carries the full row, derived from `toWorkerRow`. The UPDATE side is a second, independent derivation of "which columns a worker row has" — maintained by hand, in a different place from the mapper.

**A column missing from the UPDATE list fails silently and asymmetrically**: the first INSERT writes the correct value, and every subsequent write skips that column, pinning it forever at its first-insert value. Nothing fails — no type error, no runtime error, no failing test. The row keeps updating (`updated_at` and `pid` change), so the record looks alive while one field is frozen.

## How it surfaced

Found during #1333 (SDK Engine Phase 1) real-browser QA. `workers.sdk_session_id` was written by the engine and appended to the persisted NDJSON log, but read back `null` from the database.

Diagnosis was confirmed with a positive control rather than a timing theory: on the same row, `pid` held a real PID and `updated_at` was fresh — proving the UPDATE branch had genuinely executed several times and does correctly write *some* fields, while `sdk_session_id` stayed null across those same executions.

`sdk_session_id` is the first field in the worker-row family **designed to mutate after creation**, so it is the first to expose the gap. #1333 adds its own entry to the list, which fixes that field.

## Why this needs a structural fix rather than one more entry

Two other worker columns are also absent from the same list: `embedded_agent_id` and `deliver_initial_prompt_on_activation`. Both are currently harmless — neither is mutated after worker creation — so this Issue is not a live bug report for them.

The point is the mechanism. The next field that mutates post-creation breaks the same way, silently, and is found the same way: by someone noticing a value is wrong during manual QA. The `sessions` upsert in the same file has the same hand-maintained shape (it happens to be more complete today, which is not a property anything enforces).

[...]

Issue #1338, opened 2026-08-17
chore(deps): bump @anthropic-ai/claude-agent-sdk off the cool-off-aged pin (0.3.226) to a current release


ms2sato:
Phase 1 of the SDK engine (#1333) pins `@anthropic-ai/claude-agent-sdk` at **`0.3.226`**. That pin is a deliberate, temporary choice, not the version the design work was done against — this Issue records why, and what a bump has to re-verify.

## Why `0.3.226` and not the current release

The version Phase 1 was developed and probed against is **`0.3.233`** (published 2026-08-14T18:52Z). At implementation time it was 3 days old, inside this repo's 7-day supply-chain cool-off gate (`bunfig.toml`, `minimumReleaseAge = 604800`).

Three options were costed and routed to the owner. The owner chose to pin the newest version that clears the gate, so that **no policy exception is created**:

- `0.3.226` — published 2026-08-08T01:48Z. The newest release that cleared the 7-day gate on 2026-08-17 (verified against npm publish timestamps; `0.3.227` was 6.5 days old that day).
- Rejected: an owner-approved bypass. `bunfig.toml`'s own exemplar for a legitimate bypass is CVE-class urgency, not schedule pressure — and this package is embedded in a subprocess that runs with the user's own Claude credentials, which is precisely the case a supply-chain cool-off exists for.
- Rejected: holding the PR ~4 days. Phases 2 and 3 are serial behind Phase 1, so the delay would have stalled the whole track.

`0.3.233` clears the gate at **2026-08-21T18:52Z**. Any release at least 7 days old at bump time is equally eligible.

## Two independent version numbers — do not conflate them

- **`2.1.233`** is the bundled `claude` CLI binary's own version.
- **`0.3.233`** is the `@anthropic-ai/claude-agent-sdk` npm package version.

They are separate artifacts that happen to share a trailing patch number. The design spec's original text said "SDK 2.1.233" throughout; #1333 lands a correction-trail amendment fixing this. State both explicitly whenever a bump cites "the SDK version".

## What a bump must re-verify

The spec's empirical premises are **version-pinned by design** — they record measured behavior, not documented behavior. A bump is therefore not a lockfile-only change:

- Re-probe the event-mapping premises in `docs/design/embedded-agent-sdk-engine.md` Appendix A against the new version, and update the table with a correction trail if anything moved.
[...]

Issue #1045, opened 2026-07-12
embedded-agent: FF-2 — sandbox and MCP per-caller tool filtering (defense-in-depth)


ms2sato:
Fast-follow for umbrella #1004 (Embedded Agent Worker v1). Separate from the tool-implementation issues (FF-1a/b/c) because the scope crosses builtin tools + existing MCP surface + system-wide policy floor + OS sandbox. Architect-designed (session `84a3a530-86a8-4825-b103-bc3edfe764ee`).

## Depends on

- **FF-1a (#1042)** — `enabledTools` field exists, path confinement helper exists
- **FF-1b (#1043)** — Bash exists as the primary "high-risk tool" target for OS sandbox
- Should land AFTER FF-1a/b/c so the defense-in-depth layer covers the actual tool implementations. Can be scheduled in a separate sprint.

## Context and important prior fact

During Q2 investigation, architect (session `84a3a530`) discovered:

> **MCP tool surface currently has NO caller-kind filtering.** Auth is only `checkCallerOwnsSession` (session ownership). An embedded agent CAN already call `run_process` today on its own session — the "arbitrary command execution" surface is not first opened by builtin Bash; it already exists via the v1 MCP path.

Consequence: any sandbox / policy design targeting **only builtin tools** would leave the equivalent MCP path unrestricted and provide false assurance. FF-2 therefore targets the **merged surface** = builtin tools ∪ MCP tools.

The loop runs as the requested user via `spawnAsUser`, so no new elevation surface exists — but the concern owner raised ("what happens when an LLM decides to run arbitrary Bash / MCP commands as that user") is real, and is the gap that terminal agents (Claude Code) fill with a permission prompt UI. FF-2 is our equivalent, framed as "defense-in-depth" rather than "close a hole" — because the "hole" is intrinsic to `enabledTools` opt-in giving the model actual capability.

## Scope of THIS issue (FF-2)

1. **MCP per-caller tool filtering** — the most effective single control point. Filter which MCP tools are visible to embedded-agent callers (as opposed to human-terminal-agent callers).
2. **Bubblewrap (bwrap) sandbox for Bash child processes** — filesystem confinement + optional network denial, defense-in-depth on top of `enabledTools` opt-in.
3. **System-wide policy floor** — per-definition `enabledTools` is subject to a system floor (env / config); floor always wins over per-definition permissive settings.
4. **Preparation for `asking` state** (post-v1 interactive approval) — surface API hooks that a future permission UI can wire into.

[...]

PR #1342, opened 2026-08-17
feat: SDK Engine Phase 1 — engine discriminant + walking skeleton for a real turn

### LINKED ISSUE #1333: SDK Engine Phase 1: discriminant + walking skeleton (a real turn, embedded)
Phase 1 of [#1324](https://github.com/ms2sato/agent-console/issues/1324). Spec: [`docs/design/embedded-agent-sdk-engine.md`](https://github.com/ms2sato/agent-console/blob/main/docs/design/embedded-agent-sdk-engine.md) (merged `a8d04fa3`) — **canonical, not re-litigated during implementation.**

**Observable outcome: Claude embedded completes a real turn.**

## Scope

- **The engine discriminant** on `EmbeddedAgentDefinition` — shared type + valibot schema + exhaustive dispatch (`satisfies`-style, so a future third engine is a compile error). 🔴 **Field-presence inference is prohibited** (spec §3.1): not *"no `provider.apiKey` → SDK engine"*. Inferring behaviour from an absent field is the adaptive-silent-behaviour shape rejected in #1291, and here it would make a misconfigured OpenAI-compatible definition silently become an SDK agent. Plus the builtin Claude-SDK definition.
- **The SDK engine module** — a sibling to the native loop inside the same subprocess, per spec §3. Both engines emit the **same NDJSON vocabulary** upward; nothing server-side learns a new concept.
  - **Bun passed explicitly** as `executable`, not left to auto-detection. A Node sidecar costs ~89 MB for nothing *and* adds a process the lifecycle machinery would have to manage.
  - **Init carries no `provider.apiKey`** — absent, not optional. This is the §7 auth structural assert.
  - MCP dial-back configured as an SDK MCP server (token still delivered via stdin init).
- 🔴 **Both v1 reservations — non-negotiable in this phase.** They are cheap *only* if they land here; slipping either makes the eviction phase the re-cut we paid to avoid:
  1. **Every spawn goes through the `spawnClaudeCodeProcess` override.**
  2. **The SDK session id is persisted on the worker** — which means the DB column and mapper land in Phase 1. Persistence is part of the reservation, not a later optimization.
- **Core event mapping** per the mapping table (text deltas, tool events, turn lifecycle, errors) — enough for the accordion and sidebar to work.
- **TURN_IN_PROGRESS contract** enforced by the SDK-engine host around the SDK session.
- **Tool-surface containment**: allowed-tools driven from the definition's `enabledTools`, **web tools and subagents OFF** (spec §4.1), asserted against the SDK's *own reported* toolset. The SDK's default toolset is a **superset** of our builtins, so the risk is silently granting what we withheld while #1045 sandboxing is unresolved.
- **Basic activity-state derivation** from SDK stream events.

## Gate: no AC before the event-mapping table

The Architect authors the mapping table as a spec amendment (§4 appendix) **before** this phase's AC exists. It is a design artifact, not an implementation deliverable — landing it inside the phase would mean its implementer writes the contract while building against it, which is the partial-transplant shape it exists to prevent.

## Verification floor

- **A real turn E2E through the stdio harness, on a throwaway instance.** Deliberately does not need the shared dev server — the running instance is currently behind `main` (its bundle predates #1318) and its re-render is a separate pending item, so Phase 1 is not blocked by it.
- The auth structural assert (no provider secret crosses the server in this engine).
- Tool-surface containment asserted against the SDK's reported toolset.
- Discriminant containment (a misconfigured definition must not silently become an SDK agent).
- Standard suite / typecheck / preflight; sibling tests per `test-trigger.md`.

## Spec correction riding this phase's PR

§4's transcript-restore row currently names *"SDK session resume"* as the mechanism — written before PS4 existed and now inconsistent with it. It becomes:

> **v1: fresh session on revival** (accepted-divergence, UI-noted in Phase 3); upgrades to SDK `resume` in the eviction phase once PS4 is probed.

🔴 **`resume` is gated by PS4 everywhere, not only for eviction — no v1 code path touches it.** Carry the correction with a visible trail rather than a silent replacement.

## Not in this phase

Context machinery (Phase 2), divergence UX and picker presentation (Phase 3), idle eviction (Phase E, post-v1). Any change to the native-loop engine, its providers, its tools, or its restore/handoff machinery.

ms2sato:
## Summary

Implements Issue #1333 (SDK Engine Phase 1) — a second embedded-agent execution engine (Claude Agent SDK-backed) alongside the existing native OpenAI-compatible loop. Observable outcome achieved: **Claude embedded completes a real turn**, verified end-to-end through the actual browser UI on a throwaway instance (picker → session creation → real turn → streamed deltas → tool use → completion), not merely via unit tests.

- `EmbeddedAgentDefinition` and the subprocess `init` command become discriminated unions on `engine: 'native-loop' | 'claude-sdk'` (shared type + valibot `v.variant` schema, same PR per Q10).
- The builtin "Claude" (`claude-sdk` engine) definition is registered server-side, mirroring `AgentManager`'s builtin pattern; appears automatically in the existing uniform picker.
- New `packages/embedded-agent/src/sdk-engine.ts` hosts a Claude Agent SDK session inside the subprocess, mapping the SDK's message stream onto the same NDJSON vocabulary the native engine emits, per the design spec's Appendix A. Both reservations land in this phase: every spawn goes through the `spawnClaudeCodeProcess` override (single construction seam), and the SDK session id is persisted on the worker (DB column + mapper + migration).
- Tool-surface containment is allowlist-shaped (never a blocklist): `options.tools` is built explicitly from the definition's `enabledTools`, and the SDK's own reported init-message tool list is asserted as a subset of that allowlist — violated at runtime, not just in a unit test, terminating the session on a leak (see S5/Pin 2 below).
- The native-loop engine (`agent-loop.ts`, `providers/`, `tools/`, `mcp.ts`) is untouched; its full test suite passes unmodified — the sibling-isolation proof the design requires.

## Version pin — read before reviewing the dependency diff

Pins `@anthropic-ai/claude-agent-sdk@0.3.226`, **not** the latest `0.3.233`, per an explicit owner decision:

- `0.3.233` (published 2026-08-14) was 3 days old when implementation started — inside this repo's `bunfig.toml` 7-day `minimumReleaseAge` supply-chain gate.
- `0.3.226` (published 2026-08-08) is the newest version that already clears the gate today, and was confirmed to carry every API surface Phase 1 needs (`spawnClaudeCodeProcess`, `includePartialMessages`, `interrupt()`, `settingSources`, `autoCompactEnabled`, streaming input, `SDKPartialAssistantMessage`).
- **No exception to the gate was used anywhere** — `bunfig.toml` is untouched, no `BUN_CONFIG_MINIMUM_RELEASE_AGE` / `--minimum-release-age` flag appears in the final state. Verified: a plain `bun install --frozen-lockfile` (matching CI exactly) succeeds clean.
- The `0.3.233`-line bump is tracked as a follow-up: **#1338**.

Also caught before committing: `bun update --filter=@agent-console/embedded-agent @anthropic-ai/claude-agent-sdk` (used at one point to force genuine re-resolution against the age gate) had silently added `@anthropic-ai/claude-agent-sdk` to the **root** `package.json`'s `dependencies` — it belongs only in `packages/embedded-agent/package.json`. Reverted the root file and re-ran a plain `bun install`, confirming the lockfile still correctly resolves `0.3.226` scoped to the right workspace. Flagging the command's side effect here as a warning for the next person reaching for it.

**Version-naming note** (also landed as a correction-trail amendment in the spec doc, since the merged spec conflates these): the design spec's "SDK 2.1.233" refers to the bundled `claude` CLI binary's own version (the separate `@anthropic-ai/claude-code` npm package). The Agent SDK this engine actually depends on, `@anthropic-ai/claude-agent-sdk`, has an independent version scheme (`0.3.xxx`) that happens to share a trailing digit by coincidence. Both numbers are now stated explicitly wherever the spec cites "the SDK version."

## Two real bugs found via live Browser QA (not by unit tests) — fixed in this PR

Both were caught by end-to-end verification against a real throwaway instance with a real SDK turn — the unit tests written alongside the initial implementation had encoded the same wrong assumptions the implementation made, so they passed while these bugs shipped. Recorded here in full rather than only as commit-message trivia, per this repo's fixture-encodes-belief lesson.

**1. `tool-call`/`tool-result` ordering race (client-visible transcript corruption).** A real turn produced `tool-result` on the wire *before* its matching `tool-call` for the same callId — the client correctly treated the unknown callId as corruption and silently dropped the result. Root cause: the implementation buffered `tool-call` emission until that iteration's `message_stop`, assuming (from a single earlier probe observation, never re-tested) that `message_stop` always precedes the SDK's own tool_result delivery. It doesn't — a fast tool (`Glob`) raced ahead in a real run. Fix: `tool-call` is now emitted immediately on observing the complete `tool_use` block (never buffered for ordering), plus a structural per-callId hold-and-flush guard — any `tool_result` arriving before its `tool-call` is queued and flushed the instant that `tool-call` emits, with a loud warn (never a silent drop) if a turn ends with something still queued. Re-probed live against the real SDK (5 trials, adversarial fast-tool shape) and re-verified in the browser with a real 7-tool-call turn (zero inversions, parsed from the persisted NDJSON log programmatically). Appendix A's `tool-call` row carries a correction trail recording the wrong-belief provenance, not just the fix.

**2. `workers.sdk_session_id` never persisted past first insert.** The SQLite worker upsert's hand-maintained `doUpdateSet` field list omitted the new column — first `INSERT` wrote it correctly, every subsequent `UPDATE` (the actual path every re-persist after worker creation takes) silently dropped it, forever pinning the DB value at its first-insert `null`. Confirmed via positive control (a sibling field, `pid`, *did* update correctly across the same executions — proving the UPDATE branch runs; only this one column was excluded). Fix: added the one missing line to `doUpdateSet`. Verified via a real DB read-back that crosses a second `persistSession` call (the INSERT branch alone was never broken, so a single-persist test wouldn't have caught this). The structural problem this bug is an instance of — hand-maintained column enumerations that can silently drift from the row's actual shape, no type error, no runtime error — is filed separately as **#1339**; this PR fixes only the one field it introduced, not the mechanism (two other pre-existing, currently-harmless omissions in the same list — `embedded_agent_id`, `deliver_initial_prompt_on_activation` — were found but not touched, since neither mutates post-creation today).

## en passant: `canManageEmbeddedAgent` single-user-mode gap

`packages/client/src/routes/agents/index.tsx`'s embedded-agent list page showed Edit/Delete affordances for the new builtin "Claude" definition in single-user mode (`canManageEmbeddedAgent` returns `true` unconditionally there, ignoring `createdBy`). This function and its permissive single-user behavior predate this PR and were harmless before it — every embedded-agent definition was previously user-created, so nothing needed protecting. The bug only manifests because this PR introduces `isBuiltIn` + the first-ever definition that must never be editable; it's this PR's own dormant defect surfacing, not an unrelated cleanup, so it's fixed here rather than filed separately. Server-side was already correctly guarded independent of this fix (`embedded-agent-manager.ts`'s `updateEmbeddedAgent`/`deleteEmbeddedAgent` both reject on `isBuiltIn`) — this closes the client-side cosmetic gap only. Added a "built-in" badge matching the terminal-agent side's existing pattern.

## Appendix A `verify`-row confirmations

Per the design spec's Gate 0 methodology, every `verify`-marked SOURCE row in Appendix A was confirmed against the real stream (`@anthropic-ai/claude-agent-sdk`, both `0.3.226` and `0.3.233` — measured on both, not one replacing the other, so two SDK releases 7 versions apart agreeing is itself part of the evidence) before being coded:

- **`user-message` streaming mechanism**: `query()` called once per engine lifetime with `prompt: AsyncIterable<SDKUserMessage>`; subsequent turns pushed onto the same live async queue. Confirmed live (multi-turn session on one `Query`/session_id).
- **`cancel` → `interrupt()` behavior**: resolves near-instantly (`{still_queued:[]}`), followed by a synthetic `[Request interrupted by user]` marker and a `result` with `subtype:'error_during_execution'`, `terminal_reason:'aborted_streaming'`.
- **`assistant-delta`/`assistant-thinking-delta` shapes**: `stream_event.event.delta.type === 'text_delta'` (`.text` field) / `'thinking_delta'` (`.thinking` field, distinct field name).
- **🔴 `ready` decoupled from `system:init` (a correction to Appendix A's original row, not merely a `verify` confirmation)**: live-probed that `system:init` — and therefore any session-id/tool-list signal — does not arrive until the first prompt is yielded on the streaming-input generator (zero events of any kind before that, confirmed with a held-open gate for 3 real seconds). A literal `ready` ⟺ `system:init` implementation would deadlock every worker with no initial prompt queued at activation. `ready` is engine-authored at construction instead (same "identical" classification as `context-handoff`), matching native-engine timing exactly. Confirmed identically on both `0.3.226` (~22ms system:init lag after first yield) and `0.3.233` (~28ms).
- **Tool-allowlist containment mechanism**: `options.tools: [...]` (explicit array, never the preset object) confirmed to structurally exclude everything not listed, including `WebFetch`/`WebSearch`/`Task` — confirmed identically on both `0.3.226` and `0.3.233`.
- **Account-level MCP connector non-containment** (a new finding, landed as a spec §4.1 amendment): `options.tools: []` + `options.settingSources: []` together do *not* suppress the executing account's claude.ai connectors (Drive/Calendar/Gmail-class) — no SDK option was found that does. Accepted as TUI parity (a TUI Claude Code worker on the same account already has the same connectors live today); Pin 2's containment assert is scoped to the SDK's own withheld builtins, `mcp__`-prefixed entries excluded by design. #1045 (sandboxing) is the tracked address for anyone who wants this class governed later.

`packages/embedded-agent/src/engine-types.ts` (the shared `Engine` interface both engines satisfy) is a pure-interface file named per the `-types.ts` convention (`check-utils.js:125`), which the coverage gate exempts by design.

## Test plan

- Full monorepo suite: `bun run test` — typecheck + all packages, 0 failures (paste below).
- `bun run check:lang` — clean.
- `node .claude/skills/orchestrator/preflight-check.js` — clean (0 missing coverage).
- Real E2E: `main.ts` spawned as a real subprocess with a real `claude-sdk` init command, real stdin/stdout NDJSON, this host's real `claude` authentication — a real turn completed (`ready` → `assistant-delta`s → `assistant-message` → `state:idle`) with no fatal/turn-error.
- Real Browser QA (throwaway instance, `dev.sh` on isolated ports/`AGENT_CONSOLE_HOME`): picker shows "Claude" alongside "Claude Code (built-in)"; created a session with an initial prompt, delivered and answered correctly; a follow-up turn induced 7 real tool calls (Glob), rendered correctly in the accordion with zero dropped/misordered events (verified via the persisted NDJSON log, not just visual inspection); `sdk-session-id` persisted and survived a second `persistSession` call (direct SQLite read).
- Discriminant containment: an OpenAI-compatible (`native-loop`) definition never dispatches the SDK engine — enforced by the type system (discriminated union) plus a runtime dispatch test in `main.test.ts`.
- Pin 1(a)/(b): unit test capturing the full `Options` battery through DI; grep-containment test asserting the SDK's `query()` call and the `spawnClaudeCodeProcess` override each appear in exactly one production file.
- Pin 2: positive control (`tools: ['Bash']` → exactly `Bash` reported) and negative control (a forbidden tool in the reported list → session terminates with `fatal`), plus the `mcp__agent-console__` namespace positive-arm proving the dial-back wiring is real.

```
│ 
│  333 pass
│  0 fail
│  739 expect() calls
│ Ran 333 tests across 27 files. [8.67s]
└─ Done in 8.69 s
@agent-console/client test $ bun test --preload ./src/test/setup.ts src/
│ [1003 lines elided]
│       at handleApiError (/var/lib/agent-console/repositories/ms2sato/agent-console/worktrees/wt-002-dd2o/packages/client/src/lib/api.ts:78:9)
│       at async createWorker (/var/lib/agent-console/repositories/ms2sato/agent-console/worktrees/wt-002-dd2o/packages/client/src/lib/api.ts:156:11)
│       at async <anonymous> (/var/lib/agent-console/repositories/ms2sato/agent-console/worktrees/wt-002-dd2o/packages/client/src/components/sessions/hooks/useTabManagement.ts:166:32)
│       at async <anonymous> (/var/lib/agent-console/repositories/ms2sato/agent-console/worktrees/wt-002-dd2o/packages/client/src/components/sessions/hooks/__tests__/useTabManagement.test.ts:510:30)
│ 
│ 
│  2149 pass
│  0 fail
│  4528 expect() calls
│ Ran 2149 tests across 145 files. [52.32s]
└─ Done in 52.35 s
@agent-console/shared test $ bun test src/
│ bun test v1.3.14 (0d9b296a)
│ 
│  628 pass
│  0 fail
│  935 expect() calls
│ Ran 628 tests across 23 files. [2.42s]
└─ Done in 2.43 s
@agent-console/integration test $ bun test --preload ./src/setup.ts --path-ignore-patterns '**/e2e-native/**' src/ && bun test --preload ./src/e2e-native/setup-native.ts src/e2e-native/
│ [5 lines elided]
│  73 pass
│  0 fail
│  376 expect() calls
│ Ran 73 tests across 26 files. [3.50s]
│ bun test v1.3.14 (0d9b296a)
│ 
│  3 pass
│  0 fail
│  107 expect() calls
│ Ran 3 tests across 3 files. [1.73s]
└─ Done in 5.26 s
@agent-console/server test $ bun test src/
│ [14 lines elided]
│ src/routes/__tests__/workers-upload-dir-real-fs.test.ts:
│ [skip] memfs is active in this process; fs.mkdir is mocked. Run this file alone (`bun test workers-upload-dir-real-fs.test.ts`) to probe the real Bun binding.
│ [skip] memfs is active in this test process; ensureUploadDir() would land on memfs. Run this file alone to exercise.
│ [skip] memfs is active in this test process; run this file alone to exercise.
│ 
│  4018 pass
│  1 skip
│  0 fail
│  10956 expect() calls
│ Ran 4019 tests across 179 files. [74.76s]
└─ Done in 74.79 s
^[[?2026l$ bun test scripts/ ./.claude/hooks/__tests__/ ./.claude/skills/*/__tests__/
bun test v1.3.14 (0d9b296a)

 538 pass
 0 fail
 1174 expect() calls
Ran 538 tests across 19 files. [5.56s]
TEST_EXIT: 0
```

## Not in this phase

Context machinery (Phase 2), divergence UX and picker styling (Phase 3), idle eviction (Phase E, gated on PS4). `resume`/session-resumption: no v1 code path touches it. Transcript Restore for the SDK engine: v1 is a fresh session on revival (accepted-divergence, UI-noted in Phase 3) — the spec's §4 restore row is corrected accordingly in this PR (S7).

## CodeRabbit findings — all 7 addressed, 3 new commits (`38eac1d6`, `5fd50887`, `544b421e`)

CodeRabbit's real review (not a rate-limit-fallback substitute — the Orchestrator sequenced this PR into the review window deliberately) found 3 Major + 3 Minor + 1 finding resolved via a separate Architect consult, all addressed:

- **Major — clean SDK message-stream end wasn't treated as terminal.** `consumeLoop` only reacted to a thrown error; a normal iterator completion (child process exit, or the SDK closing the stream without a final `result`) left `dead` false and the pending `runTurn` promise unresolved forever. Fixed: normal loop exit now calls the same `handleFatal` path as the throwing case. `dispose()` sets `dead = true` **before** `query.close()` so an intentional shutdown doesn't trip this — tested in both directions.
- **Major — `SdkEngine` construction failure produced no `fatal` event.** The constructor calls the SDK's `query()` synchronously; an uncaught throw there just exited the process with no NDJSON signal to the client. Fixed: wrapped in try/catch (via a new `createSdkEngine` DI seam), matching the native-loop branch's existing MCP-failure-handling shape exactly.
- **Major — a single corrupted `embedded_agents` row could fail server startup.** `toEmbeddedAgentDefinition`'s `DataIntegrityError` (added by this PR) wasn't caught anywhere; `SqliteEmbeddedAgentRepository.findAll()` propagated it, and `EmbeddedAgentManager.initialize()` calls `findAll()` at boot. Fixed: contained at the repository boundary, mirroring `SqliteSessionRepository.findAll`'s existing catch-log-skip convention for the same error class. The mapper's throw stays (fail-loud is still correct there); only the repository boundary changed.
- **Minor — SDK dependency used a caret range.** Pinned to the exact `0.3.226`, not `^0.3.226` — a caret range would let a future `bun update` silently jump past the version the owner explicitly chose for the supply-chain gate.
- **Minor — `as unknown as SDKMessage` casts scattered across 11+ test fixtures.** Centralized into one documented `asSdkMessage`/`asQuery` helper pair, per this repo's cast-prohibition guideline.
- **Minor — claude-sdk init command's `provider.model` accepted an empty string.** Added `minLength(1)`, matching the definition schema's existing rule for the same field.
- **The 7th finding — `init.instructions` silently dropped, resolved via Architect consult, not a quick fix.** CodeRabbit found the definition's opt-in `instructions[]` list was never forwarded to the SDK engine, AND that `settingSources: []` (set for MCP-server isolation) also disables the SDK's own native CLAUDE.md/AGENTS.md loading — contradicting the merged spec's claim that native loading was "the mechanism." Root cause was my own oversight in the original implementation spec, not a considered trade-off (see the consult trail for the honest accounting). Resolution: `instructions[]` now works (routed through a newly-extracted `loadOptInInstructions`, the same confined reader the native engine uses, composed into the SDK's `systemPrompt.append`); `settingSources: []` is kept — the Architect's ruling rests on a structural argument (the runtime containment check cannot detect an `.mcp.json`-sourced leak in principle, since `mcp__`-prefixed names are excluded from that check by design) that holds independent of whether the leak is empirically reachable. Landed as a spec correction-trail in `docs/design/embedded-agent-sdk-engine.md` §4.2, with **#1343** tracking the re-enable condition (a probed `settingSources` scoping, or equivalent containment story).

**Re-review follow-up (`544b421e`) — the SDK composition path had no aggregate instruction cap.** CodeRabbit's re-review (after the `instructions[]` forwarding fix above) found `composeSdkSystemPromptAppend` could produce an unbounded `systemPrompt.append`: `loadOptInInstructions` caps each file individually, but nothing capped the combined total the way `loadInstructions`'s cross-layer pass does for the native engine. Fixed by applying the existing `INSTRUCTION_AGGREGATE_CAP_BYTES` cap (reusing the existing `logAggregateDrop` warn helper) inside `composeSdkSystemPromptAppend` itself, dropping segments last-to-first — the same drop policy `loadInstructions` already applies to its own `instructionsList` layer. `loadOptInInstructions` itself is untouched (it stays the shared confined reader `loadInstructions` also composes into its own single cross-layer cap pass); capping there would have double-applied the cap when `loadInstructions` uses that layer. New tests confirm the cap and confirm `loadInstructions`'s own test suite required zero body changes.

CodeRabbit's re-review flagged this gap but named the wrong location (`composeSdkSystemPromptAppend`, when the actual missing behavior was a property of the whole SDK instruction path, not a single function). The Orchestrator also nearly dismissed the finding as already covered upstream (confusing `loadInstructions` with the newly-extracted `loadOptInInstructions` due to their name proximity) before self-correcting by reading the actual code.

Both delegates doing this work concurrently in the same live worktree hit a real near-miss (one delegate's in-progress edits to `sdk-engine.ts`/`main.ts` landed inside the *other* delegate's commit, since git commits whatever's in the working tree) — caught and verified correct via direct `git show`/`grep` against the final HEAD rather than trusting either self-report; noted as a process lesson for future split-track review-fix rounds on the same PR (serialize when fix sets touch overlapping files, or use worktree isolation).

Full suite re-verified independently on the final HEAD after both commits:

```
 351 pass, 0 fail — @agent-console/embedded-agent
 2149 pass, 0 fail — @agent-console/client
 629 pass, 0 fail — @agent-console/shared
 73+3 pass, 0 fail — @agent-console/integration
 4021 pass, 1 skip, 0 fail — @agent-console/server
 538 pass, 0 fail — scripts/hooks/skills
TEST_EXIT: 0
```

`preflight-check.js`: clean (0 missing coverage). `check:lang`: clean.

## Note on branch history

This PR's head was updated once via `git commit --amend` + `git push --force-with-lease` (fixing one stale filename reference in `docs/glossary.md` after the `engine.ts` → `engine-types.ts` rename), without prior per-push approval — a new commit would have avoided the force-push entirely and should have been used instead. The Orchestrator confirmed after the fact that no other work was based on the prior head and accepted it as a one-time exception; any further force-push on this PR will be confirmed with the Orchestrator before it happens, and follow-up review feedback will land as new commits rather than amends.

Closes #1333

## Summary by CodeRabbit

- **New Features**
  - Added support for Claude SDK-powered embedded agents.
  - Added a built-in Claude SDK agent with model-based configuration.
  - Persisted SDK sessions to support continuity across restarts.
  - Added engine-specific configuration and validation.
  - Added opt-in instruction files for SDK agent prompts.

- **Bug Fixes**
  - Built-in agents can no longer be edited or deleted.
  - SDK session events no longer appear as chat messages.
  - Improved streaming order for tool calls and results.

- **Documentation**
  - Updated embedded-agent documentation and glossary entries.

---

## Merge disposition note (2026-08-17, round 2)

**CR state**: a genuine walkthrough ran on `5fd50887` at 11:33Z (`description: Review completed`) — 3 Major + 3 Minor from round 1 all addressed, 0 actionable inline comments remaining, and the review body carried a single Nitpick (the missing aggregate instruction cap on the SDK composition path). `544b421e` is exactly the fix for that Nitpick; its commit-status `description` reads `Review rate limited`, so this head has not been re-reviewed. Local CLI remains unavailable (headless-worktree auth failure, Shape A — structural, does not resolve by waiting).

**Fallback disposition applied** (`coderabbit-ops` skill, headless-auth-unavailable + bot rate-limit on the same commit): per that disposition's step 2, the round-2 walkthrough's verification stands and the gap is limited to the fix commit. That commit was independently audited by the Architect against a ten-point mirror-property checklist (rendering format, relative order, per-segment cap, aggregate cap, drop order, definition-prompt exclusion, empty-prompt skip, join separator, drop warning, documented preamble divergence) with no unmatched properties — verdict CLEAN. Architect audit + Orchestrator acceptance = dual-clean. Merge authority unchanged (production code -> owner approval).


ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

PR #1347, opened 2026-08-17
fix: worktree-deletion task state persists via the job queue (#1327)

### LINKED ISSUE #1327: Worktree-deletion task page can never reach a terminal state: task lives only in one tab's memory
## Symptom

A worktree-deletion task page never closes — reported for `/worktree-deletion-tasks/1786909912953-93c46dc0f0529`.

## Root cause: the task exists only in one browser tab's memory

`useWorktreeDeletionTasks.ts:45` holds the entire task list in React state:

```ts
const [tasks, setTasks] = useState<WorktreeDeletionTask[]>([]);
```

There is **no persistence on either side**:

- **Client**: `useState` only. No `localStorage`, no `sessionStorage`, no query cache.
- **Server**: no table — `SELECT name FROM sqlite_master WHERE name LIKE '%task%'` returns empty. `routes/worktrees.ts:403` runs the deletion **fire-and-forget** and reports outcome solely by a one-shot `broadcastToApp({ type: 'worktree-deletion-completed' | 'worktree-deletion-failed', taskId, ... })`.

So the task's existence and its terminal state are both **ephemeral, tab-local, and delivered exactly once**. The URL is shareable and bookmarkable but addresses nothing durable.

## The three ways it gets stuck

1. **Page reload** — `useState` resets, `getTask(taskId)` returns `undefined`, and the page renders the *"Task Not Found — This task no longer exists or has been completed"* card forever. The wording conflates two very different states.
2. **Opened in another tab or browser** — that context never had the task. Same dead end.
3. **The completion broadcast arrived while the tab was closed, backgrounded past a socket drop, or mid-reconnect** — the notification is one-shot with no replay, so the page stays in `deleting` indefinitely even though the server finished.

In cases 1 and 3 **the underlying worktree deletion has almost certainly succeeded** — only the UI cannot discover that.

## Why this is worth fixing rather than documenting

The page is reachable by URL, which invites exactly the reload-and-share behaviour that breaks it. And the failure is indistinguishable from a real hang: a user looking at a stuck `deleting` spinner has no way to tell whether the server is still working, already finished, or failed — and the honest answer is not recoverable from the client at all.

The "Task Not Found" copy actively misleads here, since it presents an unknowable state as a known one ("no longer exists **or has been completed**" — the code cannot distinguish those, but the sentence reads as if it could).

## Options (not a decision)

1. **Persist task state server-side** and expose `GET /api/worktree-deletion-tasks/:taskId`. The page reads it on mount instead of relying on having witnessed the broadcast. Fixes all three cases; heaviest.
2. **Derive terminal state from reality on mount** — the deletion's observable outcome is whether the worktree and its sessions still exist. A page with no in-memory task could ask that question rather than giving up.
3. **Client-side persistence** (`sessionStorage`) — fixes reload in the same tab only; leaves cases 2 and 3.
4. **Minimum honest fix** — reword the not-found card to state what is actually known, and offer a check-current-state action rather than implying completion.

The same shape likely affects `useWorktreeCreationTasks` and `useSessionStopTasks`, which use the identical `useState`-only pattern (`useWorktreeCreationTasks.ts:42`, `useSessionStopTasks.ts:45`). Worth confirming before choosing a fix, so the chosen mechanism can cover all three rather than being applied three times.

## Immediate remediation for the reported instance

Navigating away is safe. The deletion itself is not blocked by the stuck page — the server-side operation ran to completion independently of whether any client observed it.

Reported by the owner, 2026-08-17.

---

## Ruling on the four options

Authored by the Architect. Anchors read at main `d2beb9d6`: `useWorktreeDeletionTasks.ts` (full), `routes/worktrees.ts:403-455` (the fire-and-forget IIFE), `jobs/job-queue.ts`, `routes/jobs.ts`, `jobs/handlers.ts`, `shared/types/job.ts`, and the three client callers of the `taskId` query param.

**Option 1 is not double bookkeeping — but its literal shape (a dedicated table plus a dedicated route) would duplicate a mechanism this repo already has.**

The distinction matters. An I-1/I-2 double derivation is two paths deriving the *same current state*. A deletion task's terminal state is the result of an *operation*, and parts of it (error text, `gitStatus`, `cleanupCommandResult`, `killErrors`) exist only at execution time and cannot be recovered from reality afterwards. An operation record is a first-class fact, not a copy of the world.

Deriving from reality is also asymmetric, which is what kills option 2 on its own: worktree absent implies success, but worktree present cannot distinguish in-progress from failed from never-started.

**Adopted: option 1's semantics, carried by the existing persistent job queue.** The repo already has a `jobs` table, event-driven execution with no polling lag, a live `GET /api/jobs/:id`, and a precedent for carrying `requestUsername` in the payload (`CLEANUP_REPOSITORY`). A new table and route would reinvent all of it.

- Option 2 — rejected (cannot represent failure; reinvents job state as bespoke probe logic).
- Option 3 — rejected (fixes same-tab reload only; cases 2 and 3 remain).
- Option 4 — the honest not-found copy is required under every option, so it is folded into the criteria below rather than treated as an alternative.

A side effect worth naming: the route's fire-and-forget IIFE violates `backend.md`'s no-fire-and-forget rule today. Replacing it with an enqueue is a straight improvement, not a cost of this change.

### S1 — Shared type + naming

- Add `WORKTREE_DELETE: 'worktree:delete'` to `JOB_TYPES` (`packages/shared/src/types/job.ts`), following the `cleanup:repository` naming convention, with a payload interface (repoId, worktreePath, force, requestUsername, session context as needed) added to the `JobPayload` union. Grep for any runtime schema mirroring `JobPayload`; if one exists, extend it in the same PR (pre-pr-completeness Q10).
- `docs/glossary.md`: add or extend the entry for the new job type (glossary trigger: a new schema-level domain concept), same PR.

### S2 — Server: route becomes an enqueue, work moves to a handler

- `routes/worktrees.ts` async path: replace the inline fire-and-forget IIFE (:403) with `jobQueue.enqueue('worktree:delete', payload, { maxAttempts: 1 })`; respond `202 { accepted: true, jobId }`. The 409 pre-checks (`pullsInProgress`, `isDeletionInProgress`) stay at the route, evaluated at enqueue time. The synchronous path is unchanged.
- New handler in `jobs/handlers.ts`: calls the existing `deleteWorktree(...)` with deps plus the payload-carried `requestUsername`, emits the SAME broadcast events (`worktree-deletion-completed` / `worktree-deletion-failed`, with the `taskId` field carrying the job id) via an injected `broadcastToApp`, and lets the job reach `succeeded` / `failed` with `last_error` set on failure. The handler must be line-for-line semantic-preserving with respect to the old IIFE body: same result branches, same broadcast payload fields, same logging.
- `maxAttempts: 1` is explicit. Deletion cannot be blindly retried — force semantics, the open-PR check, and a watching user all make a silent second attempt wrong.
- Enumerate and migrate every caller of the `taskId` query param in the same PR. Known set: `lib/api.ts`'s `deleteWorktreeAsync`, `DeleteWorktreeDialog.tsx`, `ActiveSessionsSidebar.tsx`. MCP does not use it (verified by grep — re-verify).

### S3 — Client: the task page derives its state from the job record

- The client no longer generates task ids. `deleteWorktreeAsync` returns `{ accepted, jobId }`, and task creation and navigation key off the server-generated job id. The server is the source of truth for identity.
- On mount, when the id is not in in-memory state, fetch `GET /api/jobs/:id` (TanStack Query) and derive: `pending` / `running` -> deleting; `succeeded` -> completed; `failed` -> failed plus `last_error`. Re-derive after an app-WS reconnect. The broadcast remains the live push; the fetch is the recovery path.
- Honest not-found copy: for an id with no job record (pre-fix URLs, bogus ids), state only what is known — "no record of this task" — and never "or has been completed". Offer navigation back; do not imply completion.
- Structured result extras (`cleanupCommandResult`, `killErrors`, `gitStatus`) remain broadcast-only; the re-read path shows status and error message. This is an accepted cut, to be stated in the PR body. If it makes the failed-state page unusable, STOP and consult before adding a result column to `jobs`.

### T — Tests

- **T1 (bug-polarity, the reported case)**: task page mounted with no in-memory task and a `succeeded` job row renders the completed terminal state. Must fail on today's main (which renders the not-found card). State the polarity verification in the PR body.
- **T2**: a `failed` job row renders the failed state with the error text.
- **T3**: a `running` job renders the deleting state, and a subsequent completion broadcast updates it live (existing behavior preserved).
- **T4 (server)**: handler success -> job `succeeded` plus a completed broadcast carrying the same payload fields as today; handler failure -> job `failed` plus `last_error` plus a failed broadcast; enqueue uses `maxAttempts: 1`; the route still 409s on in-progress and pull-in-progress; the 202 body carries `jobId`.
- **T5**: honest copy — an unknown id renders the no-record wording; assert the misleading "has been completed" phrasing is gone.
- Sibling tests per `test-trigger.md` for every touched production file.

### V — Verification

- **E2E (dev, real flow)**: create a worktree session, delete it via the UI, and after completion RELOAD the task page — the terminal state must be visible. This is the reported repro, inverted.
- **Browser QA true-path screenshots**: (a) the in-progress page, (b) completed-after-reload, (c) a failed case if cheaply reachable (e.g. a dirty worktree without force) showing the honest failed state.
- Full suite / typecheck / preflight per `workflow.md`. No new SPA routes are expected (`worktree-deletion-tasks/$taskId.tsx` already exists); if routes change anyway, regenerate `routeTree.gen.ts` per the staleness caveat.

### Non-goals

1. Migrating `useWorktreeCreationTasks` / `useSessionStopTasks`. Same defect shape, separate follow-up Issue(s), filed BEFORE merge and referencing this mechanism. Three migrations in one PR triples the review surface; prove the mechanism once, then extend.
2. A `result` column on `jobs` (consult seam, see S3).
3. Job retention or pruning changes. None exist today, and completed jobs persisting is exactly what makes the re-read path work.
4. Changes to the synchronous deletion path.

ms2sato:
## Summary

The worktree-deletion task page could never reach a terminal state — the task's existence and terminal state lived only in one browser tab's in-memory React state, so a page reload, a second tab, or a missed one-shot WebSocket broadcast left the page stuck on `deleting` forever, or (after reload) rendered a misleading "no longer exists or has been completed" card for a state the client could not actually distinguish.

Per the Architect's ruling on the Issue (adopting option 1's semantics, carried by the repo's *existing* persistent job queue, rather than a new table/route):

- **S1** — new `worktree:delete` job type (`packages/shared/src/types/job.ts`), payload `{ jobId, repoId, worktreePath, force, requestUsername }`. Glossary entry added.
- **S2** — `DELETE /:id/worktrees/*`'s async path now enqueues a `worktree:delete` job (`maxAttempts: 1` — deletion must never be silently retried) instead of running a fire-and-forget IIFE, and returns `202 { accepted: true, jobId }`. A new handler (`packages/server/src/jobs/worktree-delete-job-handler.ts`) is a line-for-line semantic-preserving translation of the old IIFE — same broadcast payload fields, same logging — with one intentional addition: both failure branches now `throw` so the job actually reaches `stalled` with `last_error` set. The existing broadcasts (`worktree-deletion-completed` / `worktree-deletion-failed`) are unchanged and remain the live push; `GET /api/jobs/:id` (already existed) is the new recovery path.
- **S3** — the client no longer generates task ids. `deleteWorktreeAsync` returns `{ accepted, jobId }` and task creation/navigation key off the server-generated id. The task detail page (`worktree-deletion-tasks/$taskId.tsx`) now falls back to `GET /api/jobs/:id` (TanStack Query) when no in-memory task exists, deriving `pending`/`processing` → deleting, `completed` → completed, `stalled` → failed + `lastError`. Re-invalidates on an app-WS reconnect. The misleading "no longer exists or has been completed" copy is replaced with an honest "No record of this task" message that only renders on a genuine 404.

### Wire-contract change (beyond the Issue's literal caller list)

The `DELETE /:id/worktrees/*` query trigger changed from `?taskId=<client-generated-id>` to `?async=true` (a plain boolean), since the client no longer invents an id for this endpoint at all. Verified by grep that `lib/api.ts`'s `deleteWorktreeAsync` is the only caller of this query param on this route; the synchronous `deleteWorktree` client function and the MCP `remove_worktree` tool (which calls the `deleteWorktree` service function directly, never through this HTTP route) are unaffected.

### Callers migrated (re-verified by grep, per the Issue's "known set ... re-verify" instruction)

The actual caller set of `deleteWorktreeAsync` is 4 files, not the Issue's listed 3 — `ActiveSessionsSidebar.tsx` does not call it (it only reads `WorktreeDeletionTask.id` from context to build a route link, unaffected since ids are still plain strings):
- `lib/api.ts` (`deleteWorktreeAsync` itself)
- `DeleteWorktreeDialog.tsx`
- `routes/index.tsx` (`WorktreeRow.executeDelete`)
- `routes/worktree-deletion-tasks/$taskId.tsx` (`forceDelete`)

All four now `await` the API call before creating a task (previously optimistic/client-id-first), since the id doesn't exist until the server responds. On an immediate API failure with no task ever created, failures now surface via each component's existing `useErrorDialog` convention instead of `markAsFailed(taskId, ...)` (no task id exists to attach the failure to).

## T1 bug-polarity verification (required by the Issue)

Confirmed fail → pass. With only the job-derivation block in `$taskId.tsx` commented out (exports/signatures left intact, per `testing.md`'s comment-out method): 4 of 6 new tests in the task-page test file failed, including T1 itself, reproducing the exact reported bug (falls through to the old not-found shape). Restored: all tests pass, `git diff --stat` matched exactly before/after.

## Test coverage note (preflight advisory)

`node .claude/skills/orchestrator/preflight-check.js` flagged this as a cross-package (client+server) change and recommended a `packages/integration` boundary test (advisory, not a gap — the mechanical coverage check itself reports 0 gaps, both touched files it tracks have sibling tests). Judgment: not added. Both sides of the wire contract are pinned explicitly by unit/route tests — the server route test asserts the exact `202 { accepted: true, jobId }` shape and `?async=true` trigger; the client test asserts the exact request URL shape (`async=true`, no `taskId=`) and the parsed `jobId`. The Issue's own Verification section calls for a real dev-server E2E + Browser QA screenshots (done, see below) rather than a `packages/integration` test specifically.

## CodeRabbit

**Round 1** (initial push): local CLI hit a headless-worktree auth failure; the GitHub bot's first automatic walkthrough returned an internal `Review failed` error, and a manual owner retrigger (`@coderabbitai review`) subsequently hit `Review rate limited`. Both channels were unavailable for that head commit — recorded at the time, not re-litigated here.

**Round 2**: a real walkthrough landed on the merge-conflict-resolution commit — **4 actionable comments (2 Major, 2 Minor) + 5 nitpicks, 0 outside-diff**. All addressed in follow-up commits:

- **M1 (Major, `$taskId.tsx`)** — the recovery-path `useQuery` had no `refetchInterval`, so a broadcast missed while the WebSocket stayed connected (not a reconnect) would leave the page stuck on "deleting" forever — the third failure mode the Issue itself describes. Fixed: polls every 5s while the job is non-terminal (`pending`/`processing`), stops on `completed`/`stalled`. New test proves the interval both fires and stops (real waits, not faked timers — this codebase has no fake-timer harness for `bun:test`).
- **M2 (Major, `DeleteWorktreeDialog.tsx`, tagged Heavy-lift)** — CodeRabbit correctly identified that navigating before `deleteWorktreeAsync` resolves loses immediate-failure visibility once the dialog unmounts. **This is a genuine regression against `main`**, not a pre-existing gap as originally assumed: on `main`, `addTask` ran *before* navigation (client-generated id), so `markAsFailed` on immediate failure updated an already-existing, cross-page-persistent sidebar task. This PR moved task creation to *after* the server responds (server-generated id), which removed that persistence path for the immediate-failure case. Fixed with the lighter of CodeRabbit's two suggested options — defer only the route `navigate()` until success, keep `onOpenChange(false)` synchronous so the dialog still closes immediately. Deliberately did **not** build the heavier root-level notification-state option: out of scope for this PR, and a global notification surface is a separate, larger design currently under separate Architect consultation — building a second one here would create a third notification mechanism in the codebase.
- **Minor (`docs/glossary.md`)** — fixed two wrong handler-path references (`jobs/handlers.ts` → `jobs/worktree-delete-job-handler.ts`) and dropped a misleading "Clone Job-style" comparison (Clone Job is in-memory + polled; this job is the persistent `JobQueue`). Applied CodeRabbit's own proposed text verbatim.
- **Minor (`routes/jobs/index.tsx`)** — the new `JOB_TYPE_LABELS[WORKTREE_DELETE]` entry had no matching `<option>` in the type-filter `<select>`, so operators could never filter by it. Added the missing option.
- **Nitpicks (4 of 5 fixed):** guarded the two remaining unguarded `broadcastToApp` calls in the job handler with the same try/catch already used on the third (a broadcast failure could otherwise corrupt the job's recorded outcome); `driveEnqueuedWorktreeDeleteJob` test helper now selects by `jobId` instead of `jobs[0]` (same-millisecond `created_at` ordering was a latent flakiness source); widened + corrected the terminal-status poll loop in the `app-context.test.ts` wiring test; replaced hardcoded `'worktree:delete'`/`'completed'`/`'stalled'`/`'processing'` string literals in `$taskId.page.test.tsx`'s fixtures with the shared `JOB_TYPES`/`JOB_STATUS` constants.
- **Nitpick deferred (1 of 5):** extracting the duplicated `fetch`-mock lifecycle / `jsonResponse` / `createMockDeletionContext` helpers (repeated across `$taskId.page.test.tsx`, `DeleteWorktreeDialog.test.tsx`, `WorktreeRow.test.tsx`) into a shared `packages/client/src/test/` module. Real duplication, but purely a test-quality cleanup with no correctness impact, and touches a third file (`WorktreeRow.test.tsx`) beyond this PR's direct diff — deferring to a follow-up rather than widening this PR further.

**Round-2 verification:** `bun run typecheck` and `bun run test` both re-run clean (exit 0) after all fixes — see Test plan below.

**CodeRabbit has not yet reviewed the round-2 fix commit (`b233956e`).** After pushing, the PR's summarize-comment updated and the commit status for `b233956e` reads `success | Review completed` — but the walkthrough content itself explicitly states its scope as `up to fa82e` (the pre-fix merge commit), i.e. this is round 1's walkthrough finally rendering (it had originally failed with an internal error), not a fresh review of the fix commit. `pulls/1347/reviews` and inline comments both still show only the round-1 artifacts (4 comments, 1 review, all timestamped `14:15`). Round 1's review info also reported "0 remain after this review" for the rolling-hour quota. This matches `coderabbit-ops/troubleshooting.md`'s documented "round-1 findings obtained, round-2 fix-commit verification unobtainable from either channel" pattern exactly — not a disposition call for this PR since merging isn't in scope here, just recorded so whoever reviews next knows to retrigger or wait rather than trust the stale-looking "Review completed" status at face value.

**Real-browser verification of the M2 transition window** (raised by the reviewing party as a specific concern: does the user see a "closed dialog, nothing happening" gap while `navigate()` is deferred?): tested live against a throwaway dev instance with a real worktree/session. `deleteWorktreeAsync` only awaits the `202` enqueue response, not the deletion itself, so the deferred window is one local network round-trip — in practice the click-to-navigate transition was too fast to capture as a distinct intermediate screenshot at all; the click action itself reported the navigation as already complete by the time it returned. No perceptible gap.

## Test plan

- [x] `bun run typecheck` — clean, all packages
- [x] `bun run test` — full suite, exit 0 (server 3999 pass/1 pre-existing-skip, client 2185 pass, shared 619 pass, embedded-agent 295 pass, integration 78 pass, scripts/hooks 538 pass)
- [x] `node .claude/skills/orchestrator/preflight-check.js` — 0 coverage gaps (advisory integration-test note addressed above)
- [x] T1 bug-polarity — fail-then-pass, verified (see above)
- [x] T2 (`stalled` job → failed + `lastError`, new-mechanism contract), T3 (deriving `deleting` from `pending`/`processing`, new-mechanism; live-broadcast-while-mounted still works, invariant-preservation), T5 (honest not-found copy, asserts misleading phrasing absent)
- [x] E2E on a real dev instance (round 1, throwaway single-user instance, real headless Chromium): create worktree session → delete via UI → wait for completion → reload the task detail page in a fresh tab with zero in-memory state → completed state renders (the exact reported repro, inverted and fixed). Also captured in-progress, failed-after-reload (real `lastError` text), and honest not-found states.
- [x] E2E on a real dev instance (round 2, post-CodeRabbit-fixes, own throwaway instance + Chrome DevTools MCP): re-verified the full flow against the fixed code — `DeleteWorktreeDialog` delete → real failure (no git remote → open-PR check error, honest error text rendered) → Force Delete → completed → hard reload with zero in-memory state → "Worktree Deleted" renders via the recovery path (money-shot re-confirmed after the M1/M2 changes) → navigated to a bogus id → honest "No Record of This Task" copy, no misleading phrasing. Also confirmed the M2 navigation-defer transition has no perceptible "closed but nothing happening" gap (see CodeRabbit section above).
- [ ] Screenshots are not embedded in this PR body (throwaway verification instances were torn down each round); available on request

Closes #1327

## Summary by CodeRabbit

* **New Features**
  * Worktree deletion now supports asynchronous processing with a server-generated job ID.
  * Deletion requests return immediately with a `202` response while progress is tracked in Jobs.
  * Added recovery for deletion status after reconnects or page reloads.
  * Added force-delete actions for recoverable deletion tasks.
  * Added “Worktree Deletion” labeling to the Jobs page.

* **Bug Fixes**
  * Immediate deletion failures now display an error without creating misleading task entries.
  * Improved handling of missing, stalled, and failed deletion records.


ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

Issue #1348, opened 2026-08-18
Account-scoped context (personal memory) reaches the embedded SDK agent outside settingSources' reach — second member of an unenumerated class


ms2sato:
## Observed

During the owner's hands-on pass on PR #1342 (SDK engine Phase 1), the embedded Claude agent opened a conversation by volunteering a piece of the owner's **personal memory** — an incident note about a UTM Ubuntu VM dying under a WindowServer watchdog kill. Nothing about that note relates to this repository.

The decisive detail: **that session had no repository context at all.** It was a Quick Session — `list_repositories` returned empty on the dev instance, the working directory was a `/tmp` scratch path, and the agent said so itself when asked about "this repository". So the memory cannot have arrived through anything project-derived. It is **account-scoped**.

## Why this is worth recording

`docs/design/embedded-agent-sdk-engine.md` §4.2 documents `settingSources: []` and, after the measurement recorded in #1343, states that it is the wall preventing filesystem-configured settings from loading. That statement is accurate **for the settings-sources class**. It is easy to read it as a broader claim than it supports.

Account-scoped injection is a different channel, and `settingSources: []` does not govern it. There are now **two observed members of that class**:

1. **claude.ai account connectors** (Drive/Calendar/Gmail-family tools) — appeared in the reported tool list under every probed arm, including `settingSources: []`. Already documented and accepted by the owner on TUI-parity grounds.
2. **Personal memory** — this Issue. Not documented anywhere.

The spec's accepted-risk material names the connectors and is silent on memory. A reader auditing "what reaches an embedded agent that we did not put there" would come away with an incomplete list.

## What this Issue asks for

Not a behavioural change, at least not yet:

- **Record the channel.** State in the spec that account-scoped context (connectors, memory, and possibly other members not yet enumerated) reaches the SDK engine independently of `settingSources`, and that `settingSources: []` is the wall for the settings-sources class specifically — not a general boundary.
- **Enumerate deliberately.** Connectors and memory were each found by accident, one during a probe and one during a hands-on pass. Whether anything else in this class exists has never been asked as a question. Worth one focused probe rather than waiting for the third accident.
[...]

PR #1349, opened 2026-08-18
feat: SDK Engine Phase 2 — context machinery (compaction off, usage polling, session-boundary handoff)

### LINKED ISSUE #1334: SDK Engine Phase 2: context machinery (compaction off, usage polling, session-boundary handoff)
Phase 2 of [#1324](https://github.com/ms2sato/agent-console/issues/1324). Spec: [`docs/design/embedded-agent-sdk-engine.md`](https://github.com/ms2sato/agent-console/blob/main/docs/design/embedded-agent-sdk-engine.md).

**Strictly sequential after #1333** — builds on its event plumbing. No UI novelty here.

## Scope

- **`settings.autoCompactEnabled: false`** — the owner's decision that our handoff stays active rather than deferring to SDK auto-compaction. Honored at creation and mid-session via `applyFlagSettings()` (premise PS1).
- **Usage polling via `getContextUsage()`** (`totalTokens` / `maxTokens` / `percentage`), surfaced through the mapped event vocabulary as context-usage events.
- 🔴 **Hazard H2 — the transport settle race.** Calling `getContextUsage()` immediately after a turn's `result` intermittently broke the transport (`ProcessTransport is not ready for writing`); ~300–500 ms of settle fixed it reliably. **Encode as retry-with-settle, not a bare sleep**, and re-verify on every SDK upgrade — this is an empirical workaround for observed 2.1.233 behaviour, not a documented contract.
- **Session-boundary handoff**: distill → terminate → seed the successor via `systemPrompt: { type: 'preset', preset: 'claude_code', append: <distillation> }` (premise PS3's mechanism, probe-verified with a real recall test).

## Two premise caveats that must not be built past

- **PS1 is verified at small scale only.** The probe could not afford ~1M tokens to force a real full-window compaction. The handoff E2E exercises the threshold path with a small window via **our own** trigger, which is the path production actually uses.
- **PS2's must-not-assume**: `settings.autoCompactWindow` had **no observable effect** in the probe (inconclusive, not proven dead). We never needed it — our threshold is our own — but **nobody may build on SDK-side window tuning without verifying it first**.

## Verification floor

- **Handoff E2E**: an SDK-engine worker crosses **our** threshold (small window via our own trigger), the session terminates, the successor is seeded, and a recall probe demonstrates continuity. This is PS3's probe elevated to a shipping-path test.
- Standard suite / typecheck / preflight; sibling tests per `test-trigger.md`.

## Not in this phase

`resume` — 🔴 **gated by PS4 everywhere** (typed-but-unprobed; no v1 code path touches it). Divergence UX (Phase 3). Idle eviction (Phase E).

---

## Acceptance Criteria — #1334 SDK Engine Phase 2 (context machinery)

Authored by the Architect. Spec: `docs/design/embedded-agent-sdk-engine.md` §4 (context rows), §5 (PS1/PS2/PS3/H2), Appendix A.2 (`context-usage` / `context-handoff` rows). Base: Phase 1 as merged (`6f68597e`).

### S0 — Containment: no wire/server/client changes

The event vocabulary already carries `context-usage`, `context-handoff`, and the `handoff` command; the server treats them engine-agnostically. Expected diff surface is `packages/embedded-agent/` (+ spec/tests) ONLY. Any change to shared schemas, server services, or client code is STOP-and-consult. ("No UI novelty" — Issue.)

### S1 — Usage polling (turn-gated, H2-encoded)

- Poll `getContextUsage()` exactly once per completed turn (after `result`) and once after each handoff attempt — per Appendix A's row. **No timer/interval polling anywhere in Phase 2** (this is what keeps H2 interaction confined to the settle path).
- Emit `context-usage` with `promptTokens` from `totalTokens`, `estimated: false` when the value is usable; a resolved-but-unusable value (missing fields) is skip-with-warn, NOT fatal.
- **H2 encoding**: retry-with-settle — named constants (settle delay, max attempts) with the empirical basis (~300–500 ms on the probed versions) in a comment that ALSO names the version premise and points at §5's re-verification obligation. Retry budget must comfortably exceed the observed settle window (at least 3 attempts spanning at least 2 s total). On exhaustion: `fatal`, per Appendix A's pre-committed mapping (a transport that never settles is a wedged session; the next turn would fail anyway). If the implementer's live runs contradict any of this (e.g. exhaustion observed on a healthy session), STOP-and-consult with the measurement — mechanics may move with data, and the Appendix gets a correction trail.
- **Re-verification ownership (floating-obligation fix)**: this PR adds a checklist item to Issue #1338 (the pending SDK bump): "re-run the H2 settle probe; re-verify PS1/PS2 observations" — so the obligation lives where the trigger lives, not in prose.

### S2 — PS1: accepted at small scale, WITH a runtime tripwire

- `settings.autoCompactEnabled: false` stays in the options battery (already Phase 1). PS1 remains verified-at-small-scale; this AC accepts that explicitly.
- **Compaction detector**: the engine remembers the previous poll's `totalTokens`; if a poll shows a MATERIAL drop (pin a named constant, e.g. more than 20%) with no intervening handoff, log a loud warn naming PS1 ("possible SDK-side compaction despite autoCompactEnabled:false"). This converts an unverifiable premise into a monitored one. Unit test with a faked descending sequence (polarity: drop produces the warn; normal growth does not).
- **PS2 must-not-assume, mechanically pinned**: a containment test asserting the string `autoCompactWindow` appears in NO production source under `packages/embedded-agent/src/` (grep-containment, same pattern as Phase 1's query()-single-call-site test).

### S3 — Session-boundary handoff (`handoff` command replaces the Phase 1 stub)

- Flow: distill (send the handoff/distillation prompt as a turn on the CURRENT session, via the existing `factories.loadHandoffPrompt` — single-writer, no second prompt source) -> emit `context-handoff { distillation }` (engine-authored, identical class — matches native) -> terminate the old SDK query -> construct the successor `query()`.
- **Successor options via the SAME builder**: extract the Phase 1 options construction into one function used by BOTH initial construction and reseed. Every Phase 1 pin (no `resume`, allowlist `tools`, `spawnClaudeCodeProcess`, `settingSources: []`, `autoCompactEnabled: false`, no apiKey) must hold on the reseed options — locked by reusing the builder, plus one test asserting the reseed-built options pass the same battery. A second hand-rolled options object is the drift vector this pin exists to kill.
- **Successor `systemPrompt.append` composition**: the original composed append (instructions + definition prompt, Phase 1's `composeSdkSystemPromptAppend` output) plus the distillation, order pinned (original first, distillation last), joined per the existing convention. PS3's probe seeded via append; this is that mechanism.
- `sdk-session-id`: the successor's `system:init` re-emits it (Phase 1's last-write-wins contract; persistence already handles replacement). No second `ready` — `ready` is construction-time only; the handoff brackets `state: active -> idle` exactly as the native engine's handoff does.
- Handoff during an active turn: match the native engine's observable contract (the implementer verifies what native does and states it with an anchor in the PR body; do not invent new semantics).

### S4 — 🔴 Reseed vs Phase 1's clean-stream-end guard (named interaction)

Terminating the old query makes its `consumeLoop` end WITHOUT throwing. Phase 1 maps that to `handleFatal('SDK message stream ended unexpectedly')` unless `this.dead` — but during handoff the engine is NOT dead and must not become so. **Without this, every handoff kills the engine with a spurious fatal.**

Required: a per-query generation / expected-close mechanism (e.g. each consumer loop captures its query generation; a loop whose query has been superseded or intentionally closed exits silently).

Tests: (a) handoff completes -> NO fatal emitted, engine alive, a post-handoff turn works; (b) the ORIGINAL guard still fires for a genuinely unexpected stream end on the CURRENT query (invariant preservation — do not fix (a) by deleting the guard).

### V — Verification floor

- **Handoff E2E (shipping path, dev instance)**: seed a conversation containing a **conversation-local nonce** (an arbitrary token stated by the user mid-conversation, existing NOWHERE on disk or in any tool-reachable surface — this is Phase 1's lesson generalized: the recalled fact must be tool-unreachable by construction; a nonce achieves what a behavior marker achieves, more simply). Trigger handoff (the manual `handoff` command — Phase A's contract; auto-trigger is Phase B, out of scope). Assert BOTH: (i) mechanism — the emitted `context-handoff.distillation` actually contains the nonce (or an explicit reference to it); (ii) goal — the successor answers a recall question with the nonce.
- **🔴 Account-memory confound (#1348)**: the executing account's memory is now KNOWN to reach embedded agents. A successor could "recall" the nonce via account memory instead of our distillation — a false positive on (ii). Required isolation: run the E2E with a throwaway `CLAUDE_CONFIG_DIR` (the settingSources-probe technique) so no account memory exists; (i) additionally proves the mechanism regardless. Document the isolation in the PR body.
- H2 unit tests: fake `getContextUsage` failing with the transport error N times then succeeding -> retried and emitted; failing beyond budget -> fatal.
- Standard suite / typecheck / preflight; sibling tests per `test-trigger.md`.

### Docs (same PR)

- Appendix A: `context-usage` and `context-handoff` rows move from "(Phase 2)" annotations to implemented, with any measurement-driven corrections carried as correction trails.
- §4.1: one-line cross-reference to #1348 (account-scoped memory as the second member of the account-capability class; disposition tracked there, not here).
- #1338: the S1 re-verification checklist item.

### Non-goals

`resume` (PS4-gated everywhere — the reseed uses a FRESH session id, never `resume`); auto-trigger on threshold (Phase B); divergence UX (Phase 3); idle eviction (Phase E); any `autoCompactWindow` usage (PS2); UI changes; wire-vocabulary changes.

ms2sato:
## Summary

Implements SDK Engine Phase 2 (Issue #1334) on top of Phase 1 (#1333): compaction stays off (with a runtime tripwire), context-usage polling with the H2 settle-retry, and session-boundary handoff (distill → terminate → reseed).

Scope per S0: `packages/embedded-agent/` + spec/tests only. No wire/server/client changes — the event vocabulary (`context-usage`, `context-handoff`, the `handoff` command) was already carried by Phase 1's shared types.

## What changed

- **S1 — usage polling**: `SdkEngine` polls `getContextUsage()` exactly once per completed turn (after `result`) and once per handoff attempt — never on a timer. H2's transport-settle race is encoded as retry-with-settle (`CONTEXT_USAGE_SETTLE_DELAY_MS=500`, `CONTEXT_USAGE_MAX_ATTEMPTS=6`; 5 gaps = 2500ms, comfortably over the AC's 2s/3-attempt floor and the observed 300–500ms window). Exhaustion is `fatal`. A resolved-but-unusable response (no `totalTokens`) is skip-with-warn, not fatal.
- **S2 — PS1 tripwire**: the engine remembers the previous poll's `totalTokens`; a >20% drop with no intervening handoff logs a loud warning naming PS1. A handoff resets the baseline (an intentional drop is not a violation). PS2's must-not-assume is pinned by a grep-containment test: `autoCompactWindow` appears nowhere under `packages/embedded-agent/src/`.
- **S3 — session-boundary handoff**: `handoff()` runs the distillation prompt (`factories.loadHandoffPrompt`, the same single-writer prompt source as the native engine) as a turn on the current session, with its per-turn wire events suppressed (no stray `assistant-message`/`tool-call` for the internal distillation turn — mirrors the native engine's `emitDeltas: false` intent). On success it emits `context-handoff{distillation}`, closes the old query, and reseeds a successor via `buildOptions` — the SAME options builder used at construction, so every Phase 1 pin (no `resume`, allowlist `tools`, `spawnClaudeCodeProcess`, `settingSources: []`, `autoCompactEnabled: false`, no `apiKey`) holds on the reseed too. `systemPrompt.append` is the ORIGINAL composed append (fixed at construction) plus the latest distillation, original first, joined with the existing `\n\n` convention. No second `ready` is emitted.
- **S4 — reseed vs. the clean-stream-end guard**: terminating the old query to reseed makes its `consumeLoop` end without throwing — the exact shape Phase 1's guard treats as fatal. Fixed with a per-query generation counter (`queryGeneration`): each `consumeLoop` captures the generation of the query it was started for; a mismatch at stream-end means a newer query has since superseded it (expected), not the current query dying unexpectedly. The `this.dead` check that guards `dispose()`'s deliberate close is untouched — both guards now coexist for their respective cases.
- **Handoff during an active turn**: verified this is already handled identically for both engines at `main.ts`'s dispatch layer — `handoff` is rejected with `Ignoring handoff received while a turn is active` by the SAME `turnActive` gate `user-message` is subject to, engine-agnostic. No engine-specific code was needed; added a `main.test.ts` test (`runLoop — claude-sdk engine: handoff dispatch gate`) as the anchor.
- **Docs**: Appendix A's `context-usage`/`context-handoff` rows move from "(Phase 2)" to implemented, with a correction trail recording the settled implementation details (H2 constants, PS1 tripwire, S4's generation guard). §4.1 gains a one-line cross-reference to #1348 (account-scoped memory as the second member of the account-capability class). Issue #1338's re-verification checklist gained the S1 re-verification item (H2 settle probe + PS1/PS2 re-verify).

## S4 both-directions test result

Both required directions are covered in `sdk-engine.test.ts`:

- **(a) Handoff produces no spurious fatal, engine stays alive**: `SdkEngine — session-boundary handoff (S3) > distills, emits context-handoff, terminates the old query, and reseeds a successor -- no fatal, no second ready, wire-suppressed distillation turn` — asserts zero `fatal` events across the whole flow, then drives a real post-handoff `runTurn` to completion on the successor session.
- **(a, isolated mechanism)** `SdkEngine — S4: reseed vs the clean-stream-end guard > the OLD query's stream ending as a DIRECT RESULT of reseed's close() does not emit a spurious fatal (generation guard)` — a dedicated test where the old query's stream only ends once `reseed()`'s `close()` resolves it (mirroring the real close-triggers-stream-end sequence), asserting zero fatal events.
- **(b) Invariant preservation — the original guard still fires**: unchanged, pre-existing `SdkEngine — clean stream end (unexpected vs deliberate)` describe block (no handoff occurs there, so `queryGeneration` never advances) still passes: a genuinely unexpected clean stream end on the current, non-superseded query emits exactly one `fatal` with the same message as before.

Fix was **additive** (a generation-mismatch early-return ahead of the existing fatal path), not a weakening of the guard — the `this.dead` check `dispose()` relies on is untouched.

## Handoff E2E — isolation evidence and result

Ran against the REAL production shipping path: `packages/embedded-agent/src/main.ts`'s `runLoop`, the real `SdkEngine`, and the real `@anthropic-ai/claude-agent-sdk` `query()` (a real `claude` subprocess was spawned via `spawnClaudeCodeProcess`) — no fakes/mocks. This is the correct shipping-path boundary per S0: the server/wire layer is engine-agnostic and untouched by this phase, so the subprocess entry point IS the shipping path for what Phase 2 changes.

**Isolation**: `CLAUDE_CONFIG_DIR` was overridden (via the spawned process's inherited env, since `Options.env` is left unset in production) to a throwaway directory containing ONLY a copy of `.credentials.json` (auth, so the real `claude` binary could authenticate) — no `projects/`, no `history.jsonl`, no memory files of any kind. This eliminates the #1348 account-memory confound structurally, not by assertion.

**Run** (ad hoc script, not committed — matches how Phase 1's own probes were handled): seeded a conversation-local nonce token (`Remember this exact token... MUST be preserved verbatim in any future summary or distillation`), triggered the manual `handoff` command, then asked the successor to recall it.

- **Mechanism** (the nonce is actually in the emitted distillation): `context-handoff.distillation` contained the literal nonce token — confirmed.
- **Goal** (the successor recalls it): the successor's very next turn replied with exactly the nonce token — confirmed.
- Successor's `sdk-session-id` differed from the original (fresh session, no `resume`) — confirmed.
- Zero `fatal` events across the whole run; `runLoop` exited 0 on `shutdown`.

One earlier iteration (before adding an explicit verbatim-preservation instruction) had the model summarize the token away rather than quote it in its distillation. The engine functions correctly — what remains unverified is distillation fidelity under uncoached usage (no explicit instruction to preserve a value verbatim), tracked in #1350.

## H2 measurement

No live run in this PR hit the retry-with-settle path against the real SDK (the dev-instance E2E's `getContextUsage()` calls all succeeded on the first attempt) — H2's constants (500ms settle, 6 attempts) are carried forward from the design doc's empirical figures as an encoded workaround, unit-tested for both directions (retries-then-succeeds, exhausts-then-fatal) via a DI'd `getContextUsage`/`sleep`, per the AC's explicit allowance ("If the implementer's live runs contradict any of this... STOP-and-consult with the measurement"). Nothing here contradicted the AC's stated numbers, so no correction trail was needed.

## Verification

- `bun run typecheck` — clean.
- `bun run test` — full suite green (see pasted output below).
- `node .claude/skills/orchestrator/preflight-check.js` — clean (coverage, rule/skill duplication, language check, blame-shift check all pass).
- `bun run check:lang` — clean.

```
$ bun run test 2>&1 | tail -100
...
[server tests, 4038 pass / 0 fail]
[embedded-agent tests, 373 pass / 0 fail]
[client tests, 2200 pass / 0 fail]
[shared tests, 634 pass / 0 fail]
[integration tests, 75 + 3 pass / 0 fail]
[scripts/hooks tests, 538 pass / 0 fail]
TEST_EXIT: 0
```

## Test plan

- [x] `bun run typecheck` clean
- [x] `bun run test` (full workspace) green, exit 0
- [x] `node .claude/skills/orchestrator/preflight-check.js` clean
- [x] `bun run check:lang` clean
- [x] New sibling tests: `packages/embedded-agent/src/__tests__/sdk-engine.test.ts` (S1/S2/S3/S4 describe blocks), `packages/embedded-agent/src/__tests__/main.test.ts` (handoff dispatch gate + `loadHandoffPrompt` wiring)
- [x] S4 both-directions test (handoff → no fatal; unrelated unexpected end → original fatal still fires)
- [x] Handoff E2E on the real shipping path with an isolated `CLAUDE_CONFIG_DIR`, mechanism + goal both confirmed

## Non-goals (unchanged from the Issue)

`resume` (reseed uses a fresh session id, never `resume`); auto-trigger on threshold (Phase B); divergence UX (Phase 3); idle eviction (Phase E); any `autoCompactWindow` usage (PS2, containment-tested); UI changes; wire-vocabulary changes.

## Architect consultations

None needed during implementation — no AC/scope/design-decision ambiguity surfaced that required a live consultation; all decisions (H2 constants, distillation wire-suppression, generation-guard mechanism, "original append" semantics for repeated handoffs) followed directly from the AC's explicit text.

Closes #1334

## Note on CodeRabbit

Local CodeRabbit CLI hit the headless-worktree auth failure (shape B: hung on `awaiting_browser_auth` with no self-resolution — a structural limitation of this delegated session, not a rate-limit). Relying on the GitHub-side CodeRabbit bot review for this PR.

## Post-push self-review fixes (2nd commit)

Two hang/stuck-state gaps found in manual self-review (CodeRabbit CLI unavailable, see the note above), each with a bug-polarity regression test:

- `handleFatal` only settled the normal-turn deferred, not a pending distillation -- a transport crash mid-distillation would leave `handoff()`'s await hanging forever. Fixed by also settling `distillationDeferred` (with a failure outcome) on fatal.
- `reseed()`'s successor `query()` construction had no rescue for a synchronous throw -- the engine would silently keep pointing at the already-closed old query with nothing consuming the new queue. Fixed by constructing the successor query before mutating `this.queue`/`this.query`, and routing a construction failure through `handleFatal`.

## Merge disposition note (2026-08-18)

**CR state**: local CLI hit the headless-worktree auth failure (shape B) — no review obtained. GitHub-side bot, after re-triggering on the 2nd-commit push, returned `Review rate limited` (commit status `state: success`, `description: "Review rate limited"` at 2026-08-18T04:44:17Z) — no genuine walkthrough for this head. `reviewDecision` is empty. inline comments: 0.

Both CodeRabbit channels are unavailable for the current head (`4e355488`), per the `coderabbit-ops` skill's "headless auth-fail AND GitHub bot rate-limited" disposition. All other CI checks (test, typecheck via preflight, both CodeQL analyzers, comment-blame-shift-lint, language-lint, mock-module-lint, check-structure) are green. Deferring the CodeRabbit-verdict gap and the merge decision itself to the Orchestrator/Architect per PR Merge Authority — not merging this PR myself (per the task's explicit instruction).

## Summary by CodeRabbit

- **New Features**
  - Added live conversation handoffs that preserve relevant context in a successor session.
  - Added context-usage monitoring with automatic retries and handling for context compaction.
  - Added configurable handoff prompt loading for more consistent session transitions.

- **Bug Fixes**
  - Improved session recovery and termination handling to prevent hangs during handoffs.
  - Prevented duplicate or unexpected client-facing events while conversations are being transferred.
  - Improved handling of failed, incomplete, or invalid handoff operations.


ms2sato:
@coderabbitai review

Issue #1361, opened 2026-08-18
notifications: add artifact-created invalidation-hint broadcast


ms2sato:
## Context

The notification center client (#1354) composes its feed from a fetch-time read model per `docs/design/notification-center.md` (Invariant N1). WebSocket broadcasts for covered kinds are invalidation hints ONLY -- they trigger a refetch, never render their own payload.

Verified fact at `276eeb03` (server Phase 1 merge): `AppServerMessage` carries `worktree-deletion-completed` / `worktree-deletion-failed` but has NO `artifact-created` member. Artifact creation broadcasts nothing today.

## Consequence (accepted for v1, tracked here as the fast-follow)

An artifact created while a user is already looking at the notifications bell/panel on this tab is NOT reflected until the next refetch trigger (panel open, app-WS reconnect, or window focus). It is not invisible -- it appears on the next trigger -- but it is not instant, unlike the two worktree-deletion kinds which already have broadcast-driven invalidation.

## Proposed work

Add a server-side `artifact-created` invalidation-hint broadcast:

- A new `AppServerMessage` member (e.g. `artifact-created`) emitted when an artifact is created, following the same shape/precedent as `worktree-deletion-completed` / `-failed`.
- The full `pre-pr-completeness.md` Q10 checklist applies to this addition (shared-type schema parity, integration test at the wire boundary, etc.) since it is a new member of a shared wire type.
- **N1 still applies to this new broadcast**: its payload must never be rendered client-side. The client handler for it must do exactly `queryClient.invalidateQueries(notificationCenterKeys.root())` and nothing else, mirroring the existing `onWorktreeDeletionCompleted` / `onWorktreeDeletionFailed` handlers in `packages/client/src/components/notifications/NotificationBell.tsx`.

## References

- `docs/design/notification-center.md` (normative spec; N1/N2, §8 wire shape)
- Notification Center Phase 2 client PR (Issue #1354) -- this Issue is filed as its Ruling 1 required deferral-with-address.

ms2sato:
Closing: shipped in PR #1523 (artifact-created AppServerMessage, broadcastToApp, NotificationBell onArtifactCreated). The PR did not carry a Closes keyword. Triage 2026-09-03.

Issue #1335, opened 2026-08-17
SDK Engine Phase 3: divergence UX + surface completion


ms2sato:
Phase 3 of [#1324](https://github.com/ms2sato/agent-console/issues/1324). Spec: [`docs/design/embedded-agent-sdk-engine.md`](https://github.com/ms2sato/agent-console/blob/main/docs/design/embedded-agent-sdk-engine.md).

**Strictly sequential after #1334.**

## Scope

- 🔴 **Make the restore divergence visible.** In v1, server-restart revival gives an SDK worker a **fresh session** — the accepted-divergence row's consequence. **Ship it as UX, not as a surprise**: the UI says so. The owner accepted the divergence (Q2) on the understanding that its consequence would be stated at AC time rather than discovered in use; this phase is where that promise is kept.
- **Picker / badge presentation** for the Claude-SDK definition, through the existing uniform-picker-with-kind-badges doctrine. This is where the owner's user-facing choice — *Claude Code as a TUI worker, or Claude as an embedded SDK worker* — becomes visible. **No new concept**: the engine is not user-facing configuration (spec §3.1); the user picks a definition, and that determines the engine.
- **Docs.**
- **Remaining per-surface E2E** and any approximated-event refinements the mapping table deferred.

## Verification floor

- Browser QA with **true-path screenshots** — the feature-visible state, legible to a non-technical reader, per `workflow.md`'s gated/conditional-UI requirement. The divergence notice is exactly the kind of UI that must be shown in its visible state rather than its default one.
- Standard suite / typecheck / preflight; client-side sibling tests per `test-trigger.md`.

## Not in this phase

Idle eviction (Phase E, post-v1, gated on the PS4 probe).

---

## Acceptance Criteria

Sequential after [#1351](https://github.com/ms2sato/agent-console/issues/1351)'s merge (shared file: `EmbeddedAgentWorkerView.tsx`; ownership boundary below).

### S1 — Restore-divergence notice (the kept promise)

- **Contract**: after a server restart revives an SDK-engine embedded worker that HAD prior conversation (persisted transcript exists), the worker view displays a clearly visible notice: the conversation context was NOT carried across the restart (fresh session; the transcript above is history, not live context). Native-engine workers show no such notice (their restore reconstructs — polarity arm).
[...]

PR #1362, opened 2026-08-18
feat: notification center client (bell, badge, panel)

### LINKED ISSUE #1354: Notification center Phase 2: client (bell, badge, list panel)
Phase 2 of the notification center. Spec: [`docs/design/notification-center.md`](https://github.com/ms2sato/agent-console/blob/main/docs/design/notification-center.md) (merged in #1352 — normative).

**Client only, sequential after Phase 1** (needs its routes and wire shape).

## Scope

- **Bell in the header**, with a badge showing the count of items newer than the user's cursor (§5, §8).
- **List panel**: newest first, each item a pointer + summary that deep-links to its domain object.
- **Opening the panel advances the cursor** (`PUT /api/notifications/seen`) — the badge clears as a consequence of that server round-trip, not as local state.
- **WebSocket events for the covered kinds are invalidation hints only (N1)** — they trigger a refetch; their payloads are never rendered into the list. A phantom broadcast must degrade to "refetch found nothing", visibly nothing, never a phantom row.
- Empty-state copy; bell placement relative to the existing header nav (§11 leaves both to AC time).

## Verification floor

- **Browser QA with true-path screenshots** per `workflow.md`'s gated-UI rule: a real unread item producing a real badge, the panel open showing it, and the badge cleared after the seen-advance. The false path (empty state) alone does not satisfy this.
- Client sibling tests per `test-trigger.md`, including the N1 polarity: a broadcast whose backing row is absent results in no rendered item.

## Not in this phase

Additional sources; the obligations / action-required lane (§6); banner-class events ([#1344](https://github.com/ms2sato/agent-console/issues/1344)); Slack-consumer unification; a global team feed.

---

## Acceptance Criteria

Spec: `docs/design/notification-center.md` (normative; N1/N2 binding). Server surface: Phase 1 as merged (`276eeb03`) — `GET /api/notifications` / `PUT /api/notifications/seen`, wire schemas in `shared/schemas/notification-item.ts`.

### S0 — Containment: client-only phase

Expected diff surface: `packages/client/` + docs. **No server, shared-schema, or wire changes** — the one known gap (below) is resolved by deferral-with-address, not by adding wire surface here. Any wire need discovered mid-flight = STOP-and-consult (Q10 re-fires).

### Ruling 1 — `artifact-created` has NO broadcast (verified fact, not a branch)

Verified at `276eeb03`: `AppServerMessage` carries `worktree-deletion-completed` / `-failed` but no artifact member; artifact creation broadcasts nothing. Disposition:

- **v1 refetch triggers** (all of): panel open; app-WS reconnect; the two EXISTING deletion broadcasts as invalidation hints (N1); TanStack Query's `refetchOnWindowFocus` (left enabled for the badge query — returning to the tab is exactly the "wasn't watching" moment this feature serves). **No interval polling** (feed-shaped principle; chatter for no benefit).
- Consequence, stated honestly in the PR body and in spec §8's client line (one-line annotation): an artifact created while the user is ALREADY looking at this tab appears on the next trigger, not instantly.
- **Deferral address (required before merge)**: file a follow-up Issue for a server-side `artifact-created` invalidation-hint broadcast (a new `AppServerMessage` member; the full Q10 set applies THERE). Note in it that the hint's payload will still never be rendered (N1) — it exists only to invalidate.

### Ruling 2 — badge zeroing is a server round trip, structurally

- **The badge renders ONLY the latest `GET` response's `unreadCount`.** No local decrement, no optimistic zero, no client-side unread math (an optimistic zero lies when the PUT fails; client math re-derives what the server already computed pre-cap).
- Open-panel flow: fetch list → render → `PUT /seen` with **the newest fetched item's `occurredAt`** (never client `Date.now()` — pins to R4.1's tie semantics and avoids the future-timestamp 400 on clock skew) → invalidate → refetched `unreadCount` (now 0) clears the badge. Empty feed on open → NO PUT (nothing was seen; cursor unchanged).
- Test: badge shows the pre-open count until the post-PUT refetch resolves; a failing PUT leaves the badge non-zero (polarity for the no-optimistic-zero pin).

### Ruling 3 — N1 client-side, pinned by a phantom-broadcast polarity test

Broadcast handlers do exactly one thing: `queryClient.invalidateQueries(notification keys)`. **The broadcast payload must be structurally unreachable from render** (the handler receives it and discards it — no field of it is stored or passed onward). Required test: fire a deletion-completed WS event carrying a fully-populated payload while the mocked `GET` returns an EMPTY feed → assert (a) exactly one refetch was triggered, (b) NOTHING from the broadcast payload appears in the DOM (the #1330 failure mode, structurally unrepresentable client-side too).

### Ruling 4 — bell placement (§11 resolution)

- Desktop: in the header bar's right-side control cluster (with the existing NavLinks), to the LEFT of Settings. Mobile: in `MobileHeaderControls`' always-visible area, **NOT inside the collapsed menu** — a badge hidden behind a menu tap defeats its purpose. Pin: **the unread badge is visible without opening any menu, on both breakpoints**.
- Panel: an anchored dropdown/popover from the bell (no new SPA route, no `routeTree` change). Rows render kind-appropriate title + relative time + deep link (the `link` field navigates; a full-document anchor is NOT required here — these are SPA routes except `/artifacts/{id}`, which must remain a plain `<a>` per #1340's jail rule — reuse the existing View-link discipline for that one kind).

### S — Implementation

- **S1**: `lib/api.ts` — `fetchNotifications()` parsing via `NotificationsResponseSchema`, `markNotificationsSeen()` via `NotificationsSeenResponseSchema` (the #1340 `fetchArtifacts` precedent: parse at the wire, no blind cast). `lib/query-keys.ts` gains notification keys.
- **S2**: bell + badge + panel components under `components/notifications/` (or header-adjacent per existing conventions), wired per Rulings 2-4.
- **S3**: WS hint wiring via the existing app-websocket subscription (Ruling 3).

### T — Tests

Ruling 2/3 polarity tests as specified; empty-state panel; `unreadCount > items.length` renders sanely (51+ case: badge shows the server count, list shows 50); a `worktree-deletion-finished` row with `outcome: 'failed'` visually distinguishable from completed; sibling tests per `test-trigger.md` for every new component/hook.

### V — Verification

Browser QA true-path screenshots (`workflow.md` gated-UI rule): (a) badge non-zero with a REAL artifact + a REAL finished deletion (created via real flows on the dev instance); (b) panel open showing both kinds; (c) badge cleared after open (the server-round-trip zeroing — i.e. a reload still shows 0); (d) deep links landing. Full suite / typecheck / preflight.

### Non-goals

Everything spec §10 lists, plus: interval polling; an `artifact-created` broadcast (deferred with an address, Ruling 1); the obligations lane; per-item read state; toasts / OS notifications.

### Collision map (parallel dispatch with #1335 / SDK Phase 3)

- **#1354 owns**: `routes/__root.tsx` (header area), `components/header/**` (`NavLinks.tsx` / `MobileHeaderControls.tsx` / `MobileNavMenu.tsx`), new `components/notifications/**`, `lib/api.ts` / `lib/query-keys.ts` (appended lines), the app-WS subscription additions around `hooks/useAppWs`.
- **#1335 owns**: `EmbeddedAgentWorkerView.tsx` and its subtree, the agent-picker surfaces (`components/worktrees/CreateWorktreeForm.tsx` / session-creation dialogs / `routes/agents/**`), docs.
- **Zero intersection is the expectation.** If #1335 turns out to need header/nav changes, consult (and vice versa). `lib/api.ts` / `query-keys.ts` are the only shared files both may append to — trivial add-line merges.

ms2sato:
## Summary

Client half of the notification center (#1354, Phase 2 of `docs/design/notification-center.md`). Adds a bell + server-truth unread badge to the header, an anchored dropdown panel listing artifact/deletion notifications, and app-WS invalidation-hint wiring.

- **S1**: `fetchNotifications()` / `markNotificationsSeen()` in `packages/client/src/lib/api.ts` (parsed through `NotificationsResponseSchema` / `NotificationsSeenResponseSchema` at the wire boundary, no blind cast). `notificationCenterKeys` added to `packages/client/src/lib/query-keys.ts` (distinct from the pre-existing, unrelated `notificationKeys` which is the outward-Slack status key).
- **S2**: `packages/client/src/components/notifications/` — `NotificationBell.tsx` (trigger + badge + orchestration), `NotificationPanel.tsx` (anchored dropdown body), `NotificationItemRow.tsx` (kind-appropriate row).
- **S3**: WS hint wiring via `useAppWsEvent`'s existing `onWorktreeDeletionCompleted` / `onWorktreeDeletionFailed` (previously unconsumed by any client code), plus an app-WS reconnect (false→true) invalidation trigger. `refetchOnWindowFocus` left at its TanStack Query default (enabled).

### Ruling 2 (badge is a server round trip, not optimistic)

The badge renders only the latest `GET` response's `unreadCount` — no local decrement, no optimistic zero. Opening the panel: `refetch()` → if the refetch succeeded and items exist, `PUT /seen` with the newest fetched item's `occurredAt` (never `Date.now()`) → on success, invalidate → the refetched `unreadCount` clears the badge. An empty feed sends no PUT. **A failed refetch also sends no PUT** — see "Round-2 CodeRabbit findings" below; TanStack Query retains stale data on a failed refetch, so this guard is required, not redundant with the empty-feed check. On error, the badge keeps showing the last successful `unreadCount` (TanStack's stale-data-on-error default) — there is no dedicated error-badge state.

### Ruling 3 / N1 (WS broadcasts are invalidation hints only)

Both WS handlers in `NotificationBell.tsx` take no parameter — the broadcast payload is structurally unreachable, not merely unused by convention:

```ts
onWorktreeDeletionCompleted: () => {
  queryClient.invalidateQueries({ queryKey: notificationCenterKeys.root() });
},
```

### Ruling 4 (placement)

- Desktop: `NotificationBell` inserted in `routes/__root.tsx`'s nav, between `RepositoriesNavLink` and `SettingsNavLink`, gated `{!isMobile && <NotificationBell />}` (see "Round-2 CodeRabbit findings" — the desktop nav is CSS-hidden-only on mobile, not unmounted, so without this gate it double-mounts alongside `MobileHeaderControls`' own instance).
- Mobile: `NotificationBell` inserted in `MobileHeaderControls.tsx`'s always-visible `md:hidden` control cluster (sibling to the sessions/hamburger buttons), **not** inside the collapsed `MobileNavMenu`.
- Panel: hand-rolled anchored dropdown (fixed-inset overlay + Escape-key listener), mirroring the existing non-Radix `MobileNavMenu` pattern already in this codebase. No new dependency was added — `@radix-ui/react-popover` isn't installed and the existing hand-rolled pattern already solves outside-click + Escape without it.

## Collision map confirmation

No files under #1335's territory (`EmbeddedAgentWorkerView.tsx`, agent-picker surfaces, `routes/agents/**`) were touched. `lib/api.ts` and `lib/query-keys.ts` received pure appended lines per the shared-file agreement.

## Deferred (Ruling 1 required address)

`artifact-created` has no WebSocket broadcast today (verified fact at `276eeb03`). An artifact created while the tab is already open on the bell/panel is picked up on the next refetch trigger (panel open, app-WS reconnect, or window focus) rather than instantly — not invisible, just not instant. Follow-up filed: **#1361** (server-side `artifact-created` invalidation-hint broadcast; N1 still applies there). `docs/design/notification-center.md` §8's client line carries a one-line honest annotation pointing at this.

## Round-2 CodeRabbit findings (fixed, commits `deeb033d`, `780affa0`, `d374a1ce`)

First review pass (3 Minor):
- Tautological test assertion (`NotificationBell.test.tsx`, "reconnect" test compared a stored value to itself with no time passing) — fixed by awaiting a real settle window before the assertion.
- Duplicate `NotificationBell` mount on mobile (desktop `<nav>` is CSS-hidden-only, not unmounted, so it stayed mounted alongside `MobileHeaderControls`' own instance, causing duplicate WS subscriptions) — fixed by gating the desktop instance with `{!isMobile && <NotificationBell />}`.
- Missing error state on `NotificationPanel` (a failed `GET` rendered as "No notifications yet", indistinguishable from a genuinely empty feed) — **architect-ruled to fix in this PR** (not deferred): added an `isError` prop + "Failed to load notifications" + Retry button, matching the existing `ArtifactsPage` error-state convention (`routes/artifacts/index.tsx:88-96`). Badge is explicitly unaffected (see Ruling 2 above).

Second review pass (1 Major): **`openAndMarkSeen` could mark stale/cached items as seen after a failed refetch.** TanStack Query v5 retains the previous successful `data` (not `undefined`) on a failed refetch and sets `isError: true`. The original code read `result.data?.items` without checking `isError`, so a failed open-triggered refetch would advance the server-side read cursor using data that was never freshly confirmed — while the UI simultaneously showed the new error state, meaning notifications the user had not actually seen would be marked seen. Fixed by checking `result.isError` before calling `seenMutation.mutate`. Regression test seeds real prior data, forces the refetch to fail, and asserts (a) the failure genuinely reproduced (error copy visible, stale badge still showing the pre-failure count — ruling out a vacuous pass) and (b) no `PUT /notifications/seen` fired. Polarity confirmed both directions (fails without the guard, passes with it). Grepped all other `refetch()` call sites in `packages/client/src/` for the same shape — none found; every other site is a plain discard-the-return-value retry button.

## Acceptance-check follow-ups (fixed)

- Added test coverage for the panel's three dismiss paths (Escape key, outside-click overlay, unmount cleanup of the `keydown` listener) — previously untested. Cleanup test spies on `document.addEventListener`/`removeEventListener`, captures the actual registered handler reference, and asserts `removeEventListener` is called with that same reference (catches the "new arrow function every render, listener never actually removed" class of React bug, not just "no exception after unmount"). All three verified with real polarity: broke each mechanism in production code one at a time, confirmed the matching test failed, restored, confirmed `git diff` was clean.
- Updated `docs/glossary.md`'s **Notification Center** entry, which still opened with "Design only ... no implementation yet" despite both phases (server #1357, client #1354/#1362) being merged. Also added one sentence noting `unreadCount` is computed pre-cap and can exceed `items.length`, so a future reader does not derive the badge from list length.

## Browser QA (true-path screenshots, real dev instance)

All states driven by real flows against a disposable single-user dev instance (`bun run dev`, confirmed to be this session's own process — PID/cwd cross-checked against `/proc`, not a sibling delegate's server sharing the port). No fixtures.

- **Real artifact**: created via the real `create_html_artifact` MCP tool (same shipping path a terminal/embedded agent uses), attributed to an existing dev-instance session.
- **Real completed deletions**: two throwaway worktrees created via the UI's "+ Worktree" flow, deleted via the real "Delete" button — both produced genuine `worktree-deletion-finished` / `outcome: completed` notifications through the real job queue.
- **Real failed deletion — with a caveat, transparently recorded here**: first attempted to induce a genuine failure via `git worktree lock <path>` on a throwaway worktree, expecting `git worktree remove` to fail naturally. **It did not** — the app's existing force-fallback recovery logic (`WorktreeService.removeWorktree`'s catch-and-recover path, per `.claude/rules/elevation-helpers.md`'s "canonical example" PRs) successfully cleaned up the locked worktree anyway, producing `outcome: completed`. This is a genuine observation, not a QA footnote: **it means the `failed` outcome is reachable only by a narrower set of real-world conditions than "the git command didn't complete cleanly"** — the app's own resilience absorbs the common cases. Given the difficulty reproducing it naturally within a reasonable QA window, switched to the temporary dev-stub technique `workflow.md` explicitly permits for this situation: added a one-line, narrowly-gated early-return (`if (worktreePath.includes('wt-006-bq3j')) return { success: false, ... }`) to `packages/server/src/services/worktree-deletion-service.ts`'s `deleteWorktree`, matched to one specific throwaway worktree's actual generated path, drove a real deletion attempt against it (real job queue, real broadcast, real `GET` response), captured the screenshot, then reverted the stub. Verified the revert three ways before pushing: `git status --porcelain` (empty), `git diff` (empty), and — the final check — `gh pr diff 1362 | grep` for the stub's literal strings (`QA: forced failure`, the worktree path, `TEMPORARY QA STUB`) against the actual pushed PR diff: **zero matches**. The worktree itself was then deleted for real (stub already reverted) and confirmed gone from `git worktree list`.
- **Badge → panel → cleared**: badge showed a real non-zero unread count; opening the panel showed all three rendered kinds together (`artifact-created`, `worktree-deletion-finished` / completed, `worktree-deletion-finished` / failed — the failed row visually distinct with red text + left border); a subsequent full page reload showed the badge at 0, confirming the seen-cursor advance is a real server round trip, not client state.
- **Deep links**: the worktree-deletion-task link landed on the real task page (SPA `<Link>`); the artifact link opened the real artifact in a new tab via a plain `<a target="_blank">` (#1340 jail rule).
- **Mobile breakpoint**: resized to a mobile viewport; the bell renders in the always-visible control cluster (not the collapsed hamburger menu) with the badge visible without opening any menu, and the panel opens correctly there too.
- **Cleanup**: all QA-created data removed after screenshots — both throwaway worktrees deleted (`git worktree list` confirms clean), the QA artifact deleted via `DELETE /api/artifacts/:id` (confirmed `200`). Nothing QA-related left behind.

## Test plan

- [x] `bun run typecheck` — clean across all packages.
- [x] `bun run test` — full suite green (server 4089 pass / 1 skip, embedded-agent 375, client 2239, shared 659, integration 85, scripts/hooks 538 — all 0 fail). Re-verified after every fix round; final run's `TEST_EXIT: 0`.
- [x] Ruling 2 polarity: verified by temporarily injecting a naive optimistic-zero probe and confirming both Ruling-2 tests fail against it, then reverting. Extended round-2: `openAndMarkSeen`'s `isError` guard also polarity-confirmed (see above).
- [x] Ruling 3 / N1 phantom-broadcast polarity: verified by temporarily leaking the WS payload's `taskId` into render and confirming the phantom-broadcast test fails against it, then reverting.
- [x] Empty-state panel (with negative arm: error copy absent); error state panel (with negative arm: empty-state copy absent) — both directions required per architect ruling.
- [x] `unreadCount > items.length` (51 unread / 50 items) shows the server count on the badge and exactly 50 rows in the panel.
- [x] `worktree-deletion-finished` with `outcome: 'failed'` renders visually distinguishable (red text + left border) from `outcome: 'completed'` — confirmed both in unit tests and in the real Browser QA screenshot.
- [x] `artifact-created` rows use a plain full-document `<a target="_blank">` per #1340's jail rule; `worktree-deletion-finished` rows use a typed TanStack `<Link to="/worktree-deletion-tasks/$taskId" params={{ taskId: item.id }}>`.
- [x] Panel dismiss paths (Escape, outside click, unmount listener cleanup) — added in acceptance-check follow-up, all polarity-confirmed.
- [x] Sibling tests added for every new/touched production file.
- [x] `node .claude/skills/orchestrator/preflight-check.js` — all production files covered; no rule/language/blame-shift violations. (The "integration test gap" advisory is expected: this PR is client-only per S0 — the wire schemas' Q10 integration test was Phase 1's responsibility, already merged in #1357/#1353; this PR only consumes the already-tested contract via `v.parse` at the client boundary, with its own schema-mismatch test.)
- [x] Browser QA with true-path screenshots (real artifact + real completed/failed worktree deletion, real dev instance, own-server identity verified) — see "Browser QA" section above.

Closes #1354

## Summary by CodeRabbit

- **New Features**
  - Added a notification center to desktop and mobile navigation.
  - Displays unread counts, loading, empty, and retryable error states, with up to 50 notifications.
  - Supports links to created artifacts and completed worktree deletions.
  - Refreshes on relevant events, reconnects, and panel opening.
  - Supports marking notifications as seen and closing via Escape or outside clicks.
  - Improved dialog accessibility.

- **Documentation**
  - Clarified notification delivery behavior and implementation status.


ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

ms2sato:
## QA Screenshots (2026-08-19 10:02)

### 3a deeplink worktree deletion
3a deeplink worktree deletion

<details>

3a deeplink worktree deletion
</details>

### 3b deeplink artifact
3b deeplink artifact

<details>

3b deeplink artifact
</details>

### final 1 badge nonzero real unread
final 1 badge nonzero real unread

<details>

final 1 badge nonzero real unread
</details>

### final 2 panel all kinds failed red
final 2 panel all kinds failed red

<details>

final 2 panel all kinds failed red
</details>

### final 3 badge cleared after reload
final 3 badge cleared after reload

<details>

final 3 badge cleared after reload
</details>

### final 4 mobile badge with count
final 4 mobile badge with count

<details>

final 4 mobile badge with count
</details>

### final 5 mobile panel with data
final 5 mobile panel with data

<details>

final 5 mobile panel with data
</details>

### final 6 mobile badge cleared
final 6 mobile badge cleared

<details>

final 6 mobile badge cleared
</details>

ms2sato:
@coderabbitai review

ms2sato:
## Orchestrator note: CodeRabbit review coverage of the final commit

Recording this here rather than only in the session log, because it is a disposition a future reader needs in order to evaluate the merge.

**State.** CodeRabbit's round-3 review ran against head `6930cb7c` and returned `Merge Risk: Minimal` with 1 actionable comment (Minor / maintainability): three `mockFetch.mock.calls as unknown[][]` casts in `NotificationBell.test.tsx` that discard type information the mock already provides.

That finding was addressed in `13a1b07d`. **`13a1b07d` itself has not been reviewed by CodeRabbit** — the push landed inside the quota window (1 review per hour), and the bot posted a rate-limit notice rather than re-reviewing.

**Disposition: proceeding without a re-review of `13a1b07d`.** This is the Orchestrator's call, not the delegate's.

Reasoning:

- The commit touches one test file and nothing else. Zero production code, zero behavior change — it removes redundant casts at three call sites.
- It is precisely the change CodeRabbit asked for, at the three locations it named (L184 / L262 / L386), with no discretionary scope added.
- The full suite and all 10 checks are green on `13a1b07d`.
- The alternative — holding for the next quota slot — would spend the slot that PR #1368 needs, and #1368 is the fix for a bug that currently makes HTML Artifacts unusable from the only entry point the owner actually uses (#1366).

**Sibling-site check performed before accepting the finding.** `as unknown[][]` appears 16 times across 6 client test files, while the typed `Array<Parameters<typeof ...>>` form appears only in server tests. That looks like two competing idioms, and the consistency argument would have pointed at leaving the casts alone. It is not two idioms: the other files declare their mock as `mock(() => Promise.resolve(new Response()))` — no parameters, so `mock.calls` has no argument types to destructure and the cast is load-bearing there. This file declares `mock(async (input: RequestInfo | URL, init?: RequestInit) => ...)`, so the cast was discarding types that already existed. CodeRabbit is correct here and the other 15 sites are correct as they stand; they were deliberately left untouched and no follow-up Issue was filed, because "type the untyped mocks" is a different and larger change than the one flagged.

PR #1365, opened 2026-08-18
refactor: rename embedded-agent engine value native-loop to openai-api

### LINKED ISSUE #1364: Rename the embedded-agent engine value native-loop -> openai-api (vendor + access-method naming)
Owner decision, 2026-08-18. Rename the embedded-agent engine discriminant value `native-loop` to **`openai-api`**.

## Why

`native-loop` names the wrong axis and reads as the opposite of what it is. "Native" invites the reading "Claude-native", while the value actually means *our own agent loop talking to an OpenAI-compatible endpoint* — the generic side, the one that runs Ollama, DeepSeek, LM Studio, or anything else speaking that dialect. The owner's report of the confusion is the direct trigger: reading the two engine names side by side did not make the distinction legible.

## The naming scheme this establishes

**`<vendor-or-dialect>-<access method>`**, filling a 2x2 rather than a list:

| | api | sdk |
|---|---|---|
| **openai** | **`openai-api`** (this rename) | `openai-sdk` (hypothetical: OpenAI Agents SDK) |
| **claude** | `claude-api` (hypothetical: our own loop against the Anthropic API) | `claude-sdk` (exists today) |

The scheme was the owner's; it resolves an objection raised during the discussion (that naming by API dialect alone would collide once a second vendor SDK appeared) — under vendor+method there is no collision, because the loop-owner dimension is carried by the method slot.

`compat` is deliberately NOT in the name: "OpenAI API" already denotes the dialect in common usage, and the field is full of compatible implementations, so the qualifier adds length without adding meaning. The provider's actual endpoint is already carried by `baseUrl` on the definition.

## Scope

154 occurrences repo-wide; 15 production files. The substantive parts:

- **DB migration** rewriting existing `embedded_agents.engine` rows (`'native-loop'` -> `'openai-api'`), plus `schema.ts`'s column type and every doc comment that names the old value.
- Shared type + valibot schema (`packages/shared/src/types/embedded-agent.ts`, `schemas/embedded-agent.ts`) — the discriminant appears in both the definition type and the subprocess init command union.
- Server: `embedded-agent-manager.ts`, `embedded-agent-worker-service.ts`, `mappers.ts`, `worker-types.ts`.
- Subprocess: `main.ts`'s engine dispatch, `engine-types.ts`, `sdk-engine.ts`'s comparisons, `system-prompt.ts`.
- Client: `EmbeddedAgentWorkerView.tsx` (the engine-conditional banners shipped in #1335), `routes/agents/index.tsx`.
- Docs: `embedded-agent-sdk-engine.md` (§3.1, §4, §4.3's polarity table), `embedded-agent-worker.md`, `docs/glossary.md`'s Engine entry.

## Constraint worth confirming during implementation

Spec §3.1 states the engine is **not user-facing configuration** — the user picks an agent definition, and that determines the engine. If that holds, this rename should produce **zero user-visible string changes**: no picker label, no banner text, no form field. A UI copy change appearing in the diff means the engine leaked to a surface it should not have, which is itself worth reporting rather than quietly renaming.

## Timing

Do this **before the production instance is restarted**. Every day the instance runs, more `embedded_agents` rows accumulate carrying the old value; the migration is trivial either way, but the window where a stale client and a renamed server disagree is smallest while nothing is live on it.

## Not in scope

Renaming `claude-sdk` (already fits the scheme); adding any new engine; changing what either engine does.

---

## Acceptance Criteria

Naming decided by the owner (vendor × access-method convention; `compat` deliberately omitted — "OpenAI API" already names the dialect, `baseUrl` carries the real endpoint). This AC covers execution, not re-litigation.

Measured at `b006d9c1`: `native-loop` appears at 143 sites under `packages/` plus 11 in `docs/`, and **migration v29's DDL carries `DEFAULT 'native-loop'`** — that stale default is the core design point of R1.

### R1 — Migration (v32): row rewrite + table rebuild for the DEFAULT; forward-only

- **Do NOT edit migration v29** (applied migrations are frozen). New `migrateToV32`: rebuild `embedded_agents` via v29's own proven pattern (create `_new` with `DEFAULT 'openai-api'`, copy with `CASE WHEN engine = 'native-loop' THEN 'openai-api' ELSE engine END`, drop, rename, index/trigger snapshot-restore, FK check, `user_version = 32`). Rationale for the rebuild over a bare `UPDATE`: SQLite cannot alter a column DEFAULT in place, and leaving `DEFAULT 'native-loop'` behind plants a dead value that any future insert path relying on the default would resurrect — structure over convention.
- **No down migration** — the repo's migration chain is forward-only by design (no down mechanism exists); the rollback story is the deploy's DB backup, stated in the PR body rather than implied.
- Migration test per the `migration-vNN.test.ts` series: seed a v31-shaped DB with `'native-loop'` rows (+ one `'claude-sdk'` row as the must-not-touch control) → run the real `migrateToV32` → rows read `'openai-api'`, the claude-sdk row is byte-unchanged, AND a raw-SQL insert omitting `engine` gets the NEW default.
- Note (containment, not a gap): an unmigrated `'native-loop'` row read through the renamed mapper throws `DataIntegrityError` → `findAll`'s skip-with-warn (the #1342 containment). Unreachable in practice (migrations run before reads at startup); state this in the PR body instead of adding a compat branch — **do not accept both values anywhere** (a dual-accept shim would create the third state this rename exists to avoid).

### R2 — "Zero user-visible copy changes" — proven by test-diff shape, not assertion

§3.1 says the engine is not user-facing; therefore this rename must change NO rendered string. Verification form:

1. **Client-test discipline**: fixture/data values (`engine: 'native-loop'` → `'openai-api'`) change; **rendered-copy assertions (banner text, labels, picker strings) must be byte-unchanged**. Any diff hunk touching a user-visible string assertion = **STOP-and-report** — it means a surface leaks `engine`, which is a §3.1 finding to surface, not to silently rename past.
2. **PR body ledger**: `git grep -n "native-loop"` at the final head, with every remaining hit classified as exactly one of: (a) migration v32 code/tests (must contain the old literal), (b) frozen historical text — v29's own comments/DDL and correction-trail quotes (annotated in place with one line: "historical value; renamed to `openai-api` in #1364"), (c) nothing else. **Zero hits outside (a)/(b).** Live normative doc text (spec §3.1/§4.3 tables, glossary, worker.md) renames fully; history is not rewritten.

### R3 — Wire window: the existing schema-version handshake IS the mitigation

The client never sends `engine` (the create route hardcodes it server-side), so only server→new-value→old-client matters: the old client's `v.variant` literal fails to parse → but `SCHEMA_VERSION` changes with this PR (schemas touched) → the existing `SchemaVersionBanner` tells the user to reload. Accept and document: no dual-accept, no shim.

AC check: `schema-version.gen.ts` is regenerated in the same commit as the schema literal change (the two-delegates-forgot lesson from #1356 — the regen instruction belongs in the delegation prompt).

### R4 — The three replacement nets, named (this is how 143 sites are verified)

1. **TS union net**: renaming the shared union member makes every TYPED `=== 'native-loop'` comparison a TS2367 (no-overlap) error — including #1335's two banner gates in `EmbeddedAgentWorkerView` and the mappers' arms. Typecheck green = all typed comparison sites converted.
2. **Schema net**: `v.literal` / `v.variant` strings are invisible to TS2367 — covered by the existing schema tests (accept/reject arms) whose fixtures rename, plus the wire-boundary integration tests.
3. **Ledger net**: R2's grep ledger catches raw strings in neither net (SQL, docs, comments).

State all three in the PR body. The #1335 polarity table (§4.3) renames its `native-loop` row label and **keeps its 4 rows including unresolved** — the unresolved row must survive the rename (this is the regression the Issue flags).

### R5 — Docs

Spec §3.1: record the **vendor × access-method naming convention** (the 2x2, with the two hypothetical cells marked hypothetical) as the rule future engines follow, including the owner's `compat`-omission rationale — this converts a one-off rename into a deferral-proof convention. Glossary `Engine` entry renames + a one-line correction trail. §4.3 and `embedded-agent-worker.md` follow R2's live-text rule.

### V — Verification

Full suite + typecheck + preflight; the migration test per R1; the R2 ledger in the PR body; Browser QA smoke (one screenshot: the agents page and a worker view with an `openai-api` definition rendering identically to before — "nothing visibly changed" is itself the true-path proof here).

### Collision map (vs #1362, in flight)

#1364 touches: shared types/schemas (embedded-agent), server database/mappers/services fixtures, the embedded-agent package, `EmbeddedAgentWorkerView.tsx`, `routes/agents/index.tsx`, docs.

**#1362 territory it must NOT touch**: `components/header/**`, `components/notifications/**`, `lib/api.ts`, `lib/query-keys.ts`, `routes/__root.tsx`. Expected intersection: none (#1362 carries no `engine` reference), so the two can run in parallel — but a large mechanical-replacement PR is expensive for whoever rebases, so **merge order should be #1362 first, #1364 rebasing last**. The timing constraint (before the production restart) goes in the PR body as an ops note.

ms2sato:
## Summary

Renames the embedded-agent engine discriminant value `native-loop` to `openai-api` (owner decision, Issue #1364). `claude-sdk` is unchanged. Adopts a `<vendor-or-dialect>-<access method>` naming convention (2x2: `openai-api` / `claude-sdk` implemented, `claude-api` / `openai-sdk` documented as hypothetical future slots) so a future third engine has a place to go instead of an ad-hoc name.

143 sites under `packages/`, 11 in `docs/` renamed.

## R1 — Migration v32 (table rebuild, not UPDATE)

Migration v29's DDL declared `engine TEXT NOT NULL DEFAULT 'native-loop'`. SQLite cannot `ALTER` a column `DEFAULT` in place, so a bare `UPDATE` would leave that stale default behind -- any future insert path relying on the default would resurrect `native-loop`. New `migrateToV32` rebuilds `embedded_agents` following v29's own proven pattern (create `_new` with `DEFAULT 'openai-api'`, `INSERT ... SELECT` with a `CASE WHEN engine = 'native-loop' THEN 'openai-api' ELSE engine END` rewrite, drop, rename, index/trigger snapshot-restore, FK check, `user_version = 32`). `migrateToV29` itself is untouched (frozen; applied migrations are not edited), annotated in place noting its `'native-loop'` default/backfill is historical.

No down migration -- the chain is forward-only by design; rollback story is the deploy's DB backup.

`migration-v32.test.ts` seeds a v31-shaped DB with `native-loop` rows + a `claude-sdk` control row, runs the real `migrateToV32`, and asserts: rows rewritten to `openai-api`; the control row byte-unchanged; **a raw-SQL insert omitting `engine` gets the NEW `openai-api` default** (the assertion that would fail if this had been implemented as a bare `UPDATE` instead of a rebuild -- the whole point of R1); idempotent re-run; all other columns preserved.

**Containment note (not a gap):** an unmigrated `native-loop` row read through the renamed mapper throws `DataIntegrityError` -> `findAll`'s skip-with-warn (the #1342 containment). Unreachable in practice since migrations run before reads at startup. No dual-accept shim added -- both values are never accepted anywhere, per the AC.

## R2 -- Zero user-visible copy changes

Spec says the engine is not user-facing configuration. Verified by test-diff shape:

- **Client tests**: fixture/data values changed (`engine: 'native-loop'` -> `'openai-api'`); every rendered-copy assertion (banner text, labels) is byte-unchanged. No STOP-and-report trigger fired -- confirmed by direct review of every changed test hunk.
- Also verified live: Browser QA (below) created a real `openai-api`-engine definition through the UI form and opened its worker view -- the persistent restore banner rendered its pre-existing generic wording ("Conversation is restored automatically...") with zero mention of "engine"/"native-loop"/"openai-api" anywhere on either the `/agents` page or the worker chat view.

## R2.2 -- Grep ledger (at final head)

`git grep -n "native-loop" -- packages docs`, every remaining hit classified:

**(a) Migration v32 code/tests -- correctly contains the old literal (rewrite source + seed fixtures):**
- `packages/server/src/database/connection.ts` -- `migrateToV32`'s own JSDoc/DDL/`CASE WHEN` (lines ~1809-1890)
- `packages/server/src/database/__tests__/migration-v32.test.ts` -- seed helper / control-row literals (new file)

**(b) Frozen historical text -- v29's own comments/DDL, annotated in place with a one-line historical note (NOT renamed, per "do not edit migration v29"):**
- `packages/server/src/database/connection.ts` -- `migrateToV29`'s JSDoc/DDL (lines ~1625-1716)
- `packages/server/src/database/__tests__/migration-v29.test.ts` -- asserts v29's actual historical backfill value; `String(row.engine)` sidesteps the TS type (current union) vs. runtime value (historical) mismatch without an `as` cast

**(c) New correction-trail annotations added by this PR, referencing the old name as historical context (not a miss -- this is the R5 "one-line correction trail" itself):**
- `docs/glossary.md` -- Engine entry's new `Aliases:` line
- `docs/design/embedded-agent-sdk-engine.md` §3.1 -- new naming-convention paragraph's rationale sentence

Zero hits outside (a)/(b)/(c). All live normative doc text (glossary, §3.1, §4.3 table, `embedded-agent-worker.md`) renamed fully.

## R3 -- Wire window

Client never sends `engine` (create route hardcodes it server-side). `SCHEMA_VERSION` changes with this PR (schemas touched, regenerated in the same commit) -- an old client's `v.variant` literal fails to parse, and the existing `SchemaVersionBanner` tells the user to reload. No dual-accept, no shim, per the AC.

Preflight flags "Integration test gap -- cross-package change (client + server)" for the touched shared-type files. This is `pre-pr-completeness.md` Q10's heuristic firing on a **value rename inside an existing discriminated union**, not a new field -- Q10 targets new derived fields crossing the wire undetected by `v.strictObject`'s permissive parse; that gap doesn't exist here since `engine` already existed on both sides and every typed comparison (TS union net) plus every schema literal (accept/reject arms in `embedded-agent.test.ts` / `app-server-message.test.ts`) already had coverage that just got its fixtures renamed. R3 explicitly rules out a new integration test in favor of the schema-version handshake; no new integration test added.

## R4 -- The three replacement nets (how 143 sites are verified)

1. **TS union net**: renaming the shared union member turns every typed `=== 'native-loop'` comparison into a TS2367 error -- including both #1335 banner gates in `EmbeddedAgentWorkerView` and the mappers' arms. `bun run typecheck` is green across all 5 packages -- proof every typed site converted.
2. **Schema net**: `v.literal`/`v.variant` strings are invisible to TS2367 -- covered by the existing schema tests' accept/reject arms (`embedded-agent.test.ts`, `app-server-message.test.ts`), whose fixtures renamed.
3. **Ledger net**: R2.2's grep ledger above catches raw strings in neither net (SQL, docs, comments).

The §4.3 polarity table keeps its 4 rows including the "engine unresolved" row (verified unchanged -- this is the regression the Issue flagged from #1360).

## R5 -- Docs

- `embedded-agent-sdk-engine.md` §3.1: new paragraph records the vendor x access-method 2x2 (with `claude-api`/`openai-sdk` marked hypothetical) and the `compat`-omission rationale.
- `docs/glossary.md` Engine entry: renamed + new `Aliases:` line as the one-line correction trail (former literal, renamed in #1364).
- §4.3 and `embedded-agent-worker.md`: live text renamed fully, including the `#engine-embedded-agent--*` anchor slug used by 4 cross-references.

## Known preflight limitation (reported, not a real gap)

`preflight-check.js`'s comment-only-diff detector is documented as fail-closed by design ("a bare `*`-prefixed line is only treated as comment content when a `/*` opener was already confirmed earlier in the SAME hunk... the worst case is a comment-only file still being required to have a sibling test"). Three files (`sdk-engine.ts`, `system-prompt.ts`, `claude-sdk-builtin.ts`) each had a single-line JSDoc rename whose block opener fell outside the `--unified=0` hunk, so the detector couldn't recognize them as comment-only and flagged missing sibling-test coverage. Manually verified all three diffs are 100% comment-only (pasted in review). Rather than write a synthetic test, added an honest one-line explanatory note to each existing sibling test file (`sdk-engine.test.ts`, `system-prompt.test.ts`, `claude-sdk-builtin.test.ts`) -- this satisfies the mechanical gate the tool's own documentation predicts as the accepted cost of the fail-closed design.

## Verification

- `bun run typecheck`: clean across all 5 packages.
- `bun run test` (full suite): **4094+1 server / 375 embedded-agent / 2212 client / 659 shared / 82+3 integration / 538 scripts-hooks-skills = all pass, 0 fail**, exit 0 (run twice, identical).
- `node .claude/skills/orchestrator/preflight-check.js`: exit 0 (8/8 covered, integration gap explained above, no rule/skill duplication, no language violations, no blame-shift violations).
- Browser QA: created a real `openai-api`-engine definition via the `/agents` UI form, added it as a worker to a fresh session, opened the worker view. Both screenshots attached below.
- CodeRabbit CLI: per this environment's known headless-auth hang, not run at all in this session -- fell back directly to the GitHub-side bot per `coderabbit-ops`.

## Note on CodeRabbit

GitHub-side bot commit status for the head commit reads `Review rate limited` (verified via the commit-status `description`, not the rollup's `CodeRabbit=SUCCESS`, which is `success` regardless of whether a review ran). Local CLI was not attempted (known headless-auth hang in this environment). Following the rate-limit fallback disposition in `coderabbit-ops/troubleshooting.md`: proceeding to open this PR without a CodeRabbit verdict yet; confirm `reviewDecision` is `APPROVED` (or apply the documented fallback) before merge.

## Not in scope (per Issue)

Renaming `claude-sdk` (already fits the scheme). Adding any new engine. Changing what either engine does. #1362 (client header/notification work, in flight) territory was not touched.

Closes #1364

## Summary by CodeRabbit

* **Changes**
  * Renamed the embedded-agent engine from `native-loop` to `openai-api` across configuration, creation, editing, runtime initialization, and validation.
  * Existing embedded agents are automatically migrated to the new engine name while preserving provider settings and Claude SDK agents.
  * Transcript restoration and related UI messaging now use the updated engine designation.
* **Documentation**
  * Updated engine terminology, glossary entries, configuration guidance, and event semantics.
* **Tests**
  * Expanded migration coverage and updated validation scenarios for the renamed engine.


ms2sato:
## QA Screenshots (2026-08-18 21:03)

### agents page openai api definition full
agents page openai api definition full

<details>

agents page openai api definition full
</details>

### agents page openai api definition scrolled
agents page openai api definition scrolled

<details>

agents page openai api definition scrolled
</details>

### agents page openai api definition
agents page openai api definition

<details>

agents page openai api definition
</details>

### embedded agent worker view openai api
embedded agent worker view openai api

<details>

embedded agent worker view openai api
</details>

ms2sato:
@coderabbitai review

ms2sato:
@coderabbitai review

ms2sato:
## Orchestrator note: CodeRabbit nitpick on `migration-v27.test.ts:44` — verified, no change needed

Recording the disposition here rather than only in the session log, since "we looked and it was fine" is otherwise indistinguishable from "we didn't look".

**The finding.** CodeRabbit (Trivial) suggested that changing the `engine` seed in `migration-v27.test.ts` from `native-loop` to `openai-api` might bypass the v32 conversion, if that fixture seeds a pre-v32 schema and then runs the full migration chain. It offered an alternative acceptance condition: confirm `migration-v32.test.ts` covers the same seeded path.

**Why the premise does not hold here.** The fixture does not seed a pre-v32 schema. `migration-v27.test.ts` calls `initializeDatabase(':memory:')`, which runs *every* migration to completion — the same file asserts `PRAGMA user_version` is `32` immediately afterwards. Rows are inserted *after* that, into an already-migrated schema. So `openai-api` is the correct value there; seeding `native-loop` would mean writing a legacy value into a current-schema table, which is the opposite of what the test is for.

**And the alternative condition is satisfied anyway.** `migration-v32.test.ts` exercises the conversion head-on, against a hand-rebuilt legacy schema rather than the current one:

- `engine TEXT NOT NULL DEFAULT 'native-loop'` — the pre-v32 DDL, reconstructed
- seeds a row with `engine: 'native-loop'`
- calls the real `migrateToV32(db)`
- asserts `expect(nativeLoopRow.engine).toBe('openai-api')`

Its test name states both halves: *"rewrites existing 'native-loop' rows to 'openai-api', and passes a 'claude-sdk' control row through byte-unchanged"* — so the conversion has a positive case and a negative control, not just a happy path.

**Disposition: no change.** The question CodeRabbit asked was a good one — "did this rename hollow out the migration-chain coverage?" is exactly what should be asked of a mass-rename PR touching migration fixtures. It just does not apply to this file. Verified against the code rather than dismissed on severity.

Issue #1345, opened 2026-08-17
Secure-context-only browser APIs called unguarded: embedded-agent send breaks over plain-HTTP LAN addresses


ms2sato:
## Symptom (found by dogfood, 2026-08-17)

Opening the app over a plain-HTTP LAN address (`http://192.168.1.12:5173`) and sending a message to an embedded agent fails with:

```
Failed to send message: TypeError: crypto.randomUUID is not a function
  at EmbeddedAgentController.sendUserMessage (embedded-agent-store.ts:269:36)
  at useEmbeddedAgentWorker.ts:61:66
  at onSend (EmbeddedAgentWorkerView.tsx:337:17)
```

The user sees a Send button that does nothing. The error is console-only — nothing in the UI says why.

`crypto.randomUUID()` exists only in a **secure context**: HTTPS, or `localhost`. Served from an IP address over HTTP, `crypto.randomUUID` is undefined.

## This class of bug has been hit twice already, and both remedies are local

The codebase has recognised this exact problem before, in two different places, and neither remedy reached the sites below:

1. **`packages/client/src/lib/id.ts`** exports `generateTaskId()`, whose doc comment says verbatim: *"crypto.randomUUID() requires a secure context (HTTPS or localhost). When the app is accessed via IP address over HTTP, we fall back to a timestamp + random hex string."* Five call sites use it correctly.
2. **`packages/client/src/components/workers/EmbeddedAgentWorkerView.tsx:467`** guards clipboard use with `navigator.clipboard && window.isSecureContext`, and its comment cites the concrete case: *"at http://192.168.x.x:5173/"*.

So the knowledge exists twice, written down twice, and applied only where it was first needed.

## Sites to fix

| Site | API | Current state |
|---|---|---|
| `components/workers/embedded-agent-store.ts:269` | `crypto.randomUUID()` | Direct call, bypasses `generateTaskId()` |
| `components/settings/McpInstallSection.tsx:36` | `navigator.clipboard.writeText` | Unguarded |
| `components/sessions/SessionSettingsMenu.tsx:101` | `navigator.clipboard.writeText` | Unguarded |

Verified by grep at main: these are the only remaining unguarded uses of either API in `packages/client/src` (excluding tests).

## Why fix the class, not just line 269

Fixing only the reported line leaves two identical failures one click away, in a build where the same fix pattern already exists twice. The clipboard sites fail the same way — throw, console-only, silent to the user — under the same condition.

[...]

PR #1368, opened 2026-08-19
fix: serve HTML artifacts from header-blind (plain-HTTP, non-localhost) origins

### LINKED ISSUE #1366: HTML artifacts are unviewable from a plain-HTTP LAN origin (Sec-Fetch-Dest never arrives, P6 gate loops)
Found during the owner's post-restart dogfood of HTML Artifacts, 2026-08-19.

## Symptom

Opening an artifact from the LAN address the owner always uses — `http://192.168.1.12:6340/artifacts/<id>` — never renders the artifact. The viewer shell renders, then a second copy of the shell renders inside its own iframe, and the browser eventually aborts. In multi-user mode the visible end state is a raw `{"error":"Unauthorized","message":"Authentication required"}` page.

The same artifact, same server, same browser, opened via `http://localhost:6340/artifacts/<id>` renders correctly.

## Root cause (measured, not inferred)

`routes/artifacts.ts`'s P6 gate serves raw bytes only when the request carries `Sec-Fetch-Dest: iframe`, and treats **the header's absence** as "not an iframe" — redirecting to the viewer shell. The shell embeds that same endpoint in its iframe, so an absent header produces shell → iframe → shell → …

Browsers send `Sec-Fetch-*` (Fetch Metadata) **only to potentially trustworthy origins** — HTTPS, or localhost. A plain-HTTP non-localhost origin gets none of them.

Controlled experiment, real Chromium 151, same server (dev instance, `AUTH_MODE=none`, so authentication is out of the picture), same artifact, only the origin varied:

**`http://192.168.1.12:3457/artifacts/<id>`**
```
GET /artifacts/<id>      200   (shell)
GET /api/artifacts/<id>  302   -> Location: /artifacts/<id>
GET /artifacts/<id>      200   (shell again, inside the iframe)
GET /api/artifacts/<id>  net::ERR_ABORTED
```
The 302'd request's headers contain **no `Sec-Fetch-*` at all**.

**`http://localhost:3457/artifacts/<id>`**
```
GET /artifacts/<id>      200
GET /api/artifacts/<id>  200 + artifact bytes
```
That request carries `sec-fetch-dest: iframe`, `sec-fetch-mode: navigate`, `sec-fetch-site: same-origin`.

Same result reproduced against the production instance (6340) with a logged-in browser: LAN address nests, `localhost` renders.

## Why this matters more than it looks

The owner's LAN address is not a convenience path — it is the permanent, only entry point for normal use (`localhost` is reachable only while sitting at the server). So HTML Artifacts is, in practice, unusable from the way the system is actually accessed. Every artifact an agent creates is unopenable.

## The design premise that does not hold

`docs/design/html-artifacts.md` §3.3 / P6 describes absence of `Sec-Fetch-Dest` as failing **closed** — safe, because a non-browser client gets redirected to the shell rather than raw bytes. That reasoning is sound in isolation, but the shell *is* the thing that requests the endpoint, so on an origin where the header never arrives, failing closed does not degrade to "safe" — it degrades to "the feature cannot run at all", and it does so silently.

The premise is also stated unconditionally, while the browser behavior it relies on is conditional on origin trustworthiness. That conditionality is the part the spec does not record.

## What to decide (Architect)

This touches the sandbox boundary, so the fix is a design decision rather than a patch:

- Give the shell's own iframe request a credential the gate can recognize (a one-shot token minted per shell render), so the gate no longer depends on a header the browser may not send.
- Or accept `Referer` matching this origin's shell URL as an alternative condition when `Sec-Fetch-Dest` is absent — weaker, and needs deriving against the same threat model the header was chosen for.
- Or something else; the point is that the current gate's safety argument and its liveness argument are coupled through a header whose presence is not guaranteed.

Whatever is chosen, `docs/design/html-artifacts.md` should record that Fetch Metadata is origin-conditional, so the next reader does not re-derive the same unconditional premise.

## Verification requirement

Any fix must be verified from a **plain-HTTP non-localhost origin** with a real browser — that is the condition under which the bug exists, and it is exactly the condition `localhost`-based QA cannot reach. The `check:artifact-sandbox-boundary` smoke should gain an arm for it, or state explicitly why it cannot.

ms2sato:
## Summary

HTML artifacts were unviewable from any plain-HTTP, non-localhost origin — including the owner's LAN address, the permanent way this app is actually reached day to day. `routes/artifacts.ts`'s gate served raw bytes only when `Sec-Fetch-Dest: iframe` was present, and redirected to the viewer shell on absence. Browsers only attach `Sec-Fetch-*` (Fetch Metadata) headers on potentially-trustworthy origins (HTTPS, or localhost), so on a plain-HTTP LAN origin the header never arrives — including from the shell's *own* genuine iframe load. The shell then nested inside itself indefinitely (shell → iframe → shell → …), terminating either in `net::ERR_ABORTED` (single-user mode) or a raw `401 Unauthorized` (multi-user mode, since the auth cookie is withheld once the load is nested inside an opaque-origin sandboxed context).

This implements the Architect-approved fix (approach a′, full ruling and AC in the Issue's comment thread): a two-tier gate. `Sec-Fetch-Dest`, when present, always wins. When entirely absent, the gate falls back to a single-use, 60s-TTL, artifact-id-bound token the viewer shell mints on every render and embeds in the iframe `src` (`?vt=<token>`). The token is never a substitute for authentication — `authMiddleware` is unchanged on both routes; the token proves only "this request was induced by our own shell's render."

Closes #1366

## What changed

- **`packages/server/src/routes/artifacts.ts`** — two-tier gate (`Sec-Fetch-Dest` present → header wins; absent → token fallback). Added `Cache-Control: no-store` to the raw 200 response (closes a browser-cache bypass of single-use).
- **`packages/server/src/routes/artifacts-viewer.ts`** — mints a token unconditionally on every render, embeds it in the iframe `src`.
- **`packages/server/src/lib/artifact-viewer-tokens.ts`** (new) — in-memory, single-use, TTL-bound, artifact-id-bound token store. 128-bit (`crypto.randomBytes(16)`) base64url tokens, no persistence, no signing, injectable clock for deterministic TTL tests.
- **`docs/design/html-artifacts.md`** — §3.3 P6 → P6′-a/P6′-b with a correction trail; new §4.0 documenting the token mechanism and its accepted residual (a server restart invalidates outstanding tokens, degrading to the pre-existing nested-shell recovery on a stale reload); a forward-looking note pointing at #1345 (HTTPS provisioning would dissolve this whole bug class).
- **`docs/glossary.md`** — Artifact entry updated with the token mechanism.
- **`scripts/smoke/check-artifact-sandbox-boundary.mjs`** — new arm 8 (header-blind LAN-IP origin, this Issue's V1) and arm 8b (top-level-open jail re-verification, V2); `HOST` changed from a hardcoded `127.0.0.1` to `0.0.0.0` so the disposable instance is LAN-reachable for arm 8 (arms 0-7 unaffected — they keep using their own `127.0.0.1` URLs, which still work under `0.0.0.0`).
- **`scripts/smoke/check-artifact-server-story-e2e.mjs`** — added a `Cache-Control: no-store` assertion to the existing serve block.
- Sibling tests: full matrix in `routes/__tests__/artifacts.test.ts`, additions to `routes/__tests__/artifacts-viewer.test.ts`, new `lib/__tests__/artifact-viewer-tokens.test.ts`.

## Non-goals (per the Architect ruling, not implemented)

Referer-based fallback (rejected — swaps one environment-conditional browser signal for another, re-enacting this Issue's exact failure class); serving raw bytes at top level under any condition (P7 unchanged); HTTPS provisioning (recorded as a future direction only, #1345); token persistence or signing.

## Security remediation (post-review)

CodeRabbit's first pass on this PR flagged 4 findings; 3 were straightforward docs/test-precision fixes (below), and 1 was a **Major/Security** finding that changed the Architect's verdict from CLEAN to REJECT — this PR could not merge until it was addressed. All 4 are fixed and re-verified in this PR as it now stands.

1. **(Fixed) Docs — P6′-b description was ambiguous about the absent-header case on a trustworthy origin.** The gate branches purely on whether `Sec-Fetch-Dest` is present or absent — never on origin trustworthiness directly (the implementation cannot make that distinction). `docs/design/html-artifacts.md` §3.3 P6′-b now states: "header present → header decides; header absent → token decides," with origin trustworthiness kept only as explanatory context for *why* absence happens on some origins, never as a branch condition the code checks.
2. **(Fixed) Test — a test's name claimed coverage it didn't have.** `artifacts.test.ts`'s test titled "... even with an invalid token present ..." only exercised a *valid* minted token. Added a real invalid-token request (`?vt=not-a-real-token`, asserted 200) so the test now covers what its name claims.
3. **(Fixed, Major/Security) The smoke script's original credential substitution was a real security defect.** `check-artifact-sandbox-boundary.mjs` originally monkey-patched `MultiUserMode.prototype.validateOsCredentials` to accept any password, on a server bound to `0.0.0.0` for LAN reachability (needed for arm 8, this Issue's own header-blind-origin arm). For the whole script run, any host on the LAN that knew the operator's OS username could have authenticated with any password and reached `POST /api/sessions` — a real PTY spawn as the operator's OS identity. This is an RCE window, not a probe-data leak. CodeRabbit's own suggested fix (bind to a single resolved LAN IP instead of `0.0.0.0`) was tried and rejected for two reasons: it only narrows the exposure window rather than closing it (arm 8 still needs LAN reachability by definition), and it was empirically confirmed to break the script outright — a single non-loopback bind makes `127.0.0.1` unreachable on that same socket (standard BSD-socket behavior), which destroys arms 0-7's trustworthy-vs-untrustworthy origin comparison (they depend on `127.0.0.1` remaining a trustworthy origin). The actual fix: stop patching `validateOsCredentials` entirely, and mint a real session JWT directly instead, replicating every step of `MultiUserMode.login()` that runs *after* password validation (real `lookupOsUser`, real `userRepository.upsertByOsUid` against the server's own live DB connection via `initializeDatabase()`'s no-argument fast path, real `SignJWT` with the server's own on-disk per-instance secret). The real login endpoint and the real password check are never called, patched, or weakened, for any reason, at any point — this is the same JWT-mint pattern the Architect specified (see `pre-pr-completeness.md` Q13's worked example) and is a *stronger* isolation than the withdrawn patch: authentication itself is untouched, only the mechanism this test harness uses to obtain its own credential changed. `HOST` remains `0.0.0.0` — the network bind was never the actual problem once the credential mechanism itself no longer accepts arbitrary passwords. Full smoke re-run (all arms, including 8/8b under the new mechanism) is pasted under V1 above.
4. **(Fixed, trivial) A test fixture minted a token without asserting its input was valid first.** `check-artifact-sandbox-boundary.mjs` arm 8b now asserts `toolResult.artifactId` is truthy before minting a token from it, matching the existing guard pattern already used in `check-artifact-server-story-e2e.mjs`.

### V1 — TDD polarity, real browser, header-blind LAN origin (the fix's load-bearing check)

`scripts/smoke/check-artifact-sandbox-boundary.mjs` arm 8, real Chromium, navigating `http://192.168.1.12:<port>/artifacts/<id>` (plain HTTP, non-localhost — confirmed zero requests to this origin carry `Sec-Fetch-Dest`, so the reproduction premise itself is verified, not assumed).

**Against unmodified `origin/main`** (reverted the two route files, reran):
```
GET /artifacts/:id      → 200 (shell, 1st render)
GET /api/artifacts/:id  → 302 (Sec-Fetch-Dest absent → redirect)
GET /artifacts/:id      → 200 (shell, 2nd render — nested inside the 1st iframe)
GET /api/artifacts/:id  → 401 Unauthorized  ← this is the multi-user-mode terminal state Issue #1366 itself described
```
4 assertions FAIL (no nested shell / script executed / no failed request / no 401 — all failed as expected), exit code 1.

**With the fix** (same arm, same script): the shell renders exactly once, its single child iframe is the raw artifact endpoint (not a second shell), the artifact's own script executes (confirmed via the existing `postMessage` harness), no failed/aborted requests, no 401. 37 assertions pass, exit code 0. Independently re-run and confirmed a second time after implementation, outside the implementing delegate's own session.

**Re-verified a third time after the credential-issuance mechanism itself was rewritten** (see "Security remediation" below) — same arm, same assertions, same pass count (37 assertions, exit 0) — confirming the fix's own behavior is unchanged by how the smoke harness obtains its test session.

### V2 — jail re-verification (arm 8b, same header-blind origin)

Top-level open of the raw `/api/artifacts/:id` URL with no token → redirects to the shell. Same with an already-spent token → redirects to the shell. Raw bytes never reach a top-level navigation either way. Existing arm 7b (headerless Node-`fetch` → redirect) still passes unmodified — it is now the "absent + no token" row of the gate matrix.

### V3 — trustworthy-origin regression

Arms 0-7 of the existing probe suite (P1/P2/P6/P7, `localhost`) pass byte-identically before and after.

### `check:artifact-server-story-e2e`

Re-run; added assertion for the new `Cache-Control: no-store` header. 16/16 assertions pass.

### V4 — real browser, real LAN origin, real multi-user mode (the owner's actual scenario)

**Why not the literal "dev instance":** `scripts/dev-multiuser.sh` shells out through `sudo` to run the server as the `agentconsole` service user; this session has no interactive privileged-shell password to satisfy that prompt, and the script failed at its rsync step with "a password is required" when attempted. The Docker dev/verification stacks were also not used: `docker/README.md` explicitly states both stacks' ports "must never be reachable from the LAN" (they ship known test credentials, `alice`/`bob`), so exposing either to `192.168.1.12` would have contradicted documented policy for the sake of this one verification.

**⚠️ Superseded credential-issuance approach (kept here only as the record of what was withdrawn, and why — see "Security remediation" below for the full incident).** An earlier version of this PR body described V4 as run against a server with `MultiUserMode.prototype.validateOsCredentials` monkey-patched to accept any password. CodeRabbit correctly flagged the equivalent pattern in `check-artifact-sandbox-boundary.mjs` as a **Major security defect**: for the whole duration such a script runs, on a server reachable from the LAN, any host on the network could authenticate as the operator with any password and reach `POST /api/sessions` (real PTY spawn as the operator's OS identity) — an RCE window, not just a probe-data leak. The Architect REJECTED that pattern outright. **V4 was fully re-run against the corrected mechanism below before this PR was finalized; no verification in this PR currently rests on the withdrawn pattern.**

**What was substituted instead, and recorded here per `pre-pr-completeness.md` Q13's three conditions (current, corrected mechanism):**

1. **What was substituted:** only the MECHANISM for obtaining a valid session credential — this script mints a real session JWT DIRECTLY, replicating every step of `MultiUserMode.login()` (`packages/server/src/services/user-mode.ts`) that runs *after* password validation: a real `lookupOsUser()` OS lookup, a real `userRepository.upsertByOsUid()` insert against the live disposable database, and a real `SignJWT` call using the exact same payload shape, algorithm, and expiry, signed with the real per-instance secret the server itself generated and persisted to `<AGENT_CONSOLE_HOME>/jwt-secret` at boot. `validateOsCredentials` and the real `POST /api/auth/login` endpoint are **never called, patched, or weakened** — anyone else on the LAN still needs a real password to authenticate, for the entire run.
2. **Why it is upstream of, and outside, the chain under test:** this Issue's fix is entirely inside the artifact-serving gate (`Sec-Fetch-Dest` / viewer token). How this test harness itself proves its identity to obtain a session token has no bearing on that gate's behavior once a session exists — the substitution sits strictly upstream of, and outside, what V4 verifies. This is a *stronger* isolation than the withdrawn patch: the real authentication path (both the login endpoint and the token-verification side, `MultiUserMode.authenticate()`) is never touched at all — a forged or wrong-secret token is rejected exactly as it would be in production, because this substitution constructs a credential the real, unmodified verification path independently accepts as genuine, rather than disabling that path.
3. **What remained production-real:** the server process itself, bound to `0.0.0.0` (bind address is no longer part of the fix — see the CodeRabbit-suggested-fix rejection below for why narrowing it wouldn't have helped anyway); real OS-user lookup / DB upsert / JWT signing against the real per-instance secret; the *entire* login-verification path on the server side, completely unmodified; a real Chrome browser (Chrome DevTools MCP); the real LAN origin (`192.168.1.12`); and the entire gate/token/shell/iframe logic this Issue changed. The minted JWT was delivered into the real browser's cookie jar via a real top-level navigation to a tiny, non-production, throwaway helper endpoint (not part of the app, no production route touched) that issued a genuine `Set-Cookie` response header and 302-redirected to the real artifact URL — real Set-Cookie processing by the browser, not `document.cookie` (which cannot touch the production `HttpOnly` cookie) and not a same-origin-restricted `fetch()`.

**Why CodeRabbit's own suggested fix (bind the exposed server to a single resolved LAN IP instead of `0.0.0.0`) was rejected — two independent reasons:**

1. **Insufficient on its own:** narrowing the bind address only narrows the exposure window, it does not close it. Arm 8 (and this V4 check) require LAN reachability by definition, so *something* on the LAN would still be able to reach an instance where the password check accepts anything — a smaller door, not no door.
2. **Actually breaks the harness:** binding to a single non-loopback address is standard BSD-socket behavior — the kernel then refuses connections to `127.0.0.1` on that same socket. `check-artifact-sandbox-boundary.mjs`'s arms 0-7 specifically depend on `127.0.0.1` being a *trustworthy* origin (so `Sec-Fetch-Dest` IS sent) as the basis for the trustworthy-vs-untrustworthy comparison the whole script makes; a single-IP bind would silently make arms 0-7 header-blind too, destroying that comparison. This was confirmed empirically (a standalone `Bun.serve({hostname: '<lan-ip>'})` test showed `127.0.0.1` fetches failing while the LAN IP succeeded) before the JWT-mint alternative was implemented.

The JWT-mint approach sidesteps both problems: it changes only the credential-issuance mechanism, not the network bind, so it needed no bind-address change at all and could not conflict with arms 0-7's trustworthy-origin premise.

**What was observed (current run, corrected mechanism):**

1. A real minted session JWT was set into the actual Chrome browser's cookie jar via a real `Set-Cookie` response header (delivered through a genuine top-level navigation + redirect, per point 3 above) — the real `/api/auth/login` endpoint was never called.
2. Navigated (having landed post-redirect) to `http://192.168.1.12:<port>/artifacts/<id>` — the shell rendered exactly once, with the artifact's own script executed inside the sandboxed iframe (screenshot attached below).
3. Network log for that navigation: exactly two requests — `GET /artifacts/:id` (200) and `GET /api/artifacts/:id?vt=...` (200). No nesting, no redirect loop, no 401.
4. A separate, earlier screenshot (still valid, since it depends only on being unauthenticated, not on how authentication is later obtained) shows the pre-login state — `{"error":"Unauthorized","message":"Authentication required"}` — confirming the route's own auth gate is untouched by any of this.

**Screenshots attached via PR comment** (`upload-qa-screenshots.sh`) — screenshot 1 is the pre-login state (still valid evidence, unrelated to the credential mechanism), screenshot 2 is the true-path render from the corrected, JWT-mint-based rig described above.

### Full suite / typecheck / lint

Independently re-run (not just trusting the implementing delegate's report):

```
bun run test   → all packages green (server 4109 pass/1 skip/0 fail, client 2212 pass, shared 659 pass ×2, embedded-agent 375 pass, integration 82+3 pass, scripts/hooks/skills 538 pass), TEST_EXIT: 0
bun run typecheck → exit 0, all 6 workspaces
bun run check:lang → exit 0 (121 files scanned)
node .claude/skills/orchestrator/preflight-check.js → exit 0
bun scripts/check-source-comment-blame-shift.mjs → 0 new violations
```

Re-run after each of the two follow-up commits (`af45bb23` Cache-Control fix, `67268de5` CodeRabbit remediation); same result each time. CI is green on `67268de5` (all checks, including the full `test` job).

### Joint decision (delegate + Orchestrator): `packages/integration` boundary test skipped

`preflight-check.js` advisory-flags both route files for integration-test review. Per `workflow.md` ("Skipping E2E is not a unilateral agent decision"), this is not a call the implementing delegate makes alone — it was raised with, and confirmed by, the Orchestrator before this PR was finalized. Reasoning:

- What changed is a query param (`?vt=`) and a response header (`Cache-Control`) on an existing route — not a shared type/schema addition. `packages/shared/src/schemas/` is untouched, so this is not the valibot-silently-strips-a-field failure mode `pre-pr-completeness.md` Q10 targets (that gap needs a wire-schema update; this PR has no wire schema to update).
- The two smoke scripts (`check-artifact-sandbox-boundary.mjs`, `check-artifact-server-story-e2e.mjs`) already drive the real HTTP / real-browser wire contract end to end — a strictly *stronger* check than a typical `packages/integration` boundary test for this specific change, not a weaker substitute for one, since they exercise the exact header-blind-origin scenario this Issue is about.

Decision: sufficient without an additional `packages/integration` test.

### CodeRabbit

Local CLI (`0.7.3`) is installed but `coderabbit review --agent --base main` fails with `authentication_failed` in this sandbox (no interactive browser for `coderabbit auth login`) — GitHub-side review used instead. First review pass returned 4 findings (1 Major/Security, 3 Minor); all 4 fixed, see "Security remediation" above. Second review pass (after the Security remediation commits): 0 actionable findings, 1 Trivial nitpick (see "Deferred" below). Final verdict: **CLEAN**.

**Deferred (CodeRabbit nitpick, Trivial):** extracting the duplicated `mintRealSessionToken` (present in both `check-artifact-sandbox-boundary.mjs` and `check-artifact-server-story-e2e.mjs`) into a shared smoke helper. Deferred to [#1369](https://github.com/ms2sato/agent-console/issues/1369) rather than done here: the repo's extraction threshold (`.claude/rules/elevation-helpers.md`) is two-PR convergence or three-plus callsites within one PR, and this PR has one PR with two callsites — under the threshold either way; and extracting now would mean editing `check-artifact-sandbox-boundary.mjs` again for a reason unrelated to its own fixes, immediately after it just went through a Major security remediation and several review rounds. **This is the Orchestrator's call, not the implementing delegate's** — Issue #1369 also records the case for doing it now (the "duplicate once, extract at the second real consumer" convention this PR's own header comments invoke, and the drift risk of two credential-minting copies silently diverging from `MultiUserMode.login()`), so a future reader has both sides, not just the deferral's justification.

## Design-boundary consultation trail

No Architect consultation was needed beyond the ruling already recorded on the Issue — the AC was implemented as written. The one open judgment call (the `packages/integration` test gap above) was **not** decided unilaterally by the implementing delegate — it is a joint decision with the Orchestrator, recorded in full in the "Joint decision" subsection above, including the Orchestrator's own concurring rationale (query param + response header, not a shared type/schema change; the two smoke scripts are stronger coverage than a boundary test for this specific change, not weaker). Recording who agreed, not just what was agreed, matters here specifically because a prior sprint (PR #926) had a delegate and the Orchestrator independently reach the same wrong "unit tests suffice" conclusion for a similar-looking gap — surfaced only by the owner's Browser QA. Naming both parties' agreement here is what would let a future reader identify a shared blind spot instead of assuming a lone one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed artifact viewer loading on plain-HTTP LAN origins where browser metadata headers may be unavailable.
  * Added secure, single-use viewer tokens that expire after 60 seconds and are tied to the requested artifact.
  * Prevented cached viewer or artifact responses from causing repeated or nested viewer loads.

* **Tests**
  * Expanded coverage for token validation, expiration, reuse, artifact matching, and LAN-origin loading.
  * Added end-to-end checks for response caching behavior and viewer isolation.


ms2sato:
## QA Screenshots (2026-08-19 10:22)

### 1 unauthenticated shell before login expected
1 unauthenticated shell before login expected

<details>

1 unauthenticated shell before login expected
</details>

### 2 true path artifact rendered multiuser lan origin
2 true path artifact rendered multiuser lan origin

<details>

2 true path artifact rendered multiuser lan origin
</details>

ms2sato:
@coderabbitai review

ms2sato:
## QA Screenshots (2026-08-19 11:56)

### 1 unauthenticated shell before login expected
1 unauthenticated shell before login expected

<details>

1 unauthenticated shell before login expected
</details>

### 2 true path artifact rendered jwt mint rig no auth bypass
2 true path artifact rendered jwt mint rig no auth bypass

<details>

2 true path artifact rendered jwt mint rig no auth bypass
</details>

ms2sato:
@coderabbitai review

PR #1377, opened 2026-08-20
feat: surface a session's own artifacts, scoped like the Memo panel

### LINKED ISSUE #1370: artifacts: surface a session's own artifacts as session-attached information
## What

Surface the artifacts a session produced, as information attached to that session — the way the memo already is.

Owner request, 2026-08-20: "あるセッションで作成したアーティファクト一覧はどこにも表示されてないかな？ Memo機能のような形でセッションに関連づく情報として出したいです。"

## The link already exists and nothing reads it

This is not a data-model change. `create_html_artifact` already records the creating session:

- `packages/server/src/mcp/mcp-server.ts` passes `sourceSessionId: sessionId` on every create
- `packages/server/src/repositories/artifact-repository.ts:16` declares `sourceSessionId: string | null`
- the SQLite column `source_session_id` is persisted

So every artifact created so far — including the ones already in the owner's production database — carries its originating session. What is missing is any way to read it back:

- `ArtifactRepository` exposes only `findById` and `findByUserId`; there is no `findBySourceSessionId`
- `GET /api/artifacts` (`routes/artifacts.ts:57`) is user-scoped, with no session filter
- nothing in the client surfaces the association

## Scope

1. `ArtifactRepository.findBySourceSessionId(sessionId)` plus its SQLite implementation and sibling tests.
2. A read path for it. Two shapes are plausible and the choice should be made deliberately rather than by whichever is quicker:
   - a query parameter on the existing list route (`GET /api/artifacts?sessionId=...`), or
   - a session-scoped route alongside the other per-session resources.
   Whichever is chosen, the **identity-scoping question in `design-principles.md` applies**: the fetch must be scoped by the caller's identity at the query, not fetched-then-filtered, and any row cap must be applied after the scoping, not before. The existing `findByUserId` is correctly identity-scoped; a new method must not regress that.
3. Client surface, associated with the session the way the memo is (`memo-service.ts` stores at `~/.agent-console/memos/{sessionId}.md`; artifacts are in SQLite, so the storage differs — the *presentation* is what should feel analogous, not the mechanism).

## Deliberately out of scope

- Cross-session browsing beyond what `GET /api/artifacts` already does.
- Any change to who may view an artifact. `docs/design/html-artifacts.md` §7 and requirement 3 (any logged-in user may view any artifact by direct URL; listing is scoped to the caller's own) are unchanged by this Issue. A session-scoped *list* must not become a way to enumerate another user's artifacts — the scoping is an additional filter on top of the existing user scope, never a replacement for it.

## Related

- `docs/design/html-artifacts.md` — §4 (routes), §7 (visibility)
- Siblings filed from the same conversation: delete via MCP, and update (which needs a design decision first).

ms2sato:
## Summary

Surfaces the HTML artifacts a session produced as information attached to that session, the way the Memo panel already is.

The link already existed — `create_html_artifact` has always recorded `sourceSessionId` on every artifact, so no migration or backfill was needed. What was missing was a read path:

- **R1**: `GET /api/artifacts` accepts an optional `?sessionId=<id>` query parameter, added to the existing list route (not a new route). Omitting it preserves the pre-existing behavior byte-identically — the no-`sessionId` branch is untouched code, and a client-server boundary test proves the two branches are genuinely independent (see Test plan).
- **R2**: New repository method `findByUserIdAndSourceSessionId(userId, sessionId)` scopes both `user_id` and `source_session_id` in one SQL query (`WHERE user_id = ? AND source_session_id = ?`), never a post-fetch JS filter — avoiding the #1357 R-3 shape where a JS-filtered session condition could hide the caller's own older rows behind other users' newer ones. Session ownership is deliberately **not** checked — the user scope already constrains the result set, so `sessionId` is a pure secondary filter, not an authorization boundary.
- **R3**: The client surfaces a session's artifacts as a new collapsible sidebar (`SessionArtifactsPanel`), modeled on `MemoPanel`'s presentation (hidden when empty, same collapse/expand affordance) — the presentation mirrors Memo, not the storage mechanism (Memo is a file, artifacts are SQLite rows). Deep links inherit #1340's navigation-jail rule: `/artifacts/{id}` is a plain `<a target="_blank">`, never an SPA `<Link>`.

`docs/design/html-artifacts.md` is updated (§4.2 new, §5.1 and §7 amended) to keep the design doc in sync with the new read path — it previously stated `source_session_id` was "never used for lookup," which this PR makes no longer true.

## Test plan

- `SqliteArtifactRepository.findByUserIdAndSourceSessionId`: positive case, identity-scoping polarity in both directions (another user's artifact in the same session; the same user's artifact in another session), and the #1357 R-3 shape explicitly (10 other-user artifacts in the same session, all newer, the caller's own artifact still appears — proving the query is genuinely SQL-scoped, not JS-filtered-after-fetch).
- Route tests for both `sessionId` branches (present/absent), the absent branch asserting the pre-existing behavior is unchanged.
- Client: `SessionArtifactsPanel.test.tsx` and `useSessionArtifacts.test.ts` cover render-with-data, hide-when-empty, and the jail-rule deep link (`tagName`, `target`, `rel`, `href`).
- `packages/integration/src/artifact-list-boundary.test.ts` extended with a real client→HTTP→server round trip proving `fetchArtifacts('session-a')` actually filters via the real Hono RPC client and query string, and that `fetchArtifacts()` (no argument) still returns artifacts from all sessions in the same test.
- `SessionPage.test.ts` has a type-level contract check pinning `SessionArtifactsPanel`'s `{ sessionId: string }` prop shape, mirroring this file's existing `AddAgentWorkerMenu` contract-check idiom. **Scope note, stated plainly**: this test pins the *prop signature* SessionPage.tsx's JSX relies on — it does **not** verify the panel actually mounts inside SessionPage's render tree. There is no full-render test of `SessionPage` anywhere in this codebase (it requires `SessionStopTasksContext.Provider`, WebSocket-backed hooks, and several other heavy dependencies with no existing test harness), so building one was judged disproportionate to this change. Actual render/behavior coverage lives in `SessionArtifactsPanel.test.tsx` / `useSessionArtifacts.test.ts`.
- Full `bun run typecheck` and `bun run test` pass (client 2253, server 4121+1 skip pre-existing/unrelated, shared 659, integration 83+3, embedded-agent 375, scripts/hooks/skills 538 — all 0 fail).
- `preflight-check.js`: all green (test coverage, integration test coverage, rule/skill duplication, language check, source-comment blame-shift check).

Browser QA (true-path, isolated dev instance running this branch's code): created a real quick session, created a real artifact via `create_html_artifact` attributed to that session, confirmed the artifact appears in that session's Artifacts panel with a working "View" link; confirmed a second, artifact-less session shows no panel (matching Memo's hidden-when-empty behavior) and that the two sessions' artifact lists are isolated from each other.

Closes #1370

## Summary by CodeRabbit

- **New Features**
  - Added a collapsible sidebar showing artifacts created during the active session.
  - Added session-specific artifact filtering while preserving access to all personal artifacts.
  - Artifacts display titles and timestamps and can be opened in new tabs.

- **Bug Fixes**
  - Ensured artifact results remain limited to the authenticated user and selected session.

- **Tests**
  - Added coverage for filtering, ordering, empty states, loading behavior, and sidebar interactions.


ms2sato:
## QA Screenshots (2026-08-20 09:12)

### qa 1370 panel empty hidden
qa 1370 panel empty hidden

<details>

qa 1370 panel empty hidden
</details>

### qa 1370 panel populated
qa 1370 panel populated

<details>

qa 1370 panel populated
</details>

ms2sato:
@coderabbitai review

